Issue #299, opened 2025-02-17
feat: Remove requirement to use the same keypair during ledger replay


GabrielePicco:
## Description

To resume correct operation, we must restart with the previous validator keypair since delegate accounts are specifically tied to the validator identity. However, the validator keypair is not actually required during ledger replay.

## Proposed Solution

Remove this requirement and only enforce the keypair check after replay is completed (before starting the RPC). This provides the following benefits:

- Enables broader replay capabilities – Anyone can replay a session.
- Improves testing – Replay can be tested in isolation without requiring the validator keypair; any keypair would suffice.

PR #301, opened 2025-02-17
feat: ledger replay with 'fake' validator keypair

thlorenz:
## Summary

Adding ability to prepare a ledger for replay in a diagnostics/test scenario.

Originally @GabrielePicco requested a way to basically ignore the keypair until after ledger
replay. However from looking at the way that the validator is initialized I realized this would
turn out to be a heavy handed solution to the problem.

Instead I decided to add a new command to the genx CLI that would allow for the keypair to be
overwritten right before we want to replay a specific legger. This is even more secure since
we're sure to wipe the original validator keypair from disk before starting the replay.

This also allows downloading the ledger without the keypair in the first place.

It is also safe to not cause any changes on chain (as a result of replay or txs after) since
the validator keypair is _incorrect_ and cannot sign any on-chain transactions.

FIXES: #299

NOTE: I'm basing this off the previous PR (#300) since builds up on it. Once that is merged we
need to rebase this PR on master to merge.

## Details

Here is an excerpt from the genx README explaining how to overwrite the validator keypair with a
random one when replaying the ledger to diagnose/test.

> ### test-ledger
>
> Prepares a ledger for replay in a diagnostics/test scenario. At this point it ensures that the
> original keypair of the validator operator is not used. Instead it is overwritten with a random
> keypair. That keypair is then printed to the terminal so that the `VALIDATOR_KEYPAIR` env var
> can be set to it when running the epehemeral validator to replay the ledger.
>
> One can auto-set that env var using the output of this command as follows:
>
> ```sh
> export VALIDATOR_KEYPAIR=$(cargo run --bin genx -- test-ledger ./ledger)
> ```
>
> Then we can just do the following to replay the ledger (i.e. providing the above config):
>
> ```sh
> cargo run --release ledger-config.toml
> ```

## Greptile Summary

This PR adds a new `test-ledger` command to the genx CLI tool that enables ledger replay testing without requiring the original validator keypair, improving testing and diagnostic capabilities.

- Added `test-ledger` command in `tools/genx/src/main.rs` that overwrites validator keypair with a random one for testing
- Added `ledger_replay_test.rs` module to handle keypair generation and file management for test scenarios
- Updated README with clear documentation and examples for using the new test-ledger functionality
- Builds on previous PR (#300) which added progress logging for ledger replay operations


GabrielePicco:
Closing as it doesn't fix the issue. We will re-open with a proper fix for the replay

Issue #336, opened 2025-03-31
bug: failing to start validator if there is no available snapshot


GabrielePicco:
## Description

Validator will segfault on start if there is a ledger, but the validator was previously killed before the first snapshot was taken

PR #347, opened 2025-04-23
fix(accountsdb): use current slot during rollback if off by one

bmuddha:
In case if the requested slot, to rollback to, is older than the current slot just by one, then return the current slot. This solves the bug when we try to rollback to older slot when no snapshots exists and validator fails to start. Additionally this is an expected scnerio during validator restarts, since ledger is always behind accountsdb by at least one slot.

closes gh-336

## Greptile Summary

This PR modifies the accounts database rollback behavior to handle cases where the requested rollback slot is only one slot behind the current slot, preventing unnecessary failures during validator restarts.

- Modified `ensure_at_most` in `magicblock-accounts-db/src/lib.rs` to allow rollback to current slot if requested slot is one behind (`slot >= self.slot() - 1`)
- Addresses issue where validator fails to start when ledger is naturally one slot behind accountsdb
- Improves startup reliability by handling expected one-slot difference between ledger and accountsdb
- Maintains data consistency while reducing unnecessary rollback attempts

### DIFF
--- a/magicblock-accounts-db/src/lib.rs
+++ b/magicblock-accounts-db/src/lib.rs
@@ -315,7 +315,7 @@ impl AccountsDb {
             .snapshot_engine
             .try_switch_to_snapshot(slot)
             .inspect_err(log_err!(
-                "switching to snapshot befor slot {}",
+                "switching to snapshot before slot {}",
                 slot
             ))?;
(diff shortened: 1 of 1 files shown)


thlorenz (approved):
LGTM nice catch!

PR #356, opened 2025-05-02
Bmuddha/fix/limit num monitoring accounts

bmuddha:


thlorenz (changes requested):
LGTM in general.
It'd be nice to have a test to confirm that accounts get properly removed + we can then send a tx that uses them and they get properly re-cloned without anything getting confused.

Also calling those accounts _undelegated_ is a bit misleading as it sounds like those are accounts that were delegated before and then undelegated.
I think a better term is _readonly_ accounts since that is what they are to the delegator no matter if they were ever delegated or not.

GabrielePicco (approved):
LGTM!

magicblock-account-updates/src/remote_account_updates_shard.rs
```diff
@@ -112,7 +120,16 @@ impl RemoteAccountUpdatesShard {
                     }
                 }
                 // When we receive a message to start monitoring an account
-                Some(pubkey) = self.monitoring_request_receiver.recv() => {
+                Some((pubkey, unsub)) = self.monitoring_request_receiver.recv() => {
+                    if unsub {
```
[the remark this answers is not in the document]
GabrielePicco: @thlorenz Agreed, that is a bit confusing, but we refer to those accounts as `undelegated` in several places. I think readonly is even more confusing, but we could rename them to `NotDelegated` everywhere in a follow-up PR.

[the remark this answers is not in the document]
bmuddha: renamed to not delegated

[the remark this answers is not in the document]
bmuddha: added the integration test, fixed the bug which was caught during the test

test-integration/configs/accounts/account1.json
```diff
@@ -0,0 +1,14 @@
+{
```
thlorenz, line 1: If possible name those accounts such that we have some context what they are used for, i.e. what do they contain and/or which test they serve.

test-integration/test-cloning/tests/02_test_monitored_accounts_limit.rs
```diff
@@ -0,0 +1,79 @@
+use integration_test_tools::IntegrationTestContext;
```
thlorenz, line 1: Nice! Where you able to inspect from the logs that the evictions took place?
The test has no way of checking that afaik unless it was to query the relevant metrics.

thlorenz (approved):
LGTM, but maybe we can rename those account files a bit if it make sense.

test-integration/test-cloning/tests/02_test_monitored_accounts_limit.rs
bmuddha: it was removed from accountdb, that's for sure, test does check for it, and removal from accountsdb can only take place after eviction. I did check metrics manually, don't see much point in adding extra logic to parse metrics from the tests as well, thus complicating it.

test-integration/configs/accounts/account1.json
bmuddha: done, renamed to non-delegated-clonable-accountX

PR #360, opened 2025-05-09
feat: Cleaner gasless implementation

GabrielePicco:
## Summary

This PR simplifies the gasless setup by removing the automatic airdrop and introducing a native way to execute transactions without charging the payer.

- https://github.com/magicblock-labs/magicblock-svm
- Changed the `solana-svm` dependency to use a patched version that allows signers without lamports
- Added a test case in `transaction_execute.rs` to verify transactions with non-existent fee payers and zero fees
- Simplified and modified genesis to set lamports per signature from the config (base fees)
- Set the default lamports per signature to 0

## Notes

- A follow-up PR will completely remove the initial payer SOL/lamports configuration and logic
- A follow-up PR will address rent


Cargo.toml
```diff
@@ -150,7 +150,7 @@ solana-rpc = "2.2"
 solana-rpc-client = { version = "2.2" }
 solana-rpc-client-api = { version = "2.2" }
 solana-sdk = { version = "2.2" }
-solana-svm = { version = "2.2", features = [ "dev-context-only-utils" ] }
+solana-svm = { git = "https://github.com/magicblock-labs/magicblock-svm.git", rev = "e93eb57", features = [ "dev-context-only-utils" ] }
```
bmuddha, line 153: Would be nice to see the introduced changes to the SVM as well.

bmuddha (approved):
Looks quite neat, too neat to be true even, wonder why we haven't considered this approach before.

taco-paco (approved):

Dodecahedr0x (approved):

PR #367, opened 2025-05-15
fix: allow specification of multiple rpc endpoints

bmuddha:
This fix introduces a breaking change to configuration, where instead of a single remote, it's now possible to specify multiple. Only the first one will be used for HTTP requests, but all of them will be used to load balance websocket subscriptions.

### DIFF
--- a/magicblock-accounts/src/accounts_manager.rs
+++ b/magicblock-accounts/src/accounts_manager.rs
@@ -36,7 +36,7 @@ impl AccountsManager {
         validator_keypair: Keypair,
         config: AccountsConfig,
     ) -> AccountsResult<Self> {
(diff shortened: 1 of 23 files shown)


GabrielePicco (commented):
Why instead not adding to [Cluster](https://github.com/magicblock-labs/magicblock-validator/blob/5da3bf9b6999d64e68dc594c9616450abfd03a02/magicblock-mutator/src/cluster.rs#L17) a `CustomWithMultipleWs` to avoid a breaking change in the config? 
Also the nested lists looks unnecessary

bmuddha:
closed in favor of https://github.com/magicblock-labs/magicblock-validator/pull/372

PR #373, opened 2025-05-27
PubSub redesign from broadcast to multicast 

bmuddha:
This PR reworks the way events are delivered to websocket subscription handlers. 
The main changes:
1. Every subscriber registers itself in the global subscriptions database, along with event that it's interested in
2. Geyser subsystem, upon receiving a particular event checks whether there're active subscribers for the event, and sends it only to those, no filtering is required further.
3. Rust-Protobuf-Rust conversions are disabled to avoid unnecessary overhead
4. gRPC plugin is completely disabled to avoid unnecessary overhead

## Greptile Summary

This PR implements a significant architectural change to the PubSub system, transitioning from a broadcast model to a multicast model for more efficient event delivery. The changes focus on targeted message delivery through a global subscriptions database.

- Introduces `SubscriptionsDb` in `magicblock-geyser-plugin/src/types.rs` for centralized subscription management using concurrent hashmaps
- Removes gRPC and protobuf conversion overhead by removing `conversions.rs` and simplifying message handling
- Adds new `notification_builder.rs` with trait-based notification system for different event types
- Reduces default channel capacity from 250,000 to 1,024 in `config.rs`, which may impact high-throughput scenarios
- Adds comprehensive test suite in `test-integration/test-pubsub/` validating subscription functionality for accounts, logs, signatures, slots and programs


GabrielePicco (approved):
LGTM!

magicblock-accounts-db/src/lib.rs
```diff
@@ -265,8 +265,8 @@ impl AccountsDb {
 
         let used_storage = self.storage.utilized_mmap();
         if let Err(err) = self.snapshot_engine.snapshot(slot, used_storage) {
-            error!(
-                "error taking snapshot at {}, slot {slot}: {err}",
+            warn!(
```
thlorenz, line 268: Isn't that a serious issue and should be logged on `error`? Or is it expected/ok to happen at times?

bmuddha: expected to happen during tests

PR #386, opened 2025-06-04
Patch tests from failing on subscription due to rate limits

taco-paco:
## Greptile Summary

Modified error handling and test configuration in RemoteAccountUpdatesWorker to address 'too many requests' errors during testing.

- Added conditional compilation (`#[cfg]`) in `magicblock-account-updates/src/remote_account_updates_worker.rs` to panic on shard failures during tests while logging errors in production
- Reduced WebSocket connections from 2 to 1 in `magicblock-account-updates/tests/remote_account_updates.rs` to prevent rate limiting
- Added `env_logger = "0.11"` to `magicblock-account-updates/Cargo.toml` for improved test logging
- Consider adding retry mechanism or connection pooling for more robust testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3565,6 +3565,7 @@ version = "0.1.3"
 dependencies = [
  "bincode",
  "conjunto-transwise",
+ "env_logger 0.11.6",
  "futures-util",
  "log",
  "magicblock-metrics",
--- a/magicblock-account-updates/Cargo.toml
+++ b/magicblock-account-updates/Cargo.toml
@@ -24,4 +24,4 @@ thiserror = { workspace = true }
 
 [dev-dependencies]
 test-tools = { workspace = true }
-env_logger = "0.11"
+env_logger = { workspace = true }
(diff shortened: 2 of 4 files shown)


thlorenz (commented):
This is a workaround getting rate limited, but makes it more likely that we'll miss an update of a cloned account.
I did not write this implementation, but from what I understand the shards were added to make this much less likely by basically listening on multiple connections for changes.
I'm not a huge fan of that solution and was hoping to improve on it when rewriting the cloning pipeline, but for now we have to be careful since this could introduce a regression.

Dodecahedr0x (approved):
Overall LGTM

thlorenz (approved):
So yeah we lighten the load on the _plain_ solana devnet by using less shards during tests (which we don't do during prod),but in prod we also don't use _plain_ devnet/mainnet anyways, so LGTM

GabrielePicco (approved):
LGTM as a patch and let's properly remove devnet dependency in a follow-up PR as discussed

PR #374, opened 2025-05-28
Log validator version on startup

lucacillario:
## Greptile Summary

Added version logging functionality to display both package version and Git commit information during validator startup process.

- Added `git-version` dependency in `test-bins/Cargo.toml` using workspace inheritance
- Implemented `GIT_VERSION` constant using `git_version::git_version!()` macro in `test-bins/src/rpc.rs`
- Added version logging in startup message: `🏷️ Validator version: {} (Git: {})`

<sub>💡 (1/5) You can manually trigger the bot by mentioning @greptileai in a comment!</sub>

PR #390, opened 2025-06-12
Log validator version on startup

GabrielePicco:
See https://github.com/magicblock-labs/magicblock-validator/pull/374 by @lucacillario 

## Greptile Summary

Added version logging on validator startup with both Cargo package version and Git version information, along with minor string formatting improvements.

- Added `git-version` dependency in `test-bins/Cargo.toml` for Git metadata tracking
- Introduced `GIT_VERSION` constant using `git_version::git_version!()` macro in `test-bins/src/rpc.rs`
- Enhanced startup logging to display version information for better observability
- Updated string formatting in `magicblock-metrics/src/metrics/types.rs` to use modern string interpolation syntax

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10059,6 +10059,7 @@ version = "0.1.2"
 dependencies = [
  "console-subscriber",
  "env_logger 0.11.6",
+ "git-version",
  "log",
  "magicblock-api",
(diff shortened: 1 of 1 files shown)


thlorenz (approved):
Nice!

PR #393, opened 2025-06-17
Optimizing column entries bookkeeping

taco-paco:
Instead of setting counter to DiRTY_COUNT on every write as before, we increase the counter if it is not "dirty", otherwise just skip.

This improves `init_system_metrics_ticker` since before we always had to recount entries as they were set to DIRTY_COUNT on every write

## Greptile Summary

Optimizes column entry bookkeeping in the ledger database by replacing the DIRTY_COUNT-based tracking with an atomic counter system, significantly improving `init_system_metrics_ticker` performance.

- Added `entry_counter: AtomicI64` to `LedgerColumn` struct in `magicblock-ledger/src/database/ledger_column.rs` for efficient entry tracking
- Removed redundant atomic counter fields from Ledger struct in `magicblock-ledger/src/store/api.rs`, moving tracking to column families
- Implemented incremental counter updates with `try_increase_entry_counter()` instead of marking entries as dirty on every write
- Added proper counter decrease logic in `delete_slot_range` to maintain accuracy during deletions

### DIFF
--- a/magicblock-api/src/tickers.rs
+++ b/magicblock-api/src/tickers.rs
@@ -160,13 +160,9 @@ pub fn init_system_metrics_ticker(
     fn set_accounts_count(bank: &Bank) {
         metrics::set_accounts_count(bank.accounts_db.get_accounts_count());
     }
+
     let ledger = ledger.clone();
     let bank = bank.clone();
-    try_set_ledger_storage_size(&ledger);
-    set_accounts_storage_size(&bank);
-    set_accounts_count(&bank);
-    try_set_ledger_counts(&ledger);
-
     tokio::task::spawn(async move {
         loop {
             tokio::select! {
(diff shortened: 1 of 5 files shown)


GabrielePicco (approved):
LGTM!

magicblock-api/src/tickers.rs
```diff
@@ -160,13 +160,9 @@ pub fn init_system_metrics_ticker(
     fn set_accounts_count(bank: &Bank) {
         metrics::set_accounts_count(bank.accounts_db.get_accounts_count());
     }
+
     let ledger = ledger.clone();
     let bank = bank.clone();
-    try_set_ledger_storage_size(&ledger);
```
thlorenz, line 165: This should run at startup as otherwise we have not metrics at all until the first interval elapses.

thlorenz: Agreed that we can wait a few mins for the metrics to make startup faster.

PR #397, opened 2025-06-17
fix: make the max number of ws-connections configurable

bmuddha:
Introduce new paramater `max-ws-connections` to the [rpc] section of the config, by default this value is set to 16384, but setting it to anything below 200K is also fine, anything above that value will create significant memory pressure by slab preallocation, and can even trigger an OOM kill. 

 

## Greptile Summary

Adds configurable WebSocket connection limits to prevent memory exhaustion, with a new `max-ws-connections` parameter in the RPC config defaulting to 16384 connections.

- Added `max_ws_connections` field to `magicblock-pubsub/src/pubsub_service.rs` to control concurrent WebSocket connections
- Modified `magicblock-config/src/rpc.rs` to expose the new config parameter with default of 16384
- Added validation to keep connections under 200K to prevent slab preallocation memory pressure
- Updated config test files to verify proper parsing and defaults
- Integrated limit through PubsubConfig initialization in `magicblock-api/src/magic_validator.rs`

### DIFF
--- a/magicblock-pubsub/src/pubsub_service.rs
+++ b/magicblock-pubsub/src/pubsub_service.rs
@@ -27,12 +27,18 @@ use crate::{
 #[derive(Clone)]
 pub struct PubsubConfig {
     socket: SocketAddr,
+    max_connections: usize,
 }
 
 impl PubsubConfig {
-    pub fn from_rpc(rpc_addr: IpAddr, rpc_port: u16) -> Self {
+    pub fn from_rpc(
+        rpc_addr: IpAddr,
+        rpc_port: u16,
+        max_connections: usize,
+    ) -> Self {
         Self {
             socket: SocketAddr::new(rpc_addr, rpc_port + 1),
+            max_connections,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


GabrielePicco (approved):
LGTM!

magicblock-pubsub/src/pubsub_service.rs
```diff
@@ -89,6 +96,10 @@ impl PubsubService {
             |context: &RequestContext| Arc::new(Session::new(context.sender()));
 
         ServerBuilder::with_meta_extractor(self.io, extractor)
+            // NOTE: we just set the max number of allowed connections to a reasonably high value
+            // to satisfy most of the use cases, however this number cannot be arbitrarily large
+            // due to the preallocation involved, and a large value will trigger an OOM Kill
+            .max_connections(self.config.max_connections)
```
thlorenz, line 102: Should we run a verification step over this config value to ensure that the max reasonable value isn't exceeded?

thlorenz (approved):
LGTM, but maybe add the max value check before merging.

magicblock-pubsub/src/pubsub_service.rs
bmuddha: No need, it will OOM anyway. Ideally we need custom pubsub implementation which doesn't rely on the weirdness of the third party lib

Issue #58, opened 2024-06-18
Add API for accessing an accurate clock value inside SVM contracts


crypto-vincent:
# Summary

In order to implement crons, and real-time applications smart contract developers will need to be able to measure time deltas.

# Details

A simple example would be in a physics game where each player is affected by gravity every 10ms, the "gravity" instruction might want to verify that at least 10ms have elapsed since the last transaction call. Or it might want to measure exactly "how many" of those 10ms tick have elapsed in order to compute each player's movement.

PR #398, opened 2025-06-17
Add API for accessing an accurate clock value inside SVM contracts 

batmnnn:
## What This PR Does
This solves issue #58

This PR adds a new time measurement library that lets the Magicblock validator handle timing-based tasks like cron jobs and real-time processing. Think of it as a stopwatch that can tell you exactly how much time has passed and whether it's time to do something.

## Why We Need This

Smart contract developers need to measure time differences for things like:
- A physics game where gravity affects players every 10ms
- Scheduled maintenance tasks that run every few minutes  
- Rate limiting to prevent spam requests
- Health checks that happen periodically

Before this, there wasn't a clean way to track time intervals and know exactly when to trigger these events.

### Core Library (magicblock-time-delta)
- **TimeDelta**: Tracks time intervals with millisecond precision
- **SolanaTimeDelta**: Same thing but uses blockchain time instead of system time
- Super simple API - just create a timer, check if it's time to tick, and reset it

### Real Examples That Actually Work
- **Validator monitoring**: Health checks every 5 seconds (catches up if it misses any)
- **Rate limiting**: Only allow 3 requests per second, reject the rest
- **Blockchain timing**: Track slot progression using Solana's clock
- **Cron scheduler**: Framework for running maintenance tasks

### Key Features
- ✅ Millisecond accuracy
- ✅ Handles missed intervals (if something takes longer than expected)
- ✅ Works with both system time and blockchain time
- ✅ Zero memory allocations in the hot path (performance!)
- ✅ Comprehensive tests (31 tests, all passing)

## How It Works (Simple Example)

```
// Create a timer for 10ms intervals (like your gravity example)
let mut timer = TimeDelta::new(10);

// Later, when processing:
if timer.should_tick() {
    let ticks = timer.tick_count(); // How many 10ms periods passed?
    
    // Apply gravity for each missed tick
    for _ in 0..ticks {
        apply_gravity();
    }
    
    timer.update(); // Reset the timer
}
```

## Testing

Run the examples to see it in action:
```
cargo run --example validator_usage -p magicblock-time-delta
cargo run --example cron_scheduler -p magicblock-time-delta
```

The output shows real timing behavior - health checks catching up after delays, rate limiting blocking excess requests, and blockchain timing tracking slot progression.

## Impact

This gives validator developers the timing primitives they need without having to reinvent time measurement. It's designed specifically for the use cases mentioned in the original requirements and integrates cleanly with the existing codebase.

The implementation is production-ready with proper error handling, comprehensive tests, and examples that demonstrate real-world usage patterns for validator operations.

## Greptile Summary

Added a new time measurement library (`magicblock-time-delta`) that provides millisecond-precision timing capabilities for smart contracts and validator operations.

- Introduces `TimeDelta` and `SolanaTimeDelta` types in `magicblock-time-delta/src/lib.rs` for system and blockchain time tracking
- Added example implementations in `magicblock-time-delta/examples/` demonstrating health checks, rate limiting, and cron scheduling
- Added core timing primitives in `magicblock-core/src/time_delta.rs` for SVM contract integration
- Implementation includes zero-allocation hot path design and comprehensive test coverage (31 tests)
- Library enables critical timing features like physics simulations, maintenance scheduling, and rate limiting with millisecond accuracy


batmnnn:
@GabrielePicco is this right ?

PR #407, opened 2025-06-23
feat: account hydration on background task

thlorenz:
- **deps: add flume properly**
- **feat: put account hydration on a background task**


bmuddha (approved):
LGTM!

PR #414, opened 2025-06-25
Fixed graceful shutdown

taco-paco:
Right now our validator doesn't handle SIGTERM, only SIGINT. 
This pr adds handling of SIGTERM signal

## Greptile Summary

Enhanced validator shutdown handling by adding SIGTERM support alongside SIGINT, implementing a dedicated shutdown module for more robust signal management and cleanup.

- Added new `test-bins/src/shutdown.rs` implementing platform-aware signal handling for both SIGTERM and SIGINT
- Modified `test-bins/src/rpc.rs` to use the new Shutdown type for improved termination management
- Implemented proper error handling for shutdown failures
- Added platform-specific handling using `cfg` attributes to support both Unix and non-Unix systems

### DIFF
--- a/test-bins/src/shutdown.rs
+++ b/test-bins/src/shutdown.rs
@@ -1,0 +1,35 @@
+use std::io;
+
+use log::info;
+use tokio::{signal, signal::unix::SignalKind};
+
+pub struct Shutdown;
+impl Shutdown {
+    pub async fn wait() -> Result<(), io::Error> {
+        #[cfg(unix)]
+        return Self::wait_unix().await;
+        #[cfg(not(unix))]
+        return Self::wait_other().await;
+    }
+
+    #[cfg(unix)]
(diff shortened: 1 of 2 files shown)


thlorenz (approved):
LGTM

test-bins/src/shutdown.rs
```diff
@@ -0,0 +1,35 @@
+use std::io;
+
+use log::info;
+use tokio::{signal, signal::unix::SignalKind};
+
+pub struct Shutdown;
+impl Shutdown {
+    pub async fn wait() -> Result<(), io::Error> {
```
bmuddha, line 8: nit: can't we just have a single function, without extra calls? Just having two `wait` methods with different cfgs?

GabrielePicco (approved):
LGTM!

PR #455, opened 2025-07-15
chore: avoid CU budget exceeded for lookup table management

thlorenz:
## Summary

Multiply all current budget values by `7` in order to avoid CU budget exceeded issues when
managing lookup tables.

## Greptile Summary

This PR addresses compute unit (CU) budget exceeded errors in lookup table management operations by introducing a safety multiplier of 7x to all CU budget values. The change affects create/extend, extend, deactivate, and close table operations in the table-mania module.

Key changes:
- Introduces `SAFETY_MULTIPLIER: u32 = 7`
- Updates compute unit budgets for table operations:
  - Create and extend: ~18,550 CUs (up from ~2,650)
  - Extend: 8,400 CUs (up from 1,200)
  - Deactivate: 7,350 CUs (up from 1,050)
  - Close: 7,350 CUs (up from 1,050)

The PR also updates test assertions to use upper bounds instead of exact matches for CU consumption, making tests more resilient to variations in actual CU usage.

## Confidence score: 4/5

1. This PR is safe to merge as it conservatively increases compute budgets based on empirical observations
2. The score of 4 reflects that while the changes are well-tested and solve an immediate issue, the 7x multiplier is somewhat arbitrary and might be more than necessary
3. Files needing attention:
   - magicblock-table-mania/src/compute_budget.rs - verify if 7x multiplier is the optimal value

<sub>2 files reviewed, no comments</sub>
<sub>[Edit PR Review Bot Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=magicblock-validator_455)</sub>

### DIFF
--- a/magicblock-table-mania/src/compute_budget.rs
+++ b/magicblock-table-mania/src/compute_budget.rs
@@ -2,6 +2,14 @@ use solana_sdk::{
     compute_budget::ComputeBudgetInstruction, instruction::Instruction,
 };
 
+/// We multiply the CU values that we determined since we've seen compute budget exceeded errors
+/// if we use them as is. The only way to determine CUs for a particular transaction 100% is
+/// to simulate that exact transaction.
+/// This would make table management even slower, but is an option to consider for the future.
+/// The multiplier is based on the observation that [CREATE_AND_EXTEND_TABLE_CUS] were _safe_ at
+/// 16K-17K CUs which we slightly exceed if we multiply by this multiplier.
+const SAFETY_MULTIPLIER: u32 = 7;
+
 /// Compute units required to create and extend a lookup table, with the initial
 /// pubkeys. This is the same no matter how many pubkeys are added to the table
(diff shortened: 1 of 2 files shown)


Dodecahedr0x:
LGTM, but if I was the only one having the issue, doesn't that mean something might be off on my end that I need to fix? Also, I avoided the issue by just increasing the Create and extend CUs, do we really need to increase all other costs?

thlorenz:
> LGTM, but if I was the only one having the issue, doesn't that mean something might be off on my end that I need to fix? Also, I avoided the issue by just increasing the Create and extend CUs, do we really need to increase all other costs?

As I said in slack I ran into similar issues on `Close` table during some ix tests run as well, so I think it is wise for now to just give it a bit of leeway so we don't run into any issues in prod.

PR #458, opened 2025-07-15
feat(ledger): remove requirement to use the same keypair during replay

### LINKED ISSUE #299: feat: Remove requirement to use the same keypair during ledger replay

Dodecahedr0x:
Closes #299 

Add a parameter in the configuration to disable verification of the validator's pubkey during replay, enabling anyone to replay the ledger


Dodecahedr0x:
Making this PR a draft while I work on cloning tests

GabrielePicco (approved):
LGTM!

Dodecahedr0x:
Fixed git conflicts

PR #473, opened 2025-07-18
feat(cloning): make cloning concurrency configurable

### LINKED ISSUE #275: feat: make concurrency of account cloning during ledger replay configurable
Currently we hardcode `10` for this, but most clients will have RPCs that support higher concurrency without hitting rate limits. Thus we should make it configurable, i.e.:

```toml
[ledger]
reset = false
path = "/some/path"
account-clone-concurrency = 20
```

Dodecahedr0x:
Closes #275 

It introduces a parameter to choose the degree of concurrency for cloning accounts on hydration. The default stays the same, but this can now be adjusted

## Snippets

In TOML configs:

```toml
[ledger.replay]
account-hydration-concurrency = 10
```

With the CLI:

```bash
cargo run --release -- --replay-account-hydration-concurrency 10
REPLAY_ACCOUNT_HYDRATION_CONFIG=10 cargo run --release
```


thlorenz (changes requested):
I'm not sure if this belongs into `AccountsCloneConfig` since it is related to ledger replay.
We already have a `LedgerConfig` and I'd add a `field named either of the following there: `account_hydration_concurrency|clone_concurrency|hydration_clone_concorrency` (or if you have a better name).

We could either add that directly or (preferred) in a `LedgerReplayConfig`, so the toml would look similar to.

```toml
[ledger]

replay = { account_hydration_concurrency = 20 }
```


Also please add a fixture [here](http://github.com/magicblock-labs/magicblock-validator/blob/dev/magicblock-config/tests/fixtures/) and a test that shows we can parse these new configs from toml properly.

Issue #303, opened 2025-02-17
feat: Periodically or on start run transaction to claim fees


GabrielePicco:
## Description

Validator needs to explicitly claim and collect fees

## Proposed solution

We need to build and send the instruction to claim the fees when the validator starts and/or periodically

https://github.com/magicblock-labs/delegation-program/blob/main/tests/integration/tests/test-delegation.ts#L47

PR #475, opened 2025-07-18
feat: Add startup fee claiming with secure keypair handling

BretasArthur1:
Implements fee claiming on validator startup per Issue #303. 

- Adds `claim_fees method` to build/send delegation program's ValidatorClaimFees instruction
- Uses Arc<Keypair> for safe signing without clones.
- includes comments for clarity. No periodicity yet – can add if needed. 
- Fixes security cloning issue.

### DIFF
--- a/magicblock-api/src/magic_validator.rs
+++ b/magicblock-api/src/magic_validator.rs
@@ -106,7 +106,6 @@ use crate::{
         init_system_metrics_ticker,
     },
 };
-
 // -----------------
 // MagicValidatorConfig
 // -----------------
@@ -168,6 +167,7 @@ pub struct MagicValidator {
     geyser_rpc_service: Arc<GeyserRpcService>,
     pubsub_config: PubsubConfig,
     pub transaction_status_sender: TransactionStatusSender,
+    validator_keypair: Arc<Keypair>,
 }
 
 impl MagicValidator {
@@ -371,7 +371,7 @@ impl MagicValidator {
             config.validator_config.rpc.port,
             config.validator_config.rpc.max_ws_connections,
         );
-        validator::init_validator_authority(identity_keypair);
+        validator::init_validator_authority(identity_keypair.insecure_clone());
 
         // Make sure we process the ledger before we're open to handle
         // transactions via RPC
@@ -414,6 +414,7 @@ impl MagicValidator {
             accounts_manager,
             transaction_listener,
             transaction_status_sender,
+            validator_keypair: Arc::new(identity_keypair),
         })
     }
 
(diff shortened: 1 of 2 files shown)


magicblock-api/src/magic_validator.rs
```diff
@@ -698,6 +699,87 @@ impl MagicValidator {
...
+        let (validator_fees_vault, _) = Pubkey::find_program_address(
+            &[b"v-fees-vault", validator.as_ref()],
+            &delegation_program_id,
+        );
+
+        // build the instruction with the ValidatorClaimFees discriminator
+        // data[0] is the discriminator and data[8] is the option<u64> serialized as 0, meaning None so it will claim all the fees from the fees vault
+        let data = vec![7, 0, 0, 0, 0, 0, 0, 0, 0];
+
+        let ix = Instruction {
```
Dodecahedr0x, line 743: This and the above can be replaced by using the [instruction builder directly](https://github.com/magicblock-labs/delegation-program/blob/9570abf24ab6d03ba75ecdbf8a6a787bb0d69d3a/src/instruction_builder/validator_claim_fees.rs#L11)

magicblock-api/src/magic_validator.rs
```diff
@@ -706,6 +788,8 @@ impl MagicValidator {
             }
         }
 
+        self.claim_fees().await?; //this will block the main thread should we spawn a new task?
```
Dodecahedr0x, line 791: Claiming only fees here is not ideal because you need to restart to claim fees. Having a claim service scheduled would help

GabrielePicco: Spawning a new task that periodically claims fees would be ideal. It would also avoid awaiting during startup, which is critical since any delay results in downtime.

GabrielePicco (commented):
Thanks @BretasArthur1 for the contribution. It looks good to me overall. Please address the comments and it's almost ready to be merged.

magicblock-api/src/magic_validator.rs
```diff
@@ -849,6 +921,10 @@ impl MagicValidator {
         self.token.cancel();
         self.ledger_truncator.stop();
 
+        if let Some(handle) = self.claim_fees_task.take() {
```
thlorenz, line 924: Once you added a `stop` method to the `claim_fees_task` just call it here without even having to know that it is an `Option`.

thlorenz (changes requested):
Thanks for adding this feature.
Just needs some refactoring and nits cleaning, then should be good to go.
Biggest change needed is to move this into a separate crate/module like I explain [here](https://github.com/magicblock-labs/magicblock-validator/pull/475#discussion_r2227760014)

Great work!

BretasArthur1:
Thanks for the review @thlorenz ! I will address the comments and push it. Nice review btw, separating into another crate was something I want to do

BretasArthur1:
Hey @thlorenz thanks! Seems that this last one is actually a little bit trickier than I thought I think it's something around timing issues when starting the validator. But sgtm, I will be here if needed!

GabrielePicco (approved):
LGTM! 🎉

PR #476, opened 2025-07-18
feat(config): derive merge

### LINKED ISSUE #471: fix: merge all flattened structs in the config
Nested flattened structs in the config must implement and use a merge function to avoid overwriting non-default fields.

Example: if any field in `accounts.db` is non-default in `self`, all fields of `other.accounts.db` will be ignored, even if non-default and non-overlapping.

Ideally, the merge function should be derived (using a macro or another crate) to avoid having to modify it every time we add a field

Dodecahedr0x:
Closes #471 

Create a `Mergeable` derive trait that automatically merges the fields with default values with the non-default fields of another instance. It also handles nested Mergeable structs.

Note: in order to handle nested Mergeable structs, it uses the heuristic that nested mergeable struct have a type name containing `Config`. This is not a very robust method, but it works as we only use it for internal config structs.

### DIFF
--- a/magicblock-config-macro/src/lib.rs
+++ b/magicblock-config-macro/src/lib.rs
@@ -199,10 +199,10 @@ pub fn clap_from_serde(_attr: TokenStream, item: TokenStream) -> TokenStream {
 /// impl Merge for SomeOtherConfig {
 ///     fn merge(&mut self, other: Self) {
 ///         let default = Self::default();
-///         if self.field1 == default.field1 && other.field1 != default.field1 {
+///         if self.field1 == default.field1 {
 ///             self.field1 = other.field1;
 ///         }
-///         if self.field2 == default.field2 && other.field2 != default.field2 {
+///         if self.field2 == default.field2 {
 ///             self.field2 = other.field2;
 ///         }
 ///     }
@@ -217,7 +217,7 @@ pub fn clap_from_serde(_attr: TokenStream, item: TokenStream) -> TokenStream {
 /// impl Merge for MyConfig {
 ///     fn merge(&mut self, other: Self) {
 ///         let default = Self::default();
-///         if self.field1 == default.field1 && other.field1 != default.field1 {
+///         if self.field1 == default.field1 {
 ///             self.field1 = other.field1;
 ///         }
 ///         self.field2.merge(other.field2);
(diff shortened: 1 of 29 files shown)


magicblock-config-macro/src/merger.rs
```diff
@@ -0,0 +1,92 @@
...
+            if Self::type_has_merge_method(field_type) {
+                // If the field type likely has a merge method, call it
+                quote! {
+                    self.#name.merge(other.#name);
+                }
+            } else {
+                // Otherwise, use the current default-based merging logic
+                // Use .clone() to avoid move issues with non-Copy types
+                quote! {
+                    if self.#name == default.#name && other.#name != default.#name {
```
thlorenz, line 72: As I mentioned in slack already, this is more than needed. If `self.field == default.field` then the assignment is a noop anyways if `other.field == default.field`. So that second condition is not needed.

magicblock-config-macro/src/lib.rs
```diff
@@ -158,3 +161,73 @@ pub fn clap_from_serde(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// #[derive(Default)]
+/// struct SomeOtherConfig {
+///     field1: u32,
+///     field2: String,
+/// }
+///
+/// impl Merge for SomeOtherConfig {
+///     fn merge(&mut self, other: Self) {
+///         let default = Self::default();
+///         if self.field1 == default.field1 && other.field1 != default.field1 {
```
thlorenz, line 202: Make sure to update these docs again after removing the superfluous second condition.

Also BTW some tests that actually check the code that is rendered using `quote` (kind of the opposite of what I'm doing [in shank](https://github.com/metaplex-foundation/shank/blob/master/shank-macro-impl/src/parsed_enum/parsed_enum.rs#L143) would make this a living documentation that will fail once we update the implementation.

thlorenz (approved):
LGTM, but please remove the superfluous if condition as mentioned.
I'm pre-approving since all else is good.

If you feel unsure that the 2nd condition is not needed then rely on your tests (possibly adding more) to prove that it is correct.

magicblock-config-macro/src/merger.rs
Dodecahedr0x: Addressed in [1d25339](https://github.com/magicblock-labs/magicblock-validator/pull/476/commits/1d25339c9bf691dc370436ec45cfa548ae8cb0aa)

GabrielePicco (approved):
LGTM!

PR #483, opened 2025-07-28
feat: implemented request dispatchers

bmuddha:


thlorenz (approved):
The `AccountsToEnsure` piece looks ok to me even though using a `Box` maybe problematic since that wouldn't be thread-safe no?
Why not just clone the pubkeys instead of making this more complicated trying to optimize .. this takes gzillion times less time than actually ensuring those accounts.

The rest of the PR looks ok. I'd prefer really reviewing the whole thing once it is integrated and has test.
Also this PR has no description so it's hard to tell exactly what you did.

I think it's ok for you to develop this somehow in isolation (you might as well be in a separate repo) like I'm doing for the cloning pipeline and have more eyes on it when swapping this for the current RPC.

PR #495, opened 2025-08-06
test: use reduced snapshot frequency

### LINKED ISSUE #494: Use reduced snapshot frequency in tests
Some `restore_ledger` tests use the default snapshot frequency, causing them to wait for 25 seconds. This frequency is already configured to be lower, but it's not used.

Dodecahedr0x:
Closes #494 

`setup_offline_validator` already sets the snapshot frequency used by some tests to `2`, but tests are still using the default value. This updates tests to use the actual default value.


thlorenz (approved):
LGTM

PR #500, opened 2025-08-08
fix: robust tests

### LINKED ISSUE #498: Brittle integration tests
Integration tests are frequently failing (mostly restore_ledger tests). It sometimes appears that the ledger is not properly restored by the time the test starts

Dodecahedr0x:
Closes #498 

Do not claim fees on the first tick of the claim fee service. This should help solving tests flakyness

## Greptile Summary

This PR addresses test flakiness by modifying the claim fees service to skip its first tick upon startup. The change is made in `magicblock-validator-admin/src/claim_fees.rs` where the periodic fee claiming task is initialized.

The core modification replaces `tokio::time::interval()` with `tokio::time::interval_at()`, which allows specifying a custom start time. Instead of beginning execution immediately (which `interval()` does), the new implementation delays the first tick by one full tick period. This is achieved by calculating `start_time = Instant::now() + tick_period` and passing it to `interval_at()`.

This change fits into the broader MagicBlock validator architecture where various services use ticker patterns for periodic operations (similar to the slot ticker and commit accounts ticker patterns seen in the codebase). The claim fees service appears to be responsible for periodically claiming accumulated fees, and the immediate execution upon startup was causing timing conflicts during test execution, likely interfering with test setup or expected initial state.

The modification is surgical and maintains the same periodic behavior after the initial delay, preserving the service's core functionality while eliminating the problematic first-tick execution.

## Confidence score: 4/5

- This PR is safe to merge with minimal risk of breaking production functionality
- Score reflects a simple, well-targeted fix that addresses a specific timing issue without changing core logic
- No files require special attention as the change is contained to one function with clear intent

### DIFF
--- a/magicblock-rpc/src/json_rpc_request_processor.rs
+++ b/magicblock-rpc/src/json_rpc_request_processor.rs
@@ -841,9 +841,10 @@ impl JsonRpcRequestProcessor {
         signature: Signature,
         _search_transaction_history: bool,
     ) -> Option<TransactionStatus> {
+        // Looking back 30 seconds ensures tests are more robust
         let bank_result = self.bank.get_recent_signature_status(
             &signature,
-            Some(self.bank.slots_for_duration(Duration::from_secs(10))),
+            Some(self.bank.slots_for_duration(Duration::from_secs(30))),
         );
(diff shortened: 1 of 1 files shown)


GabrielePicco (approved):
LGTM!

PR #505, opened 2025-08-12
fix: CI flakyness

Dodecahedr0x:


magicblock-ledger/src/store/api.rs
```diff
@@ -753,14 +753,12 @@ impl Ledger {
                 }
             }
             None => {
-                let mut iterator = self
-                    .transaction_cf
-                    .iter_current_index_filtered(IteratorMode::From(
-                        (signature, highest_confirmed_slot),
-                        IteratorDirection::Forward,
-                    ));
+                let mut iterator = self.iter_transaction_statuses(
```
thlorenz, line 756: I'm not sure about this one as the above `get_transaction_status` call (line 745) also uses `transaction_status_cf` under the hood.
The method you are using here uses `iter_protobuf` under the hood which involves deserialization and thus is a lot more expensive, however then you ignore the status (_data).

I don't think the original implementation only iterated over successful transactions only like you say in the description, as it doesn't even get the status.

(Maybe I'm missing something here)

magicblock-ledger/src/store/api.rs
```diff
@@ -1005,21 +1003,21 @@ impl Ledger {
         )>,
     > + '_ {
         let (_lock, _) = self.ensure_lowest_cleanup_slot();
-        let iterator_mode = iterator_mode.unwrap_or(IteratorMode::Start);
+        let from = from.unwrap_or((Signature::default(), 0));
         self.transaction_status_cf
-            .iter_protobuf(iterator_mode)
+            .iter_protobuf(IteratorMode::From(from, IteratorDirection::Forward))
```
thlorenz, line 1008: Always starting at slot 0 makes this a lot slower.
The previous implementation would start at a specific slot.

Once the ledger is very large this would take a very long time. I'm not sure this should be needed.

magicblock-ledger/src/store/api.rs
```diff
@@ -1005,21 +1003,21 @@ impl Ledger {
...
                     Ok(((signature, slot), status)) => {
                         ((signature, slot), status)
                     }
                     Err(err) => return Some(Err(err)),
                 };
-                let include = status.err.is_none() == success;
-                if include {
-                    Some(Ok((slot, signature, status)))
-                } else {
+                let exclude = status.err.is_some() && success;
```
thlorenz, line 1016: This is the same just implemented in the reverse way. What does this improve?

test-integration/test-magicblock-api/tests/test_get_block_timestamp_stability.rs
```diff
@@ -1,25 +1,36 @@
...
+    native_token::LAMPORTS_PER_SOL, signature::Keypair, signer::Signer,
+};
+use solana_transaction_status::UiTransactionEncoding;
 
 const EPHEM_URL: &str = "http://localhost:8899";
 
 #[tokio::test]
 async fn test_get_block_timestamp_stability() {
-    let millis_per_slot = 50;
+    let rpc_client = RpcClient::new(EPHEM_URL.to_string());
```
thlorenz, line 13: I don't understand how this achieves the same as the previous code did.

What you should use instead are test utility methods like `wait_for_next_slot` or `wait_for_delta_slot` or `wait_for_slot_ephem` which capture the slot and return when it changes to where you want it to be.

thlorenz (changes requested):
As stated I'm not sure about some of the changes here.

Particular concerning are the changes to the ledger query methods which could affect perf negatively.

magicblock-ledger/src/store/api.rs
Dodecahedr0x: You're right performance would be worsened by this change. It's just that somehow, the old way of doing it sometimes returned no result when this way of iterating does. The issue probably comes from somewhere else, I'll keep looking

test-integration/test-magicblock-api/tests/test_get_block_timestamp_stability.rs
Dodecahedr0x: The test just makes sure that for a given transaction, `tx.block_time == block@slot.block_time == ledger@slot.block_time`. The tx time is only accessible once the slot is finalized so I wait some time and then do the verification

magicblock-ledger/src/store/api.rs
Dodecahedr0x: Addressed [41e37cc](https://github.com/magicblock-labs/magicblock-validator/pull/505/commits/41e37ccedf59e307b97dc33b0135a644c4e0bbbd)

thlorenz (approved):
LGTM now.

GabrielePicco (approved):
LGTM!

PR #508, opened 2025-08-15
feat: new resume strategies

Dodecahedr0x:
This PR changes `ResumeStrategy` and enables specifying the starting slot:
- `Replay` **becomes the default strategy**. It keeps accountsDB and the ledger, replays it, and starts from the last ledger slot.
- `Reset` discards the ledger and accountsDB and starts from the provided slot, 0 otherwise.
- `ResumeOnly` now keeps the ledger and accountsDB, does not replay the ledger, and starts from the last ledger slot

### Replay

As the default strategy, you can just leave the resume strategy field empty or make it explicit like so:

```toml
[ledger]
resume-strategy = { kind = "replay" }
```

### Reset

By default, the reset strategy removes accounts and starts at slot 0.

```toml
[ledger]
resume-strategy = { kind = "replay" }
```

You can change those defaults like so:

```toml
[ledger]
resume-strategy = { kind = "replay", reset-slot = 1000, keep-accounts = true }
```

### ResumeOnly

This strategy preserves the ledger and accounts but does not replay the ledger

```toml
[ledger]
resume-strategy = { kind = "resume-only" }
```

## Greptile Summary

This PR introduces a comprehensive refactoring of the ledger resume strategies in the MagicBlock validator, providing more granular control over how the validator restarts and handles its state components. The changes restructure the `LedgerResumeStrategy` enum from a simple binary choice to five distinct strategies, each with specific behaviors regarding ledger preservation, accountsDB handling, and slot management.

The key architectural changes include:

**New Resume Strategies:**
- `Replay` (now default): Preserves both ledger and accountsDB, replays transactions, starts from last ledger slot
- `Reset`: Discards both ledger and accountsDB, starts from specified slot or 0
- `DiscardResume`: Discards ledger but keeps accountsDB, starts from last ledger slot
- `ResumeOnly`: Keeps both components without replay, starts from last ledger slot
- `AccountsOnly`: Keeps only accountsDB, starts from specified slot or 0

**Enhanced Configuration:**
A new `starting_slot: Option<u64>` field is added to `LedgerConfig`, allowing precise control over where the validator begins processing. This enables scenarios like starting from a specific historical slot while preserving certain data structures.

**Core Implementation Changes:**
The `Bank::new` constructor receives a new `adb_init_slot_override` boolean parameter that controls whether the AccountsDB slot should be forcibly set to match the initialization slot. New utility methods (`is_resuming()`, `is_removing_ledger()`, `is_removing_accountsdb()`, etc.) provide fine-grained control over component cleanup during restart.

**Testing Infrastructure:**
A comprehensive test suite is added with individual tests for each resume strategy (`15_1_reset.rs` through `15_5_resume_only.rs`) and a unified testing framework in `resume_strategies.rs` that validates slot progression, account balance preservation, and transaction history availability based on each strategy's characteristics.

All existing integration tests are updated to accommodate the new function signatures, with most tests using `None` for the starting slot parameter to maintain existing behavior. The changes maintain backward compatibility while enabling more sophisticated validator restart scenarios, particularly useful for testing, debugging, and different deployment strategies.

## Confidence score: 4/5

- This PR involves significant architectural changes but appears well-implemented with comprehensive testing coverage
- Score reflects the complexity of state management changes and potential for subtle issues in different resume scenarios
- Pay close attention to the Bank initialization changes and the new AccountsDB slot override logic


GabrielePicco (approved):
LGTM!

magicblock-config/tests/read_config.rs
```diff
@@ -118,6 +118,7 @@ fn test_load_local_dev_with_programs_toml_envs_override() {
     env::set_var("VALIDATOR_FQDN", "magicblock.er.com");
     env::set_var("LEDGER_SIZE", "123123");
     env::set_var("LEDGER_RESUME_STRATEGY", "resume-only");
+    env::set_var("LEDGER_STARTING_SLOT", "1");
```
thlorenz, line 121: The naming of this is a bit tricky once you make it part of a variant, but since it is only valid for certain variant I'd name it something as follows:

- `LEDGER_RESUME_STRATEGY_RESET_SLOT` which still makes it very clear which path to take in the toml config to define it as well.
You can then easily detect if it was supplied for a conflicting resume stragegy when trying to construct the variant.

test-integration/test-ledger-restore/src/resume_strategies.rs
```diff
@@ -0,0 +1,129 @@
...
+    signature: &Signature,
+    slot: u64,
+    strategy: LedgerResumeStrategy,
+    starting_slot: Option<u64>,
+) -> Child {
+    let (_, mut validator, ctx) = setup_offline_validator(
+        ledger_path,
+        None,
+        None,
+        strategy.clone(),
```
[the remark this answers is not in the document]
thlorenz: Please run `cargo clippy` or `cargo clippy --fix --allow-dirty` to avoid all this. I had to clean some of those up after I pulled master recently.

We should probably add a check to CI in order to catch this on PRs.

test-integration/test-ledger-restore/tests/15_1_reset.rs
```diff
@@ -0,0 +1,7 @@
+use magicblock_config::LedgerResumeStrategy;
```
thlorenz, line 1: Just a nit, but why not name them `14_strategy_reset.rs, 15_strategy_...`, 16`?

test-integration/configs/claim-fees-test.toml
```diff
@@ -10,6 +10,9 @@ index-map-size = 2048576
 max-snapshots = 7
 snapshot-frequency = 1024
 
+[ledger]
```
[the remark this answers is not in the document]
Dodecahedr0x: @GabrielePicco mentioned that we do not want catastrophic cases where users end up deleting all their accounts and the ledger because they forgot to add the proper line in the config

test-integration/test-ledger-restore/tests/15_1_reset.rs
Dodecahedr0x: I chose these names because they're all variants of the same test. I initially had them all in a single file but the tests failed because they were parallelized and ephem ports were already used.

test-integration/test-ledger-restore/src/resume_strategies.rs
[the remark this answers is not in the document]
Dodecahedr0x: I think greptile is wrong: setup_offline_validator moves the config so we can't read it afterwards, hence the cloning. I could pass a ref and copy within but I believe it's a bad practice as it hides the cloning.

But definitely we should have the CI running on tests, will create the issue

test-integration/test-ledger-restore/tests/15_1_reset.rs
thlorenz: OK makes sense, just odd naming.

test-integration/configs/claim-fees-test.toml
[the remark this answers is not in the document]
thlorenz: OK makes sense

magicblock-config/src/ledger.rs
```diff
@@ -69,23 +76,40 @@ impl Default for LedgerConfig {
 #[strum(serialize_all = "kebab-case")]
 #[value(rename_all = "kebab-case")]
 pub enum LedgerResumeStrategy {
-    #[default]
     Reset,
+    DiscardResume,
```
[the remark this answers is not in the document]
thlorenz: I'm not sure if we need to bow to clap here as that would mean that we can never use data variants in our config.

There must be a way to express the same in a way that clap likes it, maybe you need an intermediate struct.
We have other configs that allow data variants.

What you could do is have a struct that has both a `type` and optional `slot` and configure that with clap.
Then when you convert it into the appropriate enum variant you ensure that the slot is `None` when it isn't valid.

At least in the rest of the code we operate on the much more expressive data variant instead of spreading two pieces of it (strategy, starting slot)  in different pieces.

I mentioned earlier that we may use two level of configs in some cases (this would be one).

magicblock-config/src/ledger.rs
```diff
@@ -69,23 +76,40 @@ impl Default for LedgerConfig {
...
+    AccountsOnly,
```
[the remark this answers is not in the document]
thlorenz: OK so that's super confusing then.
Again this could be rolled into the `Reset` variant via

```toml
strategy = { type = "reset", keep_accounts = true }
```

- `keep_accounts` would be `false` by default

If clap doesn't like this then use the intermediate config. We cannot let it limit us in expressing the strategy in a correct form.

magicblock-config/src/ledger.rs
```diff
@@ -69,23 +76,40 @@ impl Default for LedgerConfig {
...
-        self != &Self::Replay
+        matches!(self, Self::Reset | Self::DiscardResume | Self::AccountsOnly)
+    }
+
+    pub fn is_removing_accountsdb(&self) -> bool {
+        matches!(self, Self::Reset | Self::DiscardResume)
+    }
+
+    pub fn is_removing_validator_keypair(&self) -> bool {
+        matches!(self, Self::Reset | Self::DiscardResume | Self::AccountsOnly)
```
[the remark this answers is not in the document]
Dodecahedr0x: We currently remove the keypair if we remove the ledger (because why keep the key if there is no transactions in the ledger to match it against?)

magicblock-config/src/ledger.rs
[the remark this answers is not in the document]
Dodecahedr0x: Addressed in [3e359b6](https://github.com/magicblock-labs/magicblock-validator/pull/508/commits/3e359b6385fdbe44c31a430494c6ba1d1a39a67b) by keeping the differents variants in the config but converting to an enum afterward with bool indicating whether we keep accounts

magicblock-config/src/ledger.rs
```diff
@@ -69,23 +76,40 @@ impl Default for LedgerConfig {
...
+    AccountsOnly,
     ResumeOnly,
+    #[default]
     Replay,
 }
 
 impl LedgerResumeStrategy {
+    /// Whether the ledger should be resumed.
+    /// This assumes that a ledger exists.
     pub fn is_resuming(&self) -> bool {
```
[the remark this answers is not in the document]
Dodecahedr0x: removed in [3e359b6](https://github.com/magicblock-labs/magicblock-validator/pull/508/commits/3e359b6385fdbe44c31a430494c6ba1d1a39a67b) as I believe names are explicit

magicblock-config/src/ledger.rs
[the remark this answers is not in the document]
Dodecahedr0x: Adressed in [3e359b6](https://github.com/magicblock-labs/magicblock-validator/pull/508/commits/3e359b6385fdbe44c31a430494c6ba1d1a39a67b)

magicblock-config/tests/fixtures/11_everything-defined.toml
```diff
@@ -38,11 +38,10 @@ country-code = "US"
 claim-fees-interval-secs = 10
 
 [ledger]
-resume-strategy = "replay"
+resume-strategy = { variant = "replay", reset-slot = 100 }
```
thlorenz, line 41: `variant `is really kind of abstract, (validator operators don't care that it is an enum _variant_). `kind` or `type` might be better.
Then also adjust the `LEDGER_RESUME_STRATEGY_VARIANT` accordingly

test-integration/test-config/tests/auto_airdrop_feepayer.rs
```diff
@@ -34,7 +35,10 @@ fn test_auto_airdrop_feepayer_balance_after_tx() {
             ..Default::default()
         },
         ledger: LedgerConfig {
-            resume_strategy: LedgerResumeStrategy::Reset,
+            resume_strategy_config: LedgerResumeStrategyConfig {
+                variant: LedgerResumeStrategyType::Reset,
+                reset_slot: Some(0),
```
thlorenz, line 40: `Some(0)` shuold be same as `None` right since we default to `0` if not provided?

magicblock-config/src/ledger.rs
```diff
@@ -49,18 +45,106 @@ pub struct LedgerConfig {
...
 
+impl From<LedgerResumeStrategy> for LedgerResumeStrategyConfig {
+    fn from(strategy: LedgerResumeStrategy) -> Self {
+        match strategy {
+            LedgerResumeStrategy::Reset(slot, remove_accounts) => {
+                LedgerResumeStrategyConfig {
+                    variant: if remove_accounts {
+                        LedgerResumeStrategyType::Reset
+                    } else {
+                        LedgerResumeStrategyType::ResetWithAccounts
```
thlorenz, line 90: This variant is not necessary.
It should be part of the `Reset` variant with `{ keep_accounts = true }` or similar.

magicblock-config/src/ledger.rs
```diff
@@ -75,41 +159,44 @@ impl Default for LedgerConfig {
 #[serde(rename_all = "kebab-case")]
 #[strum(serialize_all = "kebab-case")]
 #[value(rename_all = "kebab-case")]
-pub enum LedgerResumeStrategy {
+pub enum LedgerResumeStrategyType {
     Reset,
-    DiscardResume,
-    AccountsOnly,
+    ResetWithAccounts,
```
thlorenz, line 164: As mentioned we want to roll this into `Reset` with a flag.

magicblock-config/src/ledger.rs
```diff
@@ -75,41 +159,44 @@ impl Default for LedgerConfig {
...
     #[default]
     Replay,
 }
 
+/// Validated strategies with the relevant arguments
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum LedgerResumeStrategy {
+    /// Reset the ledger and optionally the accountsdb.
+    /// (slot, remove_accounts)
+    Reset(u64, bool),
```
thlorenz, line 175: Here you defined this with the flat, but I'd prefer `Reset { reset_slot: u64, keep_accounts: bool }` for the following reasons.

It's hard to tell what each tuple value means (without looking at this doc) + we may add more in the future.
Picking a flag whose default is `false` (we are _not_ keeping accounts by default) is better since then if it's not supplied the `Default` just works.

thlorenz (changes requested):
Much better already!

Let's just roll `ResetWithAccounts` into `Reset` as I explained earlier and address the odd `variant` config option, as I mentioned `kind|type` makes more sense IMO.
Then it should be good to go.

magicblock-config/src/ledger.rs
Dodecahedr0x: Addressed in [ce5b1f2](https://github.com/magicblock-labs/magicblock-validator/pull/508/commits/ce5b1f2535581018cd9ae49501c0c1a8a780be90)

magicblock-config/src/ledger.rs
Dodecahedr0x: Addressed in [ce5b1f2](https://github.com/magicblock-labs/magicblock-validator/pull/508/commits/ce5b1f2535581018cd9ae49501c0c1a8a780be90)

magicblock-config/src/ledger.rs
Dodecahedr0x: Addressed in [ce5b1f2](https://github.com/magicblock-labs/magicblock-validator/pull/508/commits/ce5b1f2535581018cd9ae49501c0c1a8a780be90)

magicblock-config/src/ledger.rs
```diff
@@ -123,20 +119,24 @@ impl From<LedgerResumeStrategy> for LedgerResumeStrategyConfig {
 pub struct LedgerResumeStrategyConfig {
     #[derive_env_var]
     #[serde(default)]
-    pub variant: LedgerResumeStrategyType,
+    pub kind: LedgerResumeStrategyType,
     #[derive_env_var]
     #[clap_from_serde_skip]
     #[serde(default)]
     pub reset_slot: Option<u64>,
```
thlorenz, line 126: ditto not sure if you need this `Option` since it'd default to `0` if not provided

magicblock-config/src/ledger.rs
```diff
@@ -123,20 +119,24 @@ impl From<LedgerResumeStrategy> for LedgerResumeStrategyConfig {
...
+    pub keep_accounts: Option<bool>,
 }
 
 impl LedgerResumeStrategyConfig {
     pub fn validate_resume_strategy(&self) -> ConfigResult<()> {
         use LedgerResumeStrategyType::*;
-        match self.variant {
-            Replay | ResumeOnly if self.reset_slot.is_some() => {
+        match self.kind {
+            Replay | ResumeOnly if self.reset_slot.is_some() || self.keep_accounts.is_some() => {
```
thlorenz, line 137: Ah ok, looks like you need the above `Option`s for verification.

magicblock-config/src/ledger.rs
```diff
@@ -123,20 +119,24 @@ impl From<LedgerResumeStrategy> for LedgerResumeStrategyConfig {
...
-    pub variant: LedgerResumeStrategyType,
+    pub kind: LedgerResumeStrategyType,
     #[derive_env_var]
     #[clap_from_serde_skip]
     #[serde(default)]
     pub reset_slot: Option<u64>,
+    #[derive_env_var]
+    #[clap_from_serde_skip]
+    #[serde(default)]
+    pub keep_accounts: Option<bool>,
```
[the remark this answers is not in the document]
thlorenz: Realized in the next part why you need option for verification. Kinda sad though :)

magicblock-config/src/ledger.rs
thlorenz: Realized in the next part why you need option for verification. Kinda sad though :)

magicblock-config/src/ledger.rs
```diff
@@ -171,32 +170,48 @@ pub enum LedgerResumeStrategyType {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum LedgerResumeStrategy {
     /// Reset the ledger and optionally the accountsdb.
-    /// (slot, remove_accounts)
-    Reset(u64, bool),
+    Reset { slot: u64, keep_accounts: bool },
     /// Resume from the last slot found in the ledger.
-    /// (replay)
-    Resume(bool),
+    Resume { replay: bool },
```
thlorenz, line 175: Thanks for applying a similar idea regarding enum variants here.
Nice! this is so much better :)

magicblock-config/src/ledger.rs
```diff
@@ -171,32 +170,48 @@ pub enum LedgerResumeStrategyType {
...
 }
 
 impl LedgerResumeStrategy {
     pub fn is_resuming(&self) -> bool {
-        matches!(self, Self::Resume(_))
+        matches!(self, Self::Resume { replay: _ })
```
thlorenz, line 180:
```rs
Self::Resume { .. }
```
should also work.

magicblock-config/src/ledger.rs
```diff
@@ -171,32 +170,48 @@ pub enum LedgerResumeStrategyType {
...
     }
 
     pub fn is_removing_ledger(&self) -> bool {
-        matches!(self, Self::Reset(_, _))
+        matches!(
+            self,
+            Self::Reset {
```
thlorenz, line 186: ditto matching via `..`

magicblock-config/src/ledger.rs
```diff
@@ -171,32 +170,48 @@ pub enum LedgerResumeStrategyType {
...
+                keep_accounts: _
+            }
+        )
     }
 
     pub fn is_removing_accountsdb(&self) -> bool {
-        matches!(self, Self::Reset(_, true))
+        matches!(
+            self,
+            Self::Reset {
```
thlorenz, line 196: ditto matching via `..`

thlorenz: In this case `Reset { keep_accounts: false, .. }`

test-integration/configs/validator-offline.devnet.toml
```diff
@@ -21,7 +21,7 @@ max-snapshots = 7
 snapshot-frequency = 1024
 
 [ledger]
-resume-strategy = { variant = "reset" }
+resume-strategy = { kind = "reset" }
```
thlorenz, line 24: Not sure if I missed it somewhere, but does `resume-strategy = "reset"` also still work in this case?
Not absolutely needed, but would be nice.

test-integration/test-ledger-restore/tests/15_resume_strategies.rs
```diff
@@ -87,8 +96,11 @@ pub fn read(
 
     let validator_slot = expect!(ctx.get_slot_ephem(), validator);
     let target_slot = match strategy {
-        LedgerResumeStrategy::Reset(starting_slot, _) => starting_slot,
-        LedgerResumeStrategy::Resume(_) => slot,
+        LedgerResumeStrategy::Reset {
+            slot,
+            keep_accounts: _,
```
thlorenz, line 101: ditto `..`

test-integration/test-ledger-restore/tests/15_resume_strategies.rs
```diff
@@ -87,8 +96,11 @@ pub fn read(
...
+        } => slot,
+        LedgerResumeStrategy::Resume { replay: _ } => slot,
```
thlorenz, line 103: ditto `..`

thlorenz (approved):
Awesome! This is a lot better than before.

Config is getting complex, I'm wondering if we need a sample `toml` config somewhere which shows all the options, or in the readme.

Could you please update this PRs description to include some configuration examples?

Aside from the nits (which you can address to your liking) this LGTM now.

test-integration/test-ledger-restore/tests/15_resume_strategies.rs
Dodecahedr0x: Addressed in [25695f2](https://github.com/magicblock-labs/magicblock-validator/pull/508/commits/25695f21b9adc1666d5f68dc191e7e2c814a3c3c)

magicblock-config/src/ledger.rs
Dodecahedr0x: Addressed in [25695f2](https://github.com/magicblock-labs/magicblock-validator/pull/508/commits/25695f21b9adc1666d5f68dc191e7e2c814a3c3c)

magicblock-config/src/ledger.rs
Dodecahedr0x: Addressed in [25695f2](https://github.com/magicblock-labs/magicblock-validator/pull/508/commits/25695f21b9adc1666d5f68dc191e7e2c814a3c3c)

magicblock-api/src/ledger.rs
```diff
@@ -22,11 +22,14 @@ pub(crate) fn init(
 ) -> ApiResult<(Ledger, Slot)> {
     let resume_strategy = &ledger_config.resume_strategy();
     let starting_slot = match resume_strategy {
-        LedgerResumeStrategy::Resume(_) => {
+        LedgerResumeStrategy::Resume { replay: _ } => {
             let previous_ledger = Ledger::open(ledger_path.as_path())?;
             previous_ledger.get_max_blockhash().map(|(slot, _)| slot)?
         }
-        LedgerResumeStrategy::Reset(slot, _) => *slot,
+        LedgerResumeStrategy::Reset {
```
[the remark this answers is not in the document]
Dodecahedr0x: Addressed in [25695f2](https://github.com/magicblock-labs/magicblock-validator/pull/508/commits/25695f21b9adc1666d5f68dc191e7e2c814a3c3c)

PR #518, opened 2025-08-26
feat: add commit info to getVersion

GabrielePicco:
## Greptile Summary

This PR enhances the `getVersion` RPC endpoint by adding git commit information and extending the version response structure. The changes involve four main components:

1. **New git-version dependency**: Added to `magicblock-version/Cargo.toml` to enable compile-time git commit hash extraction using the `git_version::git_version!()` macro.

2. **Enhanced Version struct**: Modified `magicblock-version/src/lib.rs` to include a new `solana_core` field for tracking the underlying Solana version, updated commit computation to use the git-version macro instead of environment variables, and changed the default client identifier from SolanaLabs to MagicBlock.

3. **Extended RPC response structure**: Created a new `RpcVersionInfoExt` struct in `magicblock-rpc/src/traits/rpc_minimal.rs` that provides more detailed version information including separate fields for solana-core version, feature-set, git-commit (as hex string), and magicblock-core version.

4. **Updated RPC handler**: Modified the `get_version` method implementation in `magicblock-rpc/src/handlers/minimal.rs` to return the new extended version info structure, formatting the git commit as an 8-character hexadecimal string and properly populating all version fields.


lucacillario (approved):
LGTM

PR #525, opened 2025-09-03
feat: add lazy hydration

GabrielePicco:


thlorenz (approved):
LGTM.
This will all be deprecated with the new cloning pipeline integration where we basically clean out all undelegated accounts from the bank on restart.

PR #530, opened 2025-09-05
For newer version of committor program change its id

taco-paco:
We want old running validators still use an old version of the program. Instead of replacing and old version with new one and breaking commits of old nodes we just deploy a new version separately at new address.


GabrielePicco (approved):
LGTM!

thlorenz (approved):
LGTM

PR #533, opened 2025-09-09
fix: failed to fetch satisfactory slot

GabrielePicco:
## Description

Moving that target slot call outside the loop to reduce the FailedToFetchSatisfactorySlot probability compared to recomputing it on every retry, because it removes the “moving target” that can keep outrunning a lagging HTTP node.

### DIFF
--- a/magicblock-account-cloner/src/remote_account_cloner_worker.rs
+++ b/magicblock-account-cloner/src/remote_account_cloner_worker.rs
@@ -508,10 +508,10 @@ where
 
             // Fetch the account, repeat and retry until we have a satisfactory response
             let mut fetch_count = 0;
+            let min_context_slot =
+                self.account_updates.get_last_known_update_slot(&clock::ID);
             loop {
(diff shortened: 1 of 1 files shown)


GabrielePicco:
> I'm not entirely sure this is 100% correct since I'm missing the full context around the min context slot here. But it seems like we could get an outdated account data since we'd request with the same min context slot. We'd have to make sure we don't override more up to date data we might have received via a subscription in the meantime.
> 
> I'm approving in the hopes that this fixes our immediate problems + the new pipeline will replace all of this anyways.

The min_context_slot is set the first time once the subscription is confirmed. Assuming this is correct, there is no need to request the current slot again on retries (we just need to enforce that the account was fetched at a slot higher than the subscription confirmation slot → no missed updates).
In the current implementation, the `min_context_slot` is set to the current clock slot on retries, so even if we retry after 3–4 seconds, the error could still occur because we are chasing a moving target.

PR #549, opened 2025-09-16
fix: make transaction execution serialized

bmuddha:
this is achieved via a single permit semaphore, which is acquired by every transaction, but unlike previous implementation it doesn't involve any OS level synchronous locks, which were causing runtime shortage by blocking all the runtime threads, leading to latency spikes

## Greptile Summary

Updated On: 2025-09-16 11:13:01 UTC

This PR refactors transaction execution from synchronous OS-level locking to asynchronous semaphore-based serialization to address runtime thread blocking and latency spikes. The key change introduces a single-permit `tokio::Semaphore` (`TXN_SERIALIZER`) in `execute_transaction.rs` that replaces the previous `StWLock` mechanism. This maintains transaction serialization (one transaction at a time) while allowing the async runtime to yield threads instead of blocking them.

The changes cascade throughout the codebase, converting multiple components to async:
- Transaction execution functions (`execute_legacy_transaction`, `execute_sanitized_transaction`) are now async
- The `AccountDumper` trait and its implementations are converted to async with `#[async_trait]`
- Account cloning, dumping, and commit processing operations now use `.await` patterns
- RPC transaction handling removes synchronous metrics wrappers to accommodate async execution
- Several Cargo.toml files add `tokio` and `async-trait` dependencies

The architectural change maintains existing functionality while improving runtime efficiency. The semaphore approach ensures transactions are still processed sequentially but without blocking the entire thread pool, which was causing performance degradation in the previous synchronous implementation.

## Confidence score: 4/5

- This PR is generally safe to merge but requires careful testing due to the async conversion scope
- Score reflects the comprehensive async transformation across multiple critical components that could introduce subtle timing or concurrency issues
- Pay close attention to transaction execution files and account dumping operations for proper async handling

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3873,6 +3873,7 @@ dependencies = [
 name = "magicblock-account-dumper"
 version = "0.2.1"
 dependencies = [
+ "async-trait",
  "bincode",
  "magicblock-bank",
  "magicblock-mutator",
@@ -4330,6 +4331,7 @@ dependencies = [
  "solana-timings",
  "spl-token",
  "spl-token-2022 6.0.0",
+ "tokio",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


GabrielePicco (approved):
LGTM!

PR #554, opened 2025-09-23
hotfix: always treat dlp owned accounts as dirty

GabrielePicco:
<h3>Summary</h3>

This hotfix modifies the account cloning cache logic to always treat DLP (Delegation Program) owned accounts as dirty, forcing them to be re-cloned instead of using cached results.


thlorenz (approved):
Given this is a temporary workaround we can merge.

PR #556, opened 2025-09-25
fix: periodic committer attempts to commit undelegated accounts

taco-paco:
In `ExternalAccountsManager` ignore nondelegated accounts during periodic commits

### DIFF
--- a/magicblock-accounts/src/external_accounts_manager.rs
+++ b/magicblock-accounts/src/external_accounts_manager.rs
@@ -378,7 +378,7 @@ where
         // Therefore we use the current slot which could result in two commits with the same
         // slot. However since we most likely will phase out frequent commits we accept this
         // inconsistency for now.
-        static MESSAGE_ID: AtomicU64 = AtomicU64::new(u64::MAX - 1);
+        static MESSAGE_ID: AtomicU64 = AtomicU64::new((i64::MAX - 1) as u64);
 
         let slot = self.internal_account_provider.get_slot();
         let blockhash = self.internal_account_provider.get_blockhash();
@@ -394,6 +394,7 @@ where
                         None
                     })
             })
+            .filter(|(_, _, _, acc)| acc.is_delegated())
             .filter(|(_, _, prev_hash, acc)| {
                 prev_hash.map_or(true, |hash| hash_account(acc) != hash)
             })
--- a/magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
+++ b/magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
@@ -246,7 +246,7 @@ where
             .execute(intent.inner.clone(), persister)
             .await
             .inspect_err(|err| {
-                error!("Failed to execute BaseIntent: {:?}", err)
+                error!("Failed to execute BaseIntent. id: {}. {:?}", intent.id, err)
             })
             .map(|output| ExecutionOutputWrapper {
                 id: intent.id,
--- a/magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
+++ b/magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
@@ -246,7 +246,10 @@ where
             .execute(intent.inner.clone(), persister)
             .await
             .inspect_err(|err| {
-                error!("Failed to execute BaseIntent. id: {}. {:?}", intent.id, err)
+                error!(
+                    "Failed to execute BaseIntent. id: {}. {:?}",
+                    intent.id, err
+                )
             })
             .map(|output| ExecutionOutputWrapper {
                 id: intent.id,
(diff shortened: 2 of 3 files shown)


magicblock-accounts/src/external_accounts_manager.rs
```diff
@@ -378,7 +378,7 @@ where
         // Therefore we use the current slot which could result in two commits with the same
         // slot. However since we most likely will phase out frequent commits we accept this
         // inconsistency for now.
-        static MESSAGE_ID: AtomicU64 = AtomicU64::new(u64::MAX - 1);
+        static MESSAGE_ID: AtomicU64 = AtomicU64::new((i64::MAX - 1) as u64);
```
snawaz, line 381: I was guessing that we must be using `u64` as ID. So this change makes total sense. 👍

programs/magicblock/src/utils/account_actions.rs
```diff
@@ -14,8 +14,10 @@ pub(crate) fn set_account_owner(
     acc.borrow_mut().set_owner(pubkey);
 }
 
+/// Sets proper values on account during undelegation
 pub(crate) fn set_account_owner_to_delegation_program(
     acc: &RefCell<AccountSharedData>,
 ) {
+    acc.borrow_mut().set_delegated(false);
```
thlorenz, line 21: We should not do that here IMO.
Instead once the acocunt updates on chain and is cloned again we should change that flag since then the delegation record would no longer be there.
Is there a reason for this change? The whole cloning pipeline will be replaced soon, so ideally we'd leave the obsolete one as is to not run into more issues until then.

At any rate we'd have to rename that function if we do change the behavior.

magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
```diff
@@ -246,7 +246,10 @@ where
             .execute(intent.inner.clone(), persister)
             .await
             .inspect_err(|err| {
-                error!("Failed to execute BaseIntent: {:?}", err)
+                error!(
```
thlorenz, line 249: This seems to be just a formatting change.
I've seen that our fmt is out of date, i.e. if I run `make ci-fmt` locally then I get a huge amount of changes.
It seems our workflow that checks `fmt` is no longer working as expected.

(Unrelated to this PR, but this change here just confirms that).

magicblock-accounts/tests/commit_delegated.rs
```diff
@@ -86,13 +86,15 @@ async fn test_commit_two_delegated_accounts_one_needs_commit() {
     generate_validator_authority_if_needed();
     let commit_needed_pubkey = Pubkey::new_unique();
     let commit_needed_account = generate_account(&commit_needed_pubkey);
-    let commit_needed_account_shared =
+    let mut commit_needed_account_shared =
         AccountSharedData::from(commit_needed_account.clone());
+    commit_needed_account_shared.set_delegated(true);
```
thlorenz, line 91: Ditto, let's not change this behavior unless we have a reason why it needs to be done.
I want to avoid changing more than necessary in order to just keep the current cloning version working until we have the new one.

magicblock-accounts/tests/commit_delegated.rs
```diff
@@ -86,13 +86,15 @@ async fn test_commit_two_delegated_accounts_one_needs_commit() {
...
 
     let commit_not_needed_pubkey = Pubkey::new_unique();
     let commit_not_needed_account = generate_account(&commit_not_needed_pubkey);
-    let commit_not_needed_account_shared =
+    let mut commit_not_needed_account_shared =
         AccountSharedData::from(commit_not_needed_account.clone());
+    commit_not_needed_account_shared.set_delegated(true);
```
thlorenz, line 97: ditto, leave _delegated_ alone

thlorenz (changes requested):
Looks good except for the delegation flag changes.

magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
taco-paco: The log message changed - formatting changed. `make ci-fmt` doesn't complain on my side

magicblock-accounts/src/external_accounts_manager.rs
thlorenz: I'm not fully understanding why we have to adapt this here.
Why not convert it right when writing to Sqlite and reading it like we do for other u64s?

programs/magicblock/src/utils/account_actions.rs
taco-paco: Resolving as this change will be in separate pr

magicblock-accounts/src/external_accounts_manager.rs
taco-paco: That also works, actually would better align with how other values are treated

taco-paco: Done

PR #560, opened 2025-10-02
Hotfix: don't clone and cache not existing accounts

GabrielePicco:


magicblock-account-cloner/src/account_cloner.rs
```diff
@@ -71,6 +71,8 @@ pub enum AccountClonerUnclonableReason {
     DoesNotHaveDelegatedEscrowAccount,
     DoesNotAllowEscrowedPda,
     DoesNotAllowFeepayerWithEscrowedPda,
+    /// The account does not exist on-chain (RPC returned empty/system default)
```
thlorenz, line 74: Did you verify that the RPC returns this? I thought it returns some user error if the account is not found or `None` [see here](https://docs.rs/solana-client/latest/solana_client/rpc_client/struct.RpcClient.html#method.get_account_with_config).

thlorenz (approved):
LGTM even though I'm not sure how we'd ever encounter the case that this is fixing as I explained in my comment.

magicblock-account-cloner/src/remote_account_cloner_worker.rs
```diff
@@ -442,12 +443,21 @@ where
...
-        self.last_clone_output
-            .write()
-            .expect("RwLock of RemoteAccountClonerWorker.last_clone_output is poisoned")
-            .insert(*pubkey, updated_clone_output.clone());
-        if let Ok(map) = self.last_clone_output.read() {
-            metrics::set_cached_clone_outputs_count(map.len());
+        // Do not cache non-existent accounts
+        let should_cache = match &updated_clone_output {
+            AccountClonerOutput::Unclonable { reason, .. } => {
+                !matches!(reason, AccountClonerUnclonableReason::DoesNotExist)
```
snawaz, line 449: `!matches!` 

What kind of Rust sorcery is that? Seeing such code for the first time.

snawaz: ah, nevermind! I see it now. The other branch helped me to decipher it. 

It's a boolean expression and the first `!` is a negation operation! Wow!

PR #593, opened 2025-10-27
fix: false positive scheduled commits

### LINKED ISSUE #592: fix: false positive scheduled commits
`MagicContext::has_scheduled_commits` checks whether the first 8 bytes of the account are zero to determine whether there are scheduled commits.

An `intent_id` has recently been added, which makes this heuristic false. It should account for the id and look at the bytes `[8..16]`

Dodecahedr0x:
Fix #592 

## Summary by CodeRabbit

* **Bug Fix**
  * Corrected detection of scheduled commits so the system reliably recognizes scheduled items stored in the data layout.

* **Refactor**
  * Adjusted internal data handling to align checks with the underlying storage format, improving consistency.


thlorenz (approved):
LGTM even though @taco-paco needs to give final approval since he's most familiar here.

PR #606, opened 2025-10-31
fix: correct log imports

bmuddha:
## Summary by CodeRabbit

* **Chores**
  * Updated logging macro imports for improved code consistency and correctness.


magicblock-aperture/src/requests/http/send_transaction.rs
```diff
@@ -1,4 +1,4 @@
-use log::{trace, warn};
+use log::{debug, trace};
```
thlorenz: We actually just use `log::*` everywhere, but it's ok.

PR #624, opened 2025-11-11
feat: return ledger + accountsdb metrics

taco-paco:
## Summary by CodeRabbit

* **Performance**
  * Compaction now preserves account-modification data during cleanup to avoid losing important account-related entries.
  * Faster metadata access via column-count caching, reducing overhead during storage operations.

* **Monitoring**
  * Continuous metrics for ledger storage size, accounts storage size, and account counts.
  * New duration histogram for column-count measurements and renamed/updated execution-time histogram for improved observability.


bmuddha (approved):
LGTM, but please address the metric registration comment

PR #649, opened 2025-11-20
fix(aperture): prevent racy getLatestBlockhash

bmuddha:
This fixes the issue when the blockhash returned
to the client didn't exist in the cache do to timing differences between block update and cache update.

## Summary by CodeRabbit

* **Refactor**
  * Improved block update processing for more timely and reliable event handling by reordering event handling and removing duplicate paths
  * Replaced prior block state tracking with a lock-free, atomic approach to reduce contention and improve performance
  * Simplified initialization flow by consolidating how the latest block state is provided

* **Chores**
  * Introduced an atomic caching library to support the new lock-free latest-block tracking

### DIFF
--- a/magicblock-aperture/src/state/blocks.rs
+++ b/magicblock-aperture/src/state/blocks.rs
@@ -52,7 +52,7 @@ impl BlocksCache {
     ///
     /// # Panics
     /// Panics if `blocktime` is zero.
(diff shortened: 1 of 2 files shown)


snawaz (approved):
looks good to me. Great use of `arc_swap`.

PR #680, opened 2025-11-24
fix: make db thread safe

Dodecahedr0x:
`TaskSchedulerService` has `unsafe impl Sync for TaskSchedulerService {}` despite `SchedulerDatabase` being a wrapper around `rusqlite::Connection`, which is not thread-safe. This is not yet a critical issue because it is only used by a single thread but could become a problem.

This solves it by wrapping the database with a `RwLock`

## Summary by CodeRabbit

* **Refactor**
  * Database access converted to async and made concurrency-safe; scheduler service APIs converted to async and now await DB operations.
* **Bug Fixes**
  * More reliable persistence of execution updates and improved error/recovery propagation during scheduling and cancellation.
* **Tests**
  * Integration and unit tests updated to drive async DB calls via a runtime.
* **Chores**
  * Startup sequence now awaits scheduler initialization.

### DIFF
--- a/.github/workflows/ci-test-integration.yml
+++ b/.github/workflows/ci-test-integration.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   build:
     if: github.event_name != 'pull_request' || !github.event.pull_request.draft
-    runs-on: extra-large
+    runs-on: ubuntu-latest-m
     name: Build Project
     steps:
       - name: Checkout this magicblock-validator
@@ -39,7 +39,7 @@ jobs:
 
   run_integration_tests:
     needs: build
-    runs-on: extra-large
+    runs-on: ubuntu-latest-m
     strategy:
       matrix:
         batch_tests:
--- a/.github/workflows/ci-test-unit.yml
+++ b/.github/workflows/ci-test-unit.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   run_make_ci_test:
     if: github.event.pull_request.draft == false
-    runs-on: extra-large
+    runs-on: ubuntu-latest-m
     steps:
       - name: Checkout this magicblock-validator
         uses: actions/checkout@v2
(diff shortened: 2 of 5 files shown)


bmuddha (changes requested):
We should avoid locking in the code which runs inside of an async context, I understand that it's highly unlikely that this lock might result in runtime starvation, but considering that there are scores of locks employed throughout the codebase, it's nigh impossible to analyze the interplay of lock acquisition/release flow. In the future, we should replace all the Lock<Type> uses with lock free alternatives. 

For now I'd suggest to synchronize via non-blocking primitives like semaphores, in which case unsafe impl is justified always even if connection is used across multiple threads. 
```rust
struct SchedulerDatabase {
    connection: Connection,
    semaphore: Arc<Semaphore>,
}

impl SchedulerDatabase {
    fn new() -> Self {
        // ...
        let semaphore = Arc::new(Semaphore::new(1));
        let connection = Connection;
        Self {
            connection,
            semaphore,
        }
    }

    async fn connection(&self) -> ConnectionGuard<'_> {
        let permit = self
            .semaphore
            .acquire()
            .await
            .expect("semaphore cannot be closed");
        ConnectionGuard {
            permit,
            connection: &self.connection,
        }
    }

    async fn do_something(&self) {
        let con = self.connection();
        // Do some CRUD with connection
        // ...
        // con goes out of scope, releasing semaphore
        // permit and the next request can proceed
    }
}

struct ConnectionGuard<'db> {
    permit: SemaphorePermit<'db>,
    connection: &'db Connection,
}

impl Deref for ConnectionGuard<'_> {
    type Target = Connection;
    fn deref(&self) -> &Self::Target {
        &self.connection
    }
}
```

Dodecahedr0x:
Addressed in [5d10cf8](https://github.com/magicblock-labs/magicblock-validator/pull/680/commits/5d10cf8c66a3a5d9a505bf5a00e4fe9cf1f46286)

PR #681, opened 2025-11-25
feat: require 2/3 subscription confirmation

thlorenz:
## Summary

Improve subscription reliability by requiring consensus from multiple clients. Instead of
accepting a subscription as successful after the first client confirms, now require 2/3 of
available clients to confirm before considering the subscription active. This makes the system
more robust to individual client failures.

## Details

The subscription multiplexer now requires a configurable number of clients to confirm
subscriptions before treating them as successful. The required confirmation count is
automatically calculated as 2/3 of the total number of clients (with a minimum of 1).

### Testing

Added comprehensive test coverage for the new behavior, including:
- Single and multiple confirmation scenarios
- Partial client failures that still meet the confirmation requirement
- Total failure when not enough clients succeed
- Edge cases like zero clients and zero confirmations
- Verification that unsubscribe and shutdown operations ignore confirmation counts

## Summary by CodeRabbit

* **New Features**
  * Subscriptions now support configurable confirmation requirements with improved error handling for partial client failures and enhanced validation.

* **Tests**
  * Expanded test coverage for multi-confirmation scenarios, partial failures, and edge cases.

### DIFF
--- a/magicblock-chainlink/src/submux/subscription_task.rs
+++ b/magicblock-chainlink/src/submux/subscription_task.rs
@@ -12,9 +12,20 @@ use crate::remote_account_provider::{
 
 #[derive(Clone)]
 pub enum AccountSubscriptionTask {
-    Subscribe(Pubkey),
+    Subscribe(Pubkey, usize),
     Unsubscribe(Pubkey),
     Shutdown,
+}
+
+impl AccountSubscriptionTask {
+    fn op_name(&self) -> &'static str {
+        use AccountSubscriptionTask::*;
+        match self {
+            Subscribe(_, _) => "Subscribe",
+            Unsubscribe(_) => "Unsubscribe",
+            Shutdown => "Shutdown",
+        }
+    }
 }
 
 impl AccountSubscriptionTask {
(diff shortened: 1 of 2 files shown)


magicblock-chainlink/src/submux/subscription_task.rs
```diff
@@ -48,31 +89,21 @@ impl AccountSubscriptionTask {
...
 
             while let Some((i, result)) = futures.next().await {
                 match result {
                     Ok(_) => {
-                        if let Some(tx) = tx.take() {
-                            let _ = tx.send(Ok(()));
+                        successes += 1;
+                        if successes >= target_successes {
+                            if let Some(tx) = tx.take() {
+                                let _ = tx.send(Ok(()));
```
bmuddha: nit: this sends the Ok(()) multiple times potentially, on every success after target has been reached

GabrielePicco (approved):
LGTM!
We can evaluate whether this behaves as expected in production, since it depends on the configuration. For example, if we use ten different providers, the optimal strategy might be to wait for the first two responses, or to combine the 2/3 strategy with a timeout that still allows for minority confirmation.

Just a note: LGTM as a first version that we can monitor.

PR #682, opened 2025-11-25
fix: use isolated runtimes for critical components

bmuddha:
Separate async runtimes with manual thread allocation, which allows to isolate different components of the system in such a way that no blocking operations can interfere with the whole system.

Fixes shutdown sequence of RPC server by avoiding
infinite cancel branch in the shutdown loop.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graceful connection shutdown handling to prevent repeated termination attempts.
  * Enhanced ledger shutdown behavior for better state flushing.

* **Improvements**
  * Added enhanced logging capabilities throughout the application.
  * Optimized RPC server execution model.
  * Streamlined ledger truncation operations for better performance.
  * Refined runtime initialization for more efficient multi-threaded execution.

### DIFF
--- a/magicblock-aperture/src/server/http/mod.rs
+++ b/magicblock-aperture/src/server/http/mod.rs
@@ -104,6 +104,7 @@ impl HttpServer {
             let builder = conn::auto::Builder::new(TokioExecutor::new());
             let connection = builder.serve_connection(io, handler);
             tokio::pin!(connection);
+            let mut terminating = false;
 
             // This loop manages the connection's lifecycle.
             loop {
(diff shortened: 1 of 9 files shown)


magicblock-aperture/src/server/http/mod.rs
```diff
@@ -104,6 +104,7 @@ impl HttpServer {
             let builder = conn::auto::Builder::new(TokioExecutor::new());
             let connection = builder.serve_connection(io, handler);
             tokio::pin!(connection);
+            let mut terminating = false;
```
thlorenz: Nit: I wonder if there is a better way than using this flag here. Maybe putting the connection inside an `Option` and then doing something like `if let Some(..)`?

PR #689, opened 2025-11-27
Fix: abort truncation if validator exited

taco-paco:
In `LedgerTrunctationWorker::run` the branch of execution `LedgerTrunctationWorker::truncate` may take a very long time. If in the meantime validator shutsdown this branch may hold it until it finishes execution.

UPD: with latest changes the `LedgerTruncator` thread is just aborted by OS as we don't await on it. 

## Release Notes

* **Dependencies**
  * Updated console-subscriber from version 0.2.0 to 0.5.0.

* **Improvements**
  * Ledger truncation and compaction operations now support cancellation, enabling early termination of long-running tasks.
  * Enhanced system shutdown with improved error handling and diagnostics for ledger operations.


taco-paco:
> The `if cancelled()` check in the macro, only short circuits after the compaction operation finished, and considering that it might take a while, the truncator thread might still be blocking the shutdown sequence due to the introduction of thread joining. It's still better than waiting for the whole thing to finish though.

Not entirely correct. On `Ledger::shutdown` we call `self.db.backend.db.cancel_all_background_work(wait);` which will stop  all running compactions.

bmuddha (approved):
LGTM

PR #700, opened 2025-12-01
Hotfix: add missing check for ScheduleBaseIntent

taco-paco:
Added missing check for `ScheduleBaseIntent` instructions. `!signers.contains(&validator_id)` check was missing for BaseIntents

Additionally, to avoid similar inconsistencies centralized logic in 1 function

## Summary by CodeRabbit

* **Bug Fixes**
  * More consistent permission checks for committing schedules, reducing incorrect authorization rejections and improving error messages and logging.

* **Refactor**
  * Centralized schedule permission validation into a single validation path to remove duplicated checks and simplify access-control flow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3871,6 +3871,7 @@ dependencies = [
  "libc",
  "log",
  "magicblock-core",
+ "magicblock-metrics",
  "num-format",
  "num_cpus",
  "prost 0.11.9",
--- a/magicblock-ledger/Cargo.toml
+++ b/magicblock-ledger/Cargo.toml
@@ -19,6 +19,7 @@ num-format = { workspace = true }
 prost = { workspace = true }
 serde = { workspace = true }
 magicblock-core = { workspace = true }
+magicblock-metrics = { workspace = true }
 scc = { workspace = true }
 solana-account-decoder = { workspace = true }
 solana-measure = { workspace = true }
--- a/magicblock-ledger/src/database/ledger_column.rs
+++ b/magicblock-ledger/src/database/ledger_column.rs
@@ -254,6 +254,15 @@ where
         write_batch.delete::<C>(key);
     }
 
+    pub fn delete_range(
+        &self,
+        from: C::Index,
+        to: C::Index,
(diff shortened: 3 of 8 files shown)


programs/magicblock/src/magic_scheduled_base_intent.rs
```diff
@@ -520,3 +508,47 @@ impl UndelegateType {
...
+    parent_program_id: Option<&Pubkey>,
+    signers: &HashSet<Pubkey>,
+) -> Result<(), InstructionError> {
+    let validator_id = validator_authority_id();
+    if parent_program_id != Some(committee_owner)
+        && !signers.contains(committee_pubkey)
+        && !signers.contains(&validator_id)
+    {
+        match parent_program_id {
+            None => {
```
Dodecahedr0x: nit: only making the error depend on the value of the parent_program_id might not be clear. If there is no parent id, we could still expect commitee or validator to be a signer, which could result in a `MissingRequiredSignature` error. This would alse make the message in the `Some` branch more concise

programs/magicblock/src/magic_scheduled_base_intent.rs
```diff
@@ -351,28 +352,15 @@ impl CommitType {
...
-                            "ScheduleCommit ERR: account {} must be owned by {} or be a signer, but is owned by {}",
-                            pubkey, parent_id, owner
-                        );
-                        Err(InstructionError::InvalidAccountOwner)
-                    }
-                }
-            } else {
-                Ok(())
-            }
+            validate_commit_schedule_rights(
```
bmuddha: nit: *_permissions is more commonly accepted term

programs/magicblock/src/magic_scheduled_base_intent.rs
```diff
@@ -520,3 +508,47 @@ impl UndelegateType {
         }
     }
 }
+
+/// Validate that a committee account has the right to be committed.
```
bmuddha: ditto `right -> permission`

Dodecahedr0x (approved):
LGTM

PR #720, opened 2025-12-03
feat: include noop in binary

### LINKED ISSUE #563: Load noop program by default
The task scheduler service (#493) relies on a noop program to avoid duplicate signatures. Loading this program by default would make some tests cleaner.

Dodecahedr0x:
Fixes #563 

The task scheduler requires having the noop program loaded to work. Since the transaction executor does not check accounts with chainlink to eventually clone the noop program, this PR instead includes the binary during startup. This also prevents having to upload the program when deploying an ER

## Summary by CodeRabbit

* **New Features**
  * Validator now loads an additional program during initialization to ensure required runtime components are present.

* **Refactor**
  * Centralized program identifier into a shared core export for consistency.
  * Made program-loading routine callable from outside the loader.

* **Tests**
  * Simplified integration tests to obtain the latest ephemeral blockhash directly instead of sending a noop transaction.

### DIFF
--- a/test-integration/test-task-scheduler/src/lib.rs
+++ b/test-integration/test-task-scheduler/src/lib.rs
@@ -26,13 +26,9 @@ use program_flexi_counter::instruction::{
     create_delegate_ix_with_commit_frequency_ms, create_init_ix,
 };
 use solana_sdk::{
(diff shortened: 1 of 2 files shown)


Dodecahedr0x:
Closed in favor of #721

PR #724, opened 2025-12-04
fix: prevent removal of delegated/undelegating accounts on eviction

thlorenz:
## Summary

Fix incorrect account removal during eviction by preventing delegated or undelegating accounts
from being removed from the bank when unsubscribed.
Additionally fixing old vs. new delegation detection by persisting remote slot to bank.

## Details

When accounts are unsubscribed from chainlink, they get queued for removal from the bank. However,
accounts that are still in the process of undelegating or are still delegated should not be
removed as this could lead to data loss if the undelegation fails to complete on chain.

Additionally, if an account update is received before the undelegation completes, it could
overwrite the in-bank account state and incorrectly set `undelegating` to false.

Adding `remote_slot` to `AccountModification` and setting it when mutating account was second
part of fix.

### Test Changes

- Added integration test `test_committing_after_failed_undelegation` that verifies accounts
  are preserved when undelegation fails

## Summary by CodeRabbit

* **New Features**
  * Remote-slot tracking for account modifications.
  * New instruction to set a program counter value.

* **Improvements**
  * Undelegation flow includes an explicit failure/validation case.
  * Conditional account retention/removal based on delegation state.
  * More informative operational logs during account resolution and cloning.

* **Tests**
  * Added/updated tests for remote-slot and undelegation; one test marked ignored.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9182,7 +9182,7 @@ dependencies = [
 [[package]]
 name = "solana-svm"
 version = "2.2.1"
-source = "git+https://github.com/magicblock-labs/magicblock-svm.git?rev=6f74a99#6f74a994c6797c979b1afdf52f91675182a1bf8a"
+source = "git+https://github.com/magicblock-labs/magicblock-svm.git?rev=48787597#48787597fcbbdf301a068c2fec9c032419a2d3ce"
 dependencies = [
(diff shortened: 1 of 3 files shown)


GabrielePicco (approved):
LGTM!

PR #753, opened 2025-12-18
fix: flush on preparation for shutdown

taco-paco:
In order to decrease shutdown downtime, we move flush in preparation step. Rocksd's destructor will handle graceful flushing and cleanup by itself

## Summary by CodeRabbit

* **Refactor**
  * Moved final data-flush to an earlier shutdown preparation step and added error logging when that flush fails.
  * Removed the final pre-shutdown flush and now cancels background work directly to reduce shutdown latency.
  * Results: clearer shutdown error visibility and faster graceful termination.


bmuddha (approved):
LGTM, I wish all PRs were that small :D

PR #768, opened 2025-12-22
perf: skip account ensure for noop transfers

thlorenz:
## Summary

Optimize fetching accounts for transactions by skipping account ensure operations for noop system transfer
transactions (where from == to).

This prevents unnecessary account cloning and processing overhead for self-transfer
transactions, which are no-ops that waste transaction fees without meaningful state changes.

NOTE: noop indicator was changed to `lamports == 0` in d15d0dd4.

## Details

A new filter module has been added to `magicblock-chainlink` to detect noop system transfers
with O(1) performance characteristics.

### is_noop_system_transfer filter

The filter identifies transactions that consist of a single system program transfer instruction
where the sender and recipient are the same account. This is detected by:

1. Verifying the transaction has exactly one instruction
2. Confirming the instruction is a system program transfer
~3. Checking that the from and to account pubkeys are identical~
3. Checking that 0 lamports are transferred

When detected, the chainlink processor skips the account ensure operation, avoiding unnecessary
account fetches for these no-op transactions.

The implementation includes comprehensive unit tests covering:
~- Self-transfer detection (noop case)~
~- Different sender/recipient accounts (normal case)~
- Multiple instructions in a transaction
- Empty transactions
- Non-system program instructions

## Summary by CodeRabbit

* **New Features**
  * Detects and skips no-op (zero-value) system transfers to avoid unnecessary processing and reduce validation latency.

* **Tests**
  * Added unit tests for zero-value transfers, non-zero transfers, multi-instruction, empty, and non-system transactions.
  * Updated integration test flow to send two zero-value transfers and include improved confirmation handling and cleanup assertions.

### DIFF
--- a/test-integration/test-config/tests/auto_airdrop_feepayer.rs
+++ b/test-integration/test-config/tests/auto_airdrop_feepayer.rs
@@ -1,8 +1,10 @@
 use std::str::FromStr;
 
+use cleanass::{assert, assert_eq};
 use integration_test_tools::{
-    expect, loaded_accounts::LoadedAccounts,
-    validator::start_magicblock_validator_with_config_struct,
+    expect,
+    loaded_accounts::LoadedAccounts,
+    validator::{cleanup, start_magicblock_validator_with_config_struct},
     IntegrationTestContext,
 };
 use magicblock_config::{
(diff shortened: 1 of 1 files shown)


thlorenz:
> LGTM, but I would reduce the filtering/exception only for transfer of 0 lamports

Adapted in d15d0dd4

Issue #714, opened 2025-12-03
feat: allow to configure restriction for program that can be cloned in the ER


GabrielePicco:
## Description

As previously supported, we need to enable a configuration and relative implementation for restricting programs that can be cloned in the ER.

E.g.

```bash
allowed-programs = [{ id = "zb.......h" }]
```

would only allow to clone `zb.......h` program into the validator

PR #782, opened 2025-12-26
feat: add allowed_programs config option

thlorenz:
## Summary

Add `allowed_programs` configuration option to restrict which programs are cloned into the
validator from the remote chain. When specified, only programs in this list will be cloned; if
empty or not set, all programs are allowed.

CLOSES: #714

NOTE: I did not update the default config to show this new option since it is rarely used by
advanced operators only.

## Details

This change enables operators to limit the scope of program cloning for security purposes.

### magicblock-config

- Added `AllowedProgram` struct containing the program ID
- Added optional `allowed_programs` field to `ChainLinkConfig` that holds a list of allowed programs
- Exported `AllowedProgram` for external use

### magicblock-chainlink

- Modified `FetchCloner` to accept `allowed_programs` during construction
- Added `is_program_allowed()` method that checks if a program is in the allowed list
- Added filtering logic in two places:
  - `on_program_update()`: Skips cloning of disallowed programs during subscription updates
  - `fetch_and_clone_accounts()`: Skips cloning of disallowed programs when fetching from the remote
- Updated all call sites to pass the `allowed_programs` parameter
- Added comprehensive tests:
  - `test_allowed_programs_filters_programs()`: Verifies that blocked programs are filtered
  - `test_allowed_programs_none_allows_all()`: Verifies that when no restriction is set, all programs are allowed

### test-integration

- Added integration test `allowed_programs.rs` that validates the feature end-to-end with a running validator

## Summary by CodeRabbit

* **New Features**
  * Added an optional program allowlist to restrict which programs are cloned during validator synchronization.
  * Configuration now supports specifying an approved set of programs for cloning.

* **Tests**
  * Added integration tests validating program filtering during validator startup.
  * Tests cover restricted, unrestricted (None), and empty-allowlist behaviors.

### DIFF
--- a/test-integration/test-config/tests/allowed_programs.rs
+++ b/test-integration/test-config/tests/allowed_programs.rs
@@ -86,6 +86,9 @@ fn run_allowed_programs(allow_committor_program: bool) {
         validator
     );
 
+    // Give validator another moment to fully startup
+    std::thread::sleep(std::time::Duration::from_secs(1));
+
     // We request both account infos and then verify that only the allowed one
     // is cloned if one was blocked
     expect!(


bmuddha (approved):
LGTM

PR #823, opened 2026-01-12
fix: rocksdb dep breaking build again!

thlorenz:
## Summary

Re-enables default features for rocksdb, restoring zstd compression support which was previously
disabled due to linker conflicts.

**Please do not undo this fix again.**

**We cannot have the repo in a state where other devs cannot build it without modifying their system by custom linking tools**.

## Details

The previous rocksdb configuration had `default-features = false` with only `lz4` enabled to
avoid bzip2 linker conflicts. This change removes those restrictions, allowing rocksdb to use
its full feature set including zstd compression.

The Cargo.lock updates reflect the additional dependencies now pulled in:
- `zstd-sys` is now included as a dependency of `librocksdb-sys`
- `bindgen 0.72.1` added for zstd-sys bindings generation
- `libloading` added as a dependency of `clang-sys`

## Summary by CodeRabbit

* **Chores**
  * Reverted custom dependency flags so a native dependency now uses its default feature set.

* **Refactor**
  * Removed a deprecated RPC surface related to slot retrieval from the internal client mocks and implementations; this is an internal cleanup with no user-facing behavior changes.

### DIFF
--- a/test-integration/Cargo.lock
+++ b/test-integration/Cargo.lock
@@ -3489,7 +3489,7 @@ dependencies = [
  "lru",
  "magicblock-config",
  "magicblock-core",
- "magicblock-delegation-program 1.1.3 (git+https://github.com/magicblock-labs/delegation-program.git?rev=3edb41022)",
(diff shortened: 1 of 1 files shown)


Dodecahedr0x (approved):
Properly builds on my side

Issue #820, opened 2026-01-12
chore: increase historical slot window for helius GRPC


thlorenz:
In order to err on the side of too many account updates from Helius GRPC (instead of missing some) we should lower the `from_slot` to include more updates than would logically be needed taking account the _activate sub_ frequency.

That way we may receive some updates twice, but don't miss any either.
Thus recalculate the from slot as follows: `CURRENT_SLOT - (2 * FREQUENCY / SLOT_TIME)` or maybe a bit less depending on what seems safe.

PR #825, opened 2026-01-12
chore: increase Helius GRPC subscription lookback window

thlorenz:
## Summary

Increase subscription lookback window for Helius gRPC updates to improve account synchronization reliability.

CLOSES: #820

## Details

The chain laser actor's subscription activation mechanism now uses a larger lookback window when determining which accounts need to be reactivated. This change accounts for potential latency in receiving gRPC updates from Helius by:

- Using a `2x` multiplier on the base activation interval to increase the lookback distance
- Adding an extra 5-slot buffer for additional safety margin

This reduces the chance of missing account updates due to subscription lag, particularly when Helius gRPC streams experience delays in propagating state changes.

## Summary by CodeRabbit

* **New Features**
  * Optional backfill activation support to enable subscription backfill when available.

* **Improvements**
  * Activation now derives chain/from-slot dynamically for more accurate subscription start points, clearer logging, and reliable backfills.
  * Out-of-order subscription handling logs lowered to reduce noise.

* **Chores / Configuration**
  * Config templates reorganized (listen moved under an aperture section) and added reset flags for accounts DB.
  * Startup scripts updated to use explicit manifest path and refined log target.

### DIFF
--- a/test-manual/helius-laser/sh/04_start-validator.sh
+++ b/test-manual/helius-laser/sh/04_start-validator.sh
@@ -2,6 +2,6 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
-RUST_LOG=warn,magicblock=info,magicblock_chainlink=debug,magicblock_chainlink::fetch_cloner=trace,magicblock_accounts=debug \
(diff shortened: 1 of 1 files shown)


thlorenz:
Change to record previous sub slot instead.

PR #916, opened 2026-01-29
feat: simplify transaction scheduler, make it best effort

bmuddha:
## Summary
This PR removes the complexity associated with fairness enforcement, which led to subtle bugs (deadlocks) under synthetic high loads. This is achieved by making the locks best effort attempts, not dealing with any future contention. I.e. if the lock on account fails, the transaction is simply requeued and retried when the executor becomes available. This simplification makes deadlocks impossible by design and in practice doesn't lead to writer/reader starvation even under unrealistic loads.  

## Compatibility
- [x] No breaking changes, but the scheduling logic is now different

## Testing
- [x] tests have been rewritten to accomodate the new scheduling logic

## Checklist
- [x] docs updated README
- related to #589 

## Summary by CodeRabbit

* **Documentation**
  * README reorganized into Architecture/Core Components with a diagram and clearer ExecutionCoordinator and scheduling workflow.

* **Changes**
  * Scheduler redesigned with coordinator-driven lock/queue protocol, improved rescheduling and executor readiness prioritization; strict FIFO ordering no longer guaranteed but transaction correctness preserved and scheduler robustness improved.

* **Tests**
  * Tests updated to expect unordered completion and cover new lock/queue and rescheduling semantics.

### DIFF
--- a/magicblock-processor/src/scheduler/mod.rs
+++ b/magicblock-processor/src/scheduler/mod.rs
@@ -143,6 +143,9 @@ impl TransactionScheduler {
     }
 
     fn handle_new_transaction(&mut self, txn: ProcessableTransaction) {
+        // SAFETY:
+        // the caller ensured that executor was ready before invoking this
+        // method so the get_ready_executor should always return Some here
         let executor = self.coordinator.get_ready_executor().expect(
             "unreachable: is_ready() guard ensures an executor is available",
         );


magicblock-processor/src/scheduler/mod.rs
```diff
@@ -104,152 +80,94 @@ impl TransactionScheduler {
...
-    /// Updates the scheduler's state when a new slot begins.
     fn transition_to_new_slot(&self) {
         let block = self.latest_block.load();
         let mut cache = self.program_cache.write().unwrap();
-        // Remove duplicate entries from programs cache
-        // NOTE: this is an important cleanup, as otherwise it might
-        // lead cache corruption issues over time as it fills up
         cache.prune(block.slot, 0);
-        // Re-root the shared program cache to the new slot.
         cache.latest_root_slot = block.slot;
```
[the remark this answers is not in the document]
bmuddha: it's std lock, unwrap is common practice

magicblock-processor/src/scheduler/coordinator.rs
```diff
@@ -1,289 +1,149 @@
...
     }
 
-    /// Releases all locks held by a specific executor.
-    /// Called when an executor finishes a transaction or when a scheduling attempt fails.
     pub(crate) fn unlock_accounts(&mut self, executor: ExecutorId) {
-        let locks = &mut self.acquired_locks[executor as usize];
-        while let Some(lock) = locks.pop() {
-            lock.borrow_mut().unlock(executor);
+        for acc in self.held_accounts[executor as usize].drain(..) {
+            if let Some(lock) = self.locks.get_mut(&acc) {
```
taco-paco: I think else case here shall not ever happen, if it does it is a sign of bug, lets add error!() and alert to be able to catch that.

magicblock-processor/src/scheduler/mod.rs
```diff
@@ -30,50 +30,26 @@ use tracing::{error, info, instrument, warn};
...
 
-/// The central transaction scheduler responsible for distributing work to a
-/// pool of `TransactionExecutor` workers.
-///
-/// This struct acts as the single entry point for all transactions entering the processing
-/// pipeline. It receives transactions from a global queue and dispatches them to available
-/// worker threads for execution or simulation.
+/// The central transaction scheduler responsible for distributing work to executors.
 pub struct TransactionScheduler {
-    /// Manages the state of all executors, including locks and blocked transactions.
```
taco-paco: Let's keep the docs, i think they're quite useful. I would remove only for `accountsdb`, `shutdown` and maybe `latest_block`. Doc for transaction_rx are quite imo

magicblock-processor/src/scheduler/mod.rs
```diff
@@ -104,152 +80,94 @@ impl TransactionScheduler {
...
-                // The main transaction channel has closed, indicating a system shutdown.
-                else => {
-                    break
-                }
+                _ = self.shutdown.cancelled() => break,
+                else => break,
             }
         }
         drop(self.executors);
         self.ready_rx.recv().await;
```
taco-paco: Need to loop here

magicblock-processor/src/scheduler/mod.rs
```diff
@@ -104,152 +80,94 @@ impl TransactionScheduler {
...
-            let txn = self.coordinator.next_blocked_transaction(blocker);
-            let blocked = if let Some(txn) = txn {
-                self.schedule_transaction(exec, txn)
-            } else {
+            let Some(txn) = self.coordinator.next_blocked_transaction(blocker)
+            else {
                 self.coordinator.release_executor(exec);
                 break;
             };
-            // Here we check whether the transaction was blocked and re-queued:
```
taco-paco: I think we need to keep this one, since edge case isn't trivial

taco-paco (approved):
LGTM! The only thing - I would keep some of the comments as they're useful,

magicblock-processor/src/scheduler/mod.rs
bmuddha: added a brief explanation

magicblock-processor/src/scheduler/mod.rs
```diff
@@ -104,152 +99,112 @@ impl TransactionScheduler {
...
         cache.prune(block.slot, 0);
-        // Re-root the shared program cache to the new slot.
         cache.latest_root_slot = block.slot;
 
-        // Update the sysvars in the AccountsDB
+        // Update Clock sysvar
         if let Some(mut account) = self.accountsdb.get_account(&clock::ID) {
             let _ = account.serialize_data(&block.clock);
             let _ = self.accountsdb.insert_account(&clock::ID, &account);
         }
```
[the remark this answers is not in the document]
bmuddha: acceptable

PR #918, opened 2026-01-30
fix: continue startup when some pubsub clients fail

thlorenz:
## Summary

Make validator startup more resilient by continuing even when some pubsub clients fail to connect. Previously, a single failed pubsub connection would prevent the validator from starting entirely. Now the validator logs a warning and continues with the remaining pubsub clients, only failing if all clients fail to connect.

CLOSES: #917

### magicblock-chainlink

- Changed `RemoteAccountProvider` initialization to collect pubsub client connection results individually rather than using `try_join_all`, which would fail fast on first error
- Added `AllPubsubClientsFailed` error variant for the case when no pubsub clients could connect
- Added logging with endpoint label when a pubsub client fails to connect, so operators can identify which provider had issues
- Added `Endpoint::label()` helper method to extract the label from any endpoint variant

## Summary by CodeRabbit

* **New Features**
  * Endpoint labels added for clearer identification in logs and diagnostics

* **Bug Fixes**
  * More robust Pub/Sub connection handling: individual endpoint failures no longer cause total failure
  * Added explicit error when all Pub/Sub clients fail to connect
  * Improved endpoint-specific logging and diagnostics to aid troubleshooting

### DIFF
--- a/magicblock-chainlink/src/remote_account_provider/chain_laser_actor.rs
+++ b/magicblock-chainlink/src/remote_account_provider/chain_laser_actor.rs
@@ -149,11 +149,11 @@ impl ChainLaserActor {
         abort_sender: mpsc::Sender<()>,
         slots: Slots,
         rpc_client: ChainRpcClientImpl,
-    ) -> RemoteAccountProviderResult<(
+    ) -> (
         Self,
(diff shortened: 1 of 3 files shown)


bmuddha (approved):
LGTM

PR #921, opened 2026-02-01
feat: enable all syscalls

GabrielePicco:
## Summary

Enable all syscalls

## Compatibility
- [x] No breaking changes

## Testing
- [ ] tests (or explain)

## Checklist
- [ ] closes #

## Summary by CodeRabbit

* **Changes**
  * Feature activation simplified—all features now enabled by default instead of being individually configurable.
  * API method signature updated to require an additional parameter.


bmuddha (changes requested):
We cannot enable all features, some of them will just break things (like direct memory mapping for accounts), we should keep everything explicit for now

PR #973, opened 2026-02-18
refactor: replace BaseTask trait objects with BaseTaskImpl enum

taco-paco:
## Summary
Closing technical debt and simplifying tasks and intents moving forward. 
This is refactoring PR that focused on just replacing `dyn BaseTask` in order to keep it small. Future improvements will be made in next iterations. Some BaseTask methods were removed from the trait as unnecessary with concrete type.

- Replace `Box<dyn BaseTask>` with concrete `BaseTaskImpl` enum across committor service
- Consolidate `ArgsTask`/`BufferTask` into unified `CommitTask` with `CommitDelivery` enum for delivery strategy (StateInArgs, StateInBuffer, DiffInArgs, DiffInBuffer)
- Remove visitor pattern (`PersistorVisitor`, `TaskVisitorUtils`) in favor of direct enum matching
- Extract `CommitTask` into its own module with `BaseTask` trait implementation and separated instruction builders
- Rename `CommitStage` -> `CommitBufferStage`, `CommitDeliveryDetails` -> `CommitDelivery`
- Remove `reset_commit_id` from `BaseTask` trait (CommitTask-specific), change `try_optimize_tx_size` to `&mut self` -> `bool`
- try_optimize_tx_size shall dissappear in future as it is business of optimizer

## Compatibility
- [x] No breaking changes

## Testing
- [x] tests (or explain)
- Unit tests in `serialization_safety_test` rewritten to use `CommitTask` directly
- Integration tests in `test-committor-service` updated to use new types

## Checklist
- [ ] closes #

## Summary by CodeRabbit

* **Refactor**
  * Unified task model into a single concrete representation and updated public task/strategy APIs to use it, simplifying preparation, delivery and execution flows.

* **New Features**
  * Introduced explicit multi-stage commit delivery handling (preparation/cleanup) for state vs buffer deliveries and automated cleanup sequencing.

* **Tests**
  * Updated and added tests and helpers to exercise the unified task model and delivery-stage behavior.

* **Chores**
  * Minor repo cleanup: updated ignore patterns.

### DIFF
--- a/test-integration/Cargo.lock
+++ b/test-integration/Cargo.lock
@@ -2251,10 +2251,10 @@ dependencies = [
 
 [[package]]
 name = "guinea"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "bincode",
- "magicblock-magic-program-api 0.8.1",
+ "magicblock-magic-program-api 0.8.2",
  "serde",
  "solana-program",
 ]
@@ -3273,7 +3273,7 @@ dependencies = [
 
 [[package]]
 name = "magicblock-account-cloner"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "async-trait",
  "bincode",
@@ -3283,7 +3283,7 @@ dependencies = [
  "magicblock-config",
  "magicblock-core",
  "magicblock-ledger",
- "magicblock-magic-program-api 0.8.1",
+ "magicblock-magic-program-api 0.8.2",
  "magicblock-program",
  "magicblock-rpc-client",
  "rand 0.9.2",
@@ -3305,7 +3305,7 @@ dependencies = [
 
 [[package]]
 name = "magicblock-accounts"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "async-trait",
  "magicblock-account-cloner",
@@ -3327,7 +3327,7 @@ dependencies = [
 
 [[package]]
 name = "magicblock-accounts-db"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "lmdb-rkv",
  "magicblock-config",
@@ -3342,7 +3342,7 @@ dependencies = [
 
 [[package]]
 name = "magicblock-aperture"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "agave-geyser-plugin-interface",
  "arc-swap",
@@ -3390,7 +3390,7 @@ dependencies = [
 
 [[package]]
 name = "magicblock-api"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "anyhow",
  "borsh 1.6.0",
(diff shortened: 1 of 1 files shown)


magicblock-committor-service/src/tasks/commit_task.rs
```diff
@@ -0,0 +1,307 @@
...
+};
+
+/// Lifecycle stage of a buffer used for commit delivery.
+/// Tracks whether the on-chain buffer still needs to be initialized
+/// or is ready to be cleaned up after a successful commit.
+#[derive(Clone, Debug)]
+pub enum CommitBufferStage {
+    Preparation(PreparationTask),
+    Cleanup(CleanupTask),
+}
```
snawaz: design: this should be a `struct`, because `preparation` and `cleanup` are not "choices".. if you have one, you need the other one as well..means they exist in pair like "constructor/acquire/lock/open" and "destructor/release/unlock/close".

Once you have this as `struct`, the `fn prepare_task_handling_errors` will also be simplified. Currently, the handling there feels very unnatural because of `enum` type, as you cannot access both parts of the pair at the same time, but you do need there: so currently, you switch from one to other, and the switch back. That can be avoided.

magicblock-committor-service/src/tasks/commit_task.rs
```diff
@@ -0,0 +1,307 @@
...
+/// A task that commits a delegated account's state to the base layer.
+///
+/// The delivery strategy ([`CommitDelivery`]) determines how the data reaches
+/// the chain (inline args vs buffer, full state vs diff).
+#[derive(Clone, Debug)]
+pub struct CommitTask {
+    pub commit_id: u64,
+    pub allow_undelegation: bool,
+    pub committed_account: CommittedAccount,
+    pub delivery_details: CommitDelivery,
```
snawaz: design: the fields should be private, else we cannot enforce ["correct by construction"][1]. 

[1]: https://www.youtube.com/watch?v=nLSm3Haxz0I

magicblock-committor-service/src/tasks/commit_task.rs
snawaz: > You said that it is not state machine.
> What is blockchain/computer if not a state machine

Well, everything is a state machine, like even ChatGPT said:

>Everything is technically a state machine.That doesn't mean every pair of operations should be an enum.

So that is more philosophy than a useful argument in this context. It’s similar to saying **"code is data"**, or **"humans are animals"**. While technically true, it doesn’t help much when solving a concrete problem.

If a builder takes that literally and thinks **"humans are animals, horses are animals too"**, he might design a house that will likely end up not really suitable for either!

That is probably why you used `enum` for `prepare` and `cleanup`, but an array for `commit` and `undelegate`! Mathematically, `enum` and `array` are non-equivalent concepts!

Again, I'm talking about the specifics (the choices about the types), not philosophy! 

**If you use an `array` and think it's a state machine in your head, I'm perfectly OK with that, because it is consistent. But if you use `enum`, then it is a problem because it makes the design inconsistent!**

PR #1009, opened 2026-03-02
fix: rebuild validator for manual integration test tasks

thlorenz:
## Summary

Make validator binary build automatically for all integration tests that require it by adding
validator as a dependency to the test targets.

## Details

The integration tests and setup commands now depend on the validator binary being built before
running. This ensures the validator is always fresh when needed by the tests without
requiring manual builds.

## Summary by CodeRabbit

* **Tests**
  * Integrated validator into the test workflow so test targets now require a validator; test runs prefer a prebuilt test runner and automatically fall back to building/running a local runner when needed.

* **Chores**
  * Build system updated to ensure the validator is compiled and available before tests execute.


lucacillario (approved):
LGTM

GabrielePicco:
Closing, we can re-open the PR once it can be merged in the new CI workflow

PR #1024, opened 2026-03-04
feat(committor): enforce per-account commit limit via MagicSys syscall

### LINKED ISSUE #737: feat: limit gasless commits to what can be charged on undelegation from the rent
## Description

We need to limit the amount of gasless commits to what we can charge on undelegation from the rent.

## Proposed solution

Fails all commits if the nonce of the commit is > than x, where x is the number of commits for which we can cover fees on undelegation (from the recover rent)

This is necessary for re-enabling frequency commits

### Notes

We should also add an alternative pay to pay/topup. One discussed idea is to charge from a delegated PDA deterministically derived from the committing program ID

taco-paco:
## Summary
- Introduced `MagicSys` to enable syscalls outside of SVM.
- Added `fetch_current_commit_nonces` to `MagicSys` and `MagicSysAdapter`;
- Integrated `fetch_current_commit_nonces` in magic-program to limit commits without undelegation
- Refactor `CacheTaskInfoFetcher` to prevent race conditions and duplicate concurrent requests: replace flat nonce cache with per-account `NonceLock` (`Arc<TMutex<u64>>`), add `retiring` map for in-flight evicted locks, added `fetch_current_commit_nonces` (no-increment)

Also since intents are used both in magic-program and validator it is proposed to move it into `magicblock-core` as it was done with `CommittedAccount`, as it was needed for MagicSys

## Compatibility
- [ ] No breaking changes

## Testing
- [x] Unit tests: `process_schedule_commit_tests.rs` — commit limit exceeded returns `COMMIT_LIMIT_ERR`; commit+undelegate at limit succeeds
- [x] Integration tests: `03_commit_limit.rs` — end-to-end validation against a running validator

## Checklist
- [] docs updated (if needed)
- [ ] closes #737 

## Summary by CodeRabbit

* **New Features**
  * Commit-limit validation to prevent exceeding transaction thresholds
  * Nonce-based commit operations with current-nonce lookup for improved commit coordination
  * Metrics collection added for committor intent processing

* **Tests**
  * New integration and unit tests covering commit-limit scenarios and nonce workflows

* **Chores**
  * Public API and initialization updated to integrate the new Magic System support


magicblock-api/src/magic_sys_adapter.rs
```diff
@@ -0,0 +1,92 @@
...
+pub struct MagicSysAdapter {
+    ledger: Arc<Ledger>,
+    committor_service: Option<Arc<CommittorService>>,
+}
+
+impl MagicSysAdapter {
+    /// Returned when receiving the nonce result from the async channel fails.
+    const RECV_ERR: u32 = 0xE000_0000;
+    /// Returned when the async fetch of current commit nonces fails.
+    const FETCH_ERR: u32 = 0xE001_0000;
```
thlorenz: Why don't we increment by one here instead by 0x1_0000?

magicblock-api/src/magic_sys_adapter.rs
```diff
@@ -0,0 +1,92 @@
...
+        // committed to this tx until execution completes.
+        futures::executor::block_on(receiver)
+            .inspect_err(|err| {
+                error!(error = ?err, "Failed to receive nonces from CommittorService")
+            })
+            .map_err(|_| InstructionError::Custom(Self::RECV_ERR))?
+            .inspect_err(|err| {
+                error!(error = ?err, "Failed to fetch current commit nonces")
+            })
+            .map_err(|_| InstructionError::Custom(Self::FETCH_ERR))
```
thlorenz: Ditto condense

magicblock-committor-service/src/intent_executor/task_info_fetcher.rs
```diff
@@ -224,139 +223,415 @@ impl CacheTaskInfoFetcher {
...
+                        // 2. request for set A still ongoing
+                        // 3, another request with set A comes in, creating new locks in `CacheInner::active`
+                        // 4. 2 simultaneous requestors receive same value
+                        let old =
+                            inner.retiring.insert(evicted_pk, evicted_lock);
+                        if old.is_some() {
+                            // Safety
+                            // assume that is true:
+                            // That means that value was active & retiring at the same time
+                            // This is impossible as per logic above, contradiction. чтд.
```
thlorenz: Maybe let's not introduce cyrilic into our code base and use Q.E.D instead?

magicblock-committor-service/src/intent_executor/task_info_fetcher.rs
```diff
@@ -224,139 +223,415 @@ impl CacheTaskInfoFetcher {
...
+
+        // Acquire per-account locks sequentially in sorted order (see sort above).
+        let nonce_guards = locks_guard.lock().await;
+        let mut missing = vec![];
+        let mut result = HashMap::with_capacity(nonce_guards.len());
+        for (pubkey, guard) in nonce_guards {
+            if *guard == u64::MAX {
+                missing.push((pubkey, guard));
+            } else {
+                result.insert(*pubkey, *guard);
```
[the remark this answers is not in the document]
thlorenz: As we already discussed right now there is no synchronization at all with accounts fetched by chainlink and the metadata here.
Thus it could be possible that we get stale metadata (since we end up using a different min_context slot).
Ideally we should fetch the entire set of account + delegation record + delegation metadata together.
Once we add the nonce to the delegation record this will already be done and we just need to integrate this with chainlink fetching.
Please create a tracking issue for this though.

magicblock-core/src/traits.rs
```diff
@@ -1,6 +1,18 @@
...
 
-pub trait PersistsAccountModData: Sync + Send + fmt::Display + 'static {
+use solana_program::instruction::InstructionError;
+use solana_pubkey::Pubkey;
+
+use crate::intent::CommittedAccount;
+
+/// Trait that provides access to system calls implemented outside of SVM,
+/// accessible in magic-program.
+pub trait MagicSys: Sync + Send + 'static {
```
thlorenz: Are we making sure that this sysvar can only be used by our validator?
Or is it possible for users to request that sysvar and theoretically then invoke persist and load in order to attempt something malicious?

Or is it enforced that this functionality is only available inside the magic program and/or builtin programs?

magicblock-committor-service/src/intent_executor/task_info_fetcher.rs
taco-paco: I thought no one would spot that :) No fun in the comments, got it

taco-paco:
> commit limits can be bypassed by burst-scheduling before execution advances nonce (would be better to increment the nonce during scheduling)

That is valid concern, will address in followup PR.

> commit_and_undelegate can degrade to commit-only after undelegation failure, in which case the limit as far as I can see are not checked

This isn't a problem as account marked as undelegated and further commits aren't possible since commits require account to be delegated

magicblock-committor-service/src/intent_executor/task_info_fetcher.rs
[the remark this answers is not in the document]
taco-paco: We use remote_slot from AccountsDB itself. Race-condition will be minimum here, but yeah integration in some way with chainlink is needed

PR #1047, opened 2026-03-13
feat: add sliding block history to executor

bmuddha:
## Summary
Keep limited block history in transaction executor. This allows for the executor to set the environment per each transaction individually. 

## Compatibility
- [x] No breaking changes

## Checklist
- [x] closes #1046

## Summary by CodeRabbit

* **Refactor**
  * Improved per-slot block history and slot transition handling to make state alignment more reliable across transactions and reduce contention during cross-slot processing.
* **Logging**
  * Added warnings when attempting to transition to an untracked slot to aid troubleshooting and visibility.
* **Chores**
  * Minor internal compatibility improvement to block metadata handling.

PR #1076, opened 2026-03-20
epic: replication service

### LINKED ISSUE #935: Replication

bmuddha:
## Summary
An accumulator branch for replication service PRs 

## Checklist
- [x] closes #935

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive docs for a leader–follower replication service: startup flow, roles, event formats, snapshot recovery, failover rules, and threading model.
* **New Features**
  * New NATS-backed replication service (primary/standby) for streaming validator events.
  * Snapshots now stored as compressed archives and support external ingestion with fast‑forward recovery.
* **Chores**
  * Updated ignore rules to exclude AI-related config files.
* **Breaking Changes**
  * Removed configurable automatic snapshot cadence and its config/default entry.


bmuddha:
* **#1076** <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1076?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 7  dependent PRs ([#1007](https://github.com/magicblock-labs/magicblock-validator/pull/1007) <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1007?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1031](https://github.com/magicblock-labs/magicblock-validator/pull/1031) <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1031?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1047](https://github.com/magicblock-labs/magicblock-validator/pull/1047) <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1047?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 4 others) 👈 <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1076?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

GabrielePicco (commented):
Impressive work! This is a very good first version.

The main concern I have is that the replication stream is broker-authenticated but not end-to-end authenticated. `Primary::publish()` sends raw `Message` payloads, and `Standby::handle_message()` replays them immediately after deserialization, with no per-message signature or verification. I think it would be great to address this now so the replication envelope is signed or authenticated before replay.

magicblock-replicator/src/nats/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+
+// =============================================================================
+// Configuration
+// =============================================================================
+
+/// Resource names and configuration constants.
+mod cfg {
+    use std::time::Duration;
+
+    pub const STREAM: &str = "EVENTS";
```
GabrielePicco, line 33: These NATS resource names are global and unscoped: `EVENTS`, `SNAPSHOTS`, and `PRODUCER`. That is fine if each deployment gets its own broker/account. I would consider namespacing them by validator identity, cluster, or deployment name.

magicblock-config/src/config/validator.rs
```diff
@@ -24,15 +26,55 @@ pub struct ValidatorConfig {
...
-    /// Validator which participates in replication only as replicator (no takeover).
-    /// The `SerdePubkey` is the primary validator's pubkey used for authority signature verification.
-    ReplicateOnly(Url, SerdePubkey),
+    /// Validator which participates in replication: acting as either a primary or replicator
+    StandBy(ReplicationConfig),
+    /// Validator which participates in replication only as replicator (no takeover)
+    ReplicaOnly {
+        #[serde(flatten)]
+        config: ReplicationConfig,
+        #[serde(rename = "kebab-case")]
```
GabrielePicco: `ReplicaOnly.authority_override` looks broken at the config layer: `#[serde(rename = "kebab-case")]` makes the parser expect a literal field named `kebab-case`, not `authority-override`. More generally, we need to double-check the authority model: some paths use `effective_validator_authority_id()`, while off-chain builders and signers still hardcode `validator_authority()` / `validator_authority_id()`.

magicblock-api/src/magic_validator.rs
GabrielePicco, line 1: On `magicblock-api/src/magic_validator.rs:486`: `reserve_common_pubkeys()` still runs during validator startup, before replication role is established. In `StandBy`/`ReplicaOnly`, that can still initialize or extend lookup tables on-chain even though the node is not primary. This should be gated on leadership acquisition.

magicblock-processor/src/executor/mod.rs
GabrielePicco, line 1: On `magicblock-processor/src/executor/mod.rs:230`: `set_sysvars()` clones whatever `SlotHashes` are already in the executor cache and then appends the target block. That is not a historical reconstruction. Replay starts with the cache seeded from `ledger.latest_block()`, then `process_ledger()` walks older blocks while calling `latest_block.store(block)`, so an executor can replay an older slot while its SlotHashes view still contains future hashes. Please rebuild `SlotHashes` from `block_history` up to `prev_slot` instead of cloning the current cache, and add a forward-then-backward regression test.

magicblock-api/src/magic_validator.rs
GabrielePicco, line 1: On `magicblock-api/src/magic_validator.rs:861`: `register_validator_on_chain()` runs on node startup, not on leadership acquisition. In replication mode, that means a standby/replica can register itself on-chain before it owns primary. The same file also starts fee claiming unconditionally at lines 911-912 and unregisters on shutdown at lines 999-1003 without checking the current role.

magicblock-accounts-db/src/snapshot.rs
```diff
@@ -112,150 +99,259 @@ impl SnapshotManager {
...
         target_slot: u64,
     ) -> AccountsDbResult<u64> {
-        let mut registry = self.registry.lock();
+        let (chosen_archive, chosen_slot, index) =
+            self.find_and_remove_snapshot(target_slot)?;
+
+        let extracted_dir = self.extract_archive(&chosen_archive)?;
+        self.atomic_swap(&extracted_dir)?;
+        self.prune_invalidated_snapshots(index);
+        let _ = fs::remove_file(&chosen_archive);
```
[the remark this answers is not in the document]
bmuddha: it's a temporary solution which will be gone in the future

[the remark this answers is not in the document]
bmuddha: temporary solution will be removed in the future

magicblock-replicator/src/nats/lock_watcher.rs
```diff
@@ -0,0 +1,60 @@
...
+                    warn!(%e, "lock watch error");
+                    continue;
+                }
+            };
+            if matches!(operation, Operation::Delete | Operation::Purge) {
+                return;
+            }
+        }
+        warn!("lock watch stream ended unexpectedly");
+    }
```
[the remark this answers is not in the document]
bmuddha: false take over is impossible, as it is gated behind a network lock

magicblock-replicator/src/nats/lock_watcher.rs
```diff
@@ -0,0 +1,60 @@
...
+                }
+            };
+            match store.watch(cfg::LOCK_KEY).await {
+                Ok(w) => break Box::new(w),
+                Err(error) => {
+                    tracing::error!(%error, "failed to create lock watcher");
+                    continue;
+                }
+            }
+        };
```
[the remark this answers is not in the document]
bmuddha: that's the point, we should react as fast as possible

magicblock-replicator/src/service/primary.rs
```diff
@@ -0,0 +1,93 @@
...
+        let mut lock_tick = tokio::time::interval(LOCK_REFRESH_INTERVAL);
+
+        loop {
+            tokio::select! {
+                Some(msg) = self.messages.recv() => {
+                    if let Err(error) = self.publish(msg).await {
+                        // publish should not easily fail, if that happens, it means
+                        // the message broker has become unrecoverably unreacheable
+                        warn!(%error, "failed to publish the message");
+                        return self.ctx.into_standby(self.messages, true).await;
```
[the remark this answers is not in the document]
bmuddha: will be reviewed later

magicblock-api/src/magic_validator.rs
bmuddha: Removed ALT preparation altogether, so this should not be a problem

bmuddha: this method is idempotent, so if other primary beats us to it, it will do no harm

magicblock-processor/src/executor/mod.rs
bmuddha: yes, that's a valid point, but it's not easy to fix with current convolution of state happening in the codebase. Reading from the ledger is not a solution, since it would introduce a lag which can be longer than block time itself. I'd keep it this way for now, as it's not a regression from previous behavior which also just appended only to slot hash. I'll keep it TODO list during the overhaul, so the sysvars always perfectly synced with the ledger.

magicblock-replicator/src/nats/mod.rs
bmuddha: The idea is to have a single cheap NATs server for a each network (sharing the same keypair). Using one global NATs server will hurt geo distribution and scales poorly. Complicating the namespacing would be unjustified in this case.

magicblock-config/src/config/validator.rs
bmuddha: fixed

PR #1078, opened 2026-03-20
refactor(committor): add IntentClient, move common executor logic in utils

taco-paco:
## Summary
Refactor of the intent executor internals: extracts shared logic, centralizes timeout handling, and cleans up the two-stage state machine. Scope is purely structural — no behavioral changes.

### New: `IntentExecutionClient`
RPC execution and retry logic extracted from `mod.rs` into its own file (`intent_execution_client.rs`). `mod.rs` no longer owns raw RPC concerns.

### New: `utils.rs`
Shared executor utilities extracted from `mod.rs` and `single_stage_executor.rs`:
- `prepare_and_execute_strategy` — prepare + execute a single strategy
- `handle_commit_id_error` / `handle_cpi_limit_error` / `handle_undelegation_error` — error recovery helpers
- `execute_with_timeout` — centralizes the timeout + callback-firing pattern that was previously duplicated inline for every stage
- `StageExecutor` trait + `SingleStage` / `CommitStage` / `FinalizeStage` wrappers — uniform interface over the three execution phases

### `mod.rs`
`single_stage_execution_flow` and `two_stage_execution_flow` now each call `execute_with_timeout` once (or twice for two-stage) instead of duplicating the `if has_callbacks { if let Some(time_left) { match timeout {...} } else {...} } else {...}` block per stage. `IntentExecutionReport` is now local per `execute()` call rather than held as mutable state on the executor between calls.

### `TransactionStrategy::remove_actions`
New method on `TransactionStrategy` that strips action tasks and returns them as a separate strategy for cleanup, used by both single-stage and two-stage error handlers.

### `TwoStageExecutor`
Callback handling on commit failure now fires both commit and finalize callbacks (since commit failure means finalize won't run). `patch_commit_strategy` and `patch_finalize_strategy` extracted as public methods.

## Compatibility
- [x] No breaking changes

## Testing
- [x] tests (or explain)

## Checklist
- [ ] closes #

## Summary by CodeRabbit

* **New Features**
  * New dedicated execution client for sending intents with retry/backoff and async metrics reporting.

* **Improvements**
  * Refactored execution flow into clear single- and two-stage stages with timeout-aware callback handling.
  * Better recovery: targeted handling for commit-id, CPI-limit, and undelegation failures; action-only extraction.

* **Tests**
  * Integration tests updated to validate new orchestration and reporting/report-collection flows.

### DIFF
--- a/test-integration/Cargo.lock
+++ b/test-integration/Cargo.lock
@@ -2340,10 +2340,10 @@ dependencies = [
 
 [[package]]
 name = "guinea"
-version = "0.8.5"
+version = "0.8.6"
 dependencies = [
  "bincode",
- "magicblock-magic-program-api 0.8.5",
+ "magicblock-magic-program-api 0.8.6",
  "serde",
  "solana-program",
 ]
@@ -3426,7 +3426,7 @@ dependencies = [
 
 [[package]]
 name = "magicblock-account-cloner"
-version = "0.8.5"
+version = "0.8.6"
 dependencies = [
  "async-trait",
  "bincode",
@@ -3437,7 +3437,7 @@ dependencies = [
  "magicblock-config",
  "magicblock-core",
  "magicblock-ledger",
- "magicblock-magic-program-api 0.8.5",
+ "magicblock-magic-program-api 0.8.6",
  "magicblock-program",
  "magicblock-rpc-client",
  "rand 0.9.2",
@@ -3459,7 +3459,7 @@ dependencies = [
 
 [[package]]
 name = "magicblock-accounts"
-version = "0.8.5"
+version = "0.8.6"
 dependencies = [
  "async-trait",
  "magicblock-account-cloner",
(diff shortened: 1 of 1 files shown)


test-integration/test-committor-service/tests/test_intent_executor.rs
```diff
@@ -1320,17 +1360,25 @@ async fn test_callbacks_fired_in_two_stage() {
...
-        .commit(&committed_pubkeys, &None::<IntentPersisterImpl>)
+        .commit(
+            &committed_pubkeys,
+            &transaction_preparator,
+            &task_info_fetcher,
+            &None::<IntentPersisterImpl>,
+        )
         .await
         .expect("commit must succeed");
     executor.execute_callbacks(Some(commit_sig), Ok::<(), ActionError>(()));
```
thlorenz: `executor.execute_callbacks(...)` is called here manually *after* `commit()`, but `commit()` already calls `self.execute_callbacks(...)` internally (L157-160 in `two_stage_executor.rs`).
This means commit-stage callbacks are fired twice in this test. If the test is intentionally verifying the manual path, `commit()` should not fire callbacks internally, or this manual call should be removed.

PR #1121, opened 2026-03-31
Propagate cleanup boundary error in get_block

nikolaysamoil0ff:
## Summary
get_block() currently ignores the result of check_lowest_cleanup_slot(slot):
```rust
let _lock = self.check_lowest_cleanup_slot(slot);
```

If the requested slot has already been cleaned up, the error is silently dropped and the method continues reading from storage. This can lead to returning partial or inconsistent block data past the cleanup boundary.

The cleanup check is intended to enforce a strict boundary, so failures should be propagated instead of ignored.

This change uses `?` to return the error early, ensuring predictable behavior and preventing invalid reads beyond the cleanup range.

## Compatibility
- [ ] No breaking changes

## Testing
- [ ] tests (or explain)

## Checklist
- [ ] closes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when accessing cleaned-up ledger blocks so failures are reported and block reads are aborted instead of continuing.
  * Prevented panics during transaction metadata decoding by returning a clear conversion error when metadata is invalid.


nikolaysamoil0ff:
@GabrielePicco Could you please update the branch pls?

GabrielePicco:
> @GabrielePicco Could you please update the branch pls?

Done. Thanks for the contribution

PR #1133, opened 2026-04-06
feat: added a way to verify validator identity within ER

taco-paco:
## Summary
With callbacks it is importmant for developers to verify if callback was triggered by validator and not someone else. For that purpose they can now call VerifyValidatorIdentity in magicblock-program with "validator" account, magic-program will compare pubkeys and error if they do not match.

## Compatibility
- [x] No breaking changes

## Testing
- [x] tests (or explain)

## Checklist
- [ ] closes #

## Summary by CodeRabbit

* **New Features**
  * Added a new validator identity verification instruction to the MagicBlock program, enabling validation of validator authority accounts against expected identities.

### DIFF
--- a/programs/magicblock/src/verify_validator_identity.rs
+++ b/programs/magicblock/src/verify_validator_identity.rs
@@ -1,0 +1,187 @@
+use std::collections::HashSet;
+
+use solana_instruction::error::InstructionError;
+use solana_log_collector::ic_msg;
+use solana_program_runtime::invoke_context::InvokeContext;
+use solana_pubkey::Pubkey;
+
+use crate::{
+    schedule_transactions, utils::accounts::get_instruction_pubkey_with_idx,
+    validator::validator_authority_id,
+};
+
+pub(crate) fn process_verify_validator_identity(
+    _signers: HashSet<Pubkey>,
+    invoke_context: &mut InvokeContext,
+) -> Result<(), InstructionError> {
+    const MAGIC_CONTEXT_IDX: u16 = 0;
+    const VALIDATOR_ACCOUNT_IDX: u16 = MAGIC_CONTEXT_IDX + 1;
+
+    schedule_transactions::check_magic_context_id(
+        invoke_context,
+        MAGIC_CONTEXT_IDX,
+    )?;
+
+    let transaction_context = &invoke_context.transaction_context;
+    let ix_ctx = transaction_context.get_current_instruction_context()?;
+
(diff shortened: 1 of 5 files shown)


GabrielePicco (changes requested):
The current implementation exposes the active validator identity directly to user programs and requires each callback handler to remember to authenticate it explicitly. “opt-in security” here is unavoidable, but could rework to make it much harder to mis-use/forget.

Additionally, the current implementation couples app behavior to validator identity / authority overrides, and pushes a security-sensitive check into every integrator's code.

 I suggest reworking this so that the the service submit a DispatchCallback instruction to the magic program. The magic program validates internal state, then CPI’s into the destination program with invoke_signed using a deterministic callback-authority PDA. The app checks that PDA signer, not the active validator identity.
This keeps the trust boundary inside the magic program, and gives materially better developer experience with a much lower misuse risk as the program just need to validate that the PDA is a signer (with Anchor, just marking it as a signer).

To evaluate, but the PDA could also be scoped (derived) from the callback domain as well: intent_id, action_index, source_program.

programs/magicblock/src/verify_validator_identity.rs
```diff
@@ -0,0 +1,187 @@
...
+use crate::{
+    schedule_transactions, utils::accounts::get_instruction_pubkey_with_idx,
+    validator::validator_authority_id,
+};
+
+pub(crate) fn process_verify_validator_identity(
+    _signers: HashSet<Pubkey>,
+    invoke_context: &mut InvokeContext,
+) -> Result<(), InstructionError> {
+    const MAGIC_CONTEXT_IDX: u16 = 0;
```
Dodecahedr0x, line 17: Makes no sense to include magic context

programs/magicblock/src/verify_validator_identity.rs
```diff
@@ -0,0 +1,187 @@
...
+            );
+            InstructionError::UnsupportedProgramId
+        })?;
+
+    let validator_identity = get_instruction_pubkey_with_idx(
+        transaction_context,
+        VALIDATOR_ACCOUNT_IDX,
+    )?;
+    let expected_validator_identity = validator_authority_id();
+    if validator_identity == &expected_validator_identity {
```
Dodecahedr0x, line 44: Ix could also be idempotent but use return_data (depending on the use case)

Dodecahedr0x (changes requested):
Magic context seems like a mistake here

PR #1145, opened 2026-04-13
feat: execute cranks with PDA

Dodecahedr0x:
## Summary
- Introduces a new Magic Program instruction that executes crank instructions through CPI, replacing the validator signer with an immutable PDA signer.

## Compatibility
Breaking change: cranks can no longer rely on the validator pubkey to gate instructions to crank, but can instead use the CRANK_SIGNER

## Summary by CodeRabbit

* **New Features**
  * Added an Execute-Crank capability to run embedded instructions and a helper to build/send execute-task transactions.
  * Introduced a dedicated crank signer PDA for crank execution.

* **Refactor**
  * Simplified scheduling APIs by removing explicit extra account-list parameters.
  * Transaction construction now wraps inner instructions into a single execute-task instruction.

* **Bug Fixes**
  * Stricter validation to prevent improper signers, writable crank signer accounts, and invalid account layouts.

* **Tests**
  * Added unit and integration tests covering success and multiple failure scenarios.

### DIFF
--- a/magicblock-magic-program-api/src/lib.rs
+++ b/magicblock-magic-program-api/src/lib.rs
@@ -1,5 +1,6 @@
 pub mod args;
 pub mod instruction;
+pub mod pda;
 pub mod response;
 
 pub use solana_program::{declare_id, pubkey, pubkey::Pubkey};
@@ -23,9 +24,3 @@ pub const MAGIC_CONTEXT_SIZE: usize = 1024 * 1024 * 5; // 5 MB
 /// Rent rate for ephemeral accounts: 32 lamports per byte.
 /// This is ~109x cheaper than Solana's base rent (3,480 lamports/byte).
 pub const EPHEMERAL_RENT_PER_BYTE: u64 = 32;
-
-pub const CRANK_SEED: &[u8] = b"crank-executor";
-const CRANK_SIGNER_PDA: ([u8; 32], u8) =
-    const_crypto::ed25519::derive_program_address(&[CRANK_SEED], ID.as_array());
-pub const CRANK_SIGNER: Pubkey = Pubkey::new_from_array(CRANK_SIGNER_PDA.0);
-pub const CRANK_SIGNER_BUMP: u8 = CRANK_SIGNER_PDA.1;
--- a/magicblock-magic-program-api/src/lib.rs
+++ b/magicblock-magic-program-api/src/lib.rs
@@ -23,3 +23,9 @@ pub const MAGIC_CONTEXT_SIZE: usize = 1024 * 1024 * 5; // 5 MB
 /// Rent rate for ephemeral accounts: 32 lamports per byte.
 /// This is ~109x cheaper than Solana's base rent (3,480 lamports/byte).
 pub const EPHEMERAL_RENT_PER_BYTE: u64 = 32;
+
+pub const CRANK_SEED: &[u8] = b"crank-executor";
+const CRANK_SIGNER_PDA: ([u8; 32], u8) =
+    const_crypto::ed25519::derive_program_address(&[CRANK_SEED], ID.as_array());
+pub const CRANK_SIGNER: Pubkey = Pubkey::new_from_array(CRANK_SIGNER_PDA.0);
+pub const CRANK_SIGNER_BUMP: u8 = CRANK_SIGNER_PDA.1;
--- a/programs/magicblock/src/schedule_task/process_execute_task.rs
+++ b/programs/magicblock/src/schedule_task/process_execute_task.rs
@@ -57,7 +57,7 @@ pub(crate) fn process_execute_crank(
         );
         return Err(InstructionError::IncorrectAuthority);
     }
-    if !signers.contains(&validator_pubkey) {
+    if !signers.contains(validator_pubkey) {
         ic_msg!(
             invoke_context,
             "ExecuteCrank ERR: validator pubkey {} is not in signers",
--- a/programs/magicblock/src/schedule_task/process_execute_task.rs
+++ b/programs/magicblock/src/schedule_task/process_execute_task.rs
@@ -1,0 +1,187 @@
+use std::collections::HashSet;
+
+use magicblock_magic_program_api::CRANK_SIGNER;
+use solana_instruction::{error::InstructionError, Instruction};
+use solana_log_collector::ic_msg;
+use solana_program_runtime::invoke_context::InvokeContext;
+use solana_pubkey::Pubkey;
+
+use crate::utils::accounts::get_instruction_pubkey_with_idx;
+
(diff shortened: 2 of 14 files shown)


programs/magicblock/src/utils/instruction_utils.rs
```diff
@@ -273,6 +267,41 @@ impl InstructionUtils {
...
+    // -----------------
+    pub fn execute_task_instruction(
+        instructions: Vec<Instruction>,
+    ) -> Instruction {
+        let mut account_metas = vec![
+            AccountMeta::new_readonly(validator_authority_id(), true),
+            AccountMeta::new_readonly(CRANK_SIGNER, false),
+        ];
+        for instruction in &instructions {
+            account_metas.push(AccountMeta::new(instruction.program_id, false));
```
GabrielePicco, line 282: This copies inner `account.is_signer` into the outer `ExecuteCrank` metas. Wouldn't all signer PDAs fail with a missing top-level signature?

GabrielePicco: The existing execute-crank unit test uses the mock processor, so this may not be a tested case

GabrielePicco (commented):
LGTM in general. 

Some considerations:

-  If there is no legitimate use case (and I don't think there is), rejecting validator_authority_id() anywhere in scheduled inner instructions would be safer
- We should at least reject it as `is_writable` or `is_signer`, but unless there is a legit use-case, we should reject it all-together.

magicblock-magic-program-api/src/lib.rs
```diff
@@ -23,3 +23,9 @@ pub const MAGIC_CONTEXT_SIZE: usize = 1024 * 1024 * 5; // 5 MB
 /// Rent rate for ephemeral accounts: 32 lamports per byte.
 /// This is ~109x cheaper than Solana's base rent (3,480 lamports/byte).
 pub const EPHEMERAL_RENT_PER_BYTE: u64 = 32;
+
+pub const CRANK_SEED: &[u8] = b"crank-executor";
+const CRANK_SIGNER_PDA: ([u8; 32], u8) =
```
taco-paco: I'd suggest having `pda.rs` as we do in `dlp` and move this there

GabrielePicco (changes requested):

programs/magicblock/src/schedule_task/process_execute_task.rs
```diff
@@ -0,0 +1,297 @@
...
+use solana_log_collector::ic_msg;
+use solana_program_runtime::invoke_context::InvokeContext;
+use solana_pubkey::Pubkey;
+
+use crate::{
+    utils::accounts::get_instruction_pubkey_with_idx,
+    validator::validator_authority_id,
+};
+
+pub(crate) fn process_execute_crank(
```
taco-paco, line 15: This an interesting case. Logic here relies on the fact that validator will execute it only with what was scheduled via crank and under no other condition.
There's no  symmetrical instruction validation as in "schedule" entrypoint so validator keypair could create a valid instruction assymetrical to the rules of "schedule" having accounts as signers and crank pda as writable(as no check here). 
Now this would be a legitimate transaction. 

As result we make an assumption about validator operator - he doesn't call entrypoints he shouldn't. While he could, as tx execution is completely correct.
If the check would be here and tx would pass it would mean we could fraud proof easily, but with this it would be much more tricky and would require context of all cranks and time elapsed and so on

taco-paco (approved):
Looks good overall with 1 nit and duscission.

programs/magicblock/src/utils/instruction_utils.rs
Dodecahedr0x: I don't understand the case you're mentioning: there cannot be signer PDAs for those instructions, but they can call programs that will CPI into another program, in which case the signer is set within the program.

Instructions should already be checked to avoid having any signer except the crank signer, so there should be at most one signer in there.

Dodecahedr0x: Rejecting any use of the validator here [37c3134](https://github.com/magicblock-labs/magicblock-validator/pull/1145/commits/37c3134a69eef09cbb150899b1185bd10d093b6f)

magicblock-magic-program-api/src/lib.rs
Dodecahedr0x: Addressed [8346ef4](https://github.com/magicblock-labs/magicblock-validator/pull/1145/commits/8346ef45725bd0e4613f1fce0da37a794314b344)

programs/magicblock/src/utils/instruction_utils.rs
GabrielePicco: wouldn't a signer PDA in an inner instruction (internal CPI) be part of the account metas pushed to the outer ix?

```rust
account_metas.extend(instruction.accounts.iter().map(|account| {
    AccountMeta {
        pubkey: account.pubkey,
        is_signer: account.is_signer,
        is_writable: account.is_writable,
    }
}));
```

The case I'm considering is, for example, an `account` which is a pda of ix 2,  with `account.is_signer = true`, but the account cannot be a signer in the outer account_metas

programs/magicblock/src/schedule_task/process_execute_task.rs
Dodecahedr0x: True. We can limit the potential harm by also checking that instructions are permissionless in the execute function (no signer except crank signer).
Addressed [17db03a](https://github.com/magicblock-labs/magicblock-validator/pull/1145/commits/17db03a0ed866bcb00a17142e843feb893d821c9)

Dodecahedr0x: I also believe the validator can do worse than this (like cloning wrong accounts?)

programs/magicblock/src/utils/instruction_utils.rs
Dodecahedr0x: You're right, missed this. Fixed here [bad368f](https://github.com/magicblock-labs/magicblock-validator/pull/1145/commits/bad368f526ba5b7a3cfd015b729e1da4c587cc0d)

GabrielePicco: Thanks, yes I suppose setting it always to `false` in the outer instruction works as there are no other possible "normal" signers on a crank instruction

GabrielePicco (approved):
LGTM!

programs/magicblock/src/schedule_task/process_execute_task.rs
GabrielePicco: It’s a broader problem. We should always operate under the assumption that the validator can change the code. As long as everything is deterministic and replication can enforce or detect it, it will be fraud-provable.

PR #1155, opened 2026-04-16
feat: use getAccountInfo when ensure_accounts fetches only 1 account

GabrielePicco:
## Summary

- use the cheaper getAccountInfo when ensure_accounts fetches only 1 account

## Compatibility
- [ ] No breaking changes

## Testing
- [ ] tests (or explain)

## Checklist
- [ ] closes #


thlorenz (approved):
LGTM even though it makes the code slightly more complex.

If we don't see any savings we should revert it.

PR #1158, opened 2026-04-16
chore: adjust logs level

GabrielePicco:
## Summary

- adjust logs level

## Compatibility
- [ ] No breaking changes

## Testing
- [ ] tests (or explain)

## Checklist
- [ ] closes #

## Summary by CodeRabbit

* **Chores**
  * Adjusted logging severity for transaction failures to reduce unnecessary error-level verbosity during normal system operations, improving overall log clarity and filtering.
  * Refined logging levels for account cloning operations to more appropriately categorize different types of failures based on their operational significance and recovery requirements.

### DIFF
--- a/magicblock-account-cloner/src/lib.rs
+++ b/magicblock-account-cloner/src/lib.rs
@@ -608,7 +608,7 @@ impl ChainlinkCloner {
 
         if let Err(err) = self.send_sanitized_tx(sanitized_tx).await {
             self.lock_sent_action_txs().pop(&action_tx_sig);
-            error!(
+            debug!(
                 tx_sig = %action_tx_sig,
                 error = ?err,
                 "Failed to execute post-delegation actions transaction"
--- a/magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
+++ b/magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
@@ -556,7 +556,7 @@ where
                 })
                 .await
             {
-                warn!(
+                error!(
                     pubkey = %pubkey,
                     error = %err,
                     "Failed to clone account into bank"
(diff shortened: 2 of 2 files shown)


magicblock-account-cloner/src/lib.rs
```diff
@@ -608,7 +608,7 @@ impl ChainlinkCloner {
 
         if let Err(err) = self.send_sanitized_tx(sanitized_tx).await {
             self.lock_sent_action_txs().pop(&action_tx_sig);
-            error!(
+            debug!(
```
thlorenz, line 611: This is just about the post-delegation not the cloning itself afaik.
I'm not sure if it is acceptable for a clone to succeed, but those actions to fail.
Maybe we should at least keep this on `warn`?

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -556,7 +556,7 @@ where
                 })
                 .await
             {
-                error!(
+                warn!(
```
thlorenz: The errors we saw were due to duplicate clones blocked by the runtime right?
I'm not sure if they're still there since the dedupe logic was just merged recently and most likely was not deployed when those occurred.

At any rate if cloning fails in general then that is something we should be alerted to I think.
Maybe we can decide the loglevel more fine grained, depending on the type of error, i.e. only log on warn when we see exactly this error:

```
TransactionError(InstructionError(0, Custom(11)))
```

We could match on the error type to do that.

magicblock-account-cloner/src/lib.rs
GabrielePicco: This is actually benign as the tx would anyways fail. I think `debug` is enough.

> I'm not sure if it is acceptable for a clone to succeed, but those actions to fail.

Agreed with this concern, but this is the current behaviour. Need to be properly documented and we could rethink it. Just downgrading the log severity in this PR. Will follow up on this.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
GabrielePicco: Reverted to error, we can keep monitoring this

PR #1169, opened 2026-04-27
feat: use backfill for triton

thlorenz:
## Summary

Triton gRPC now supports backfill via the `from_slot` as helius does.
We now use that feature for triton as well.

Additionally this PR fixes a race condition where we did not have a chain_slot by the time we
make the first subscription and thus used `0` which triggered an error with the Triton
provider.
Instead we now fall back to _no backfill_ behavior for subscriptions made _before_ we have the
first `chain_slot`.

### Return Option from compute_from_slot

Changed `ChainSlot::compute_from_slot()` to return `Option<u64>` instead of `u64`. It now
returns `None` while chain slot is still `0` (uninitialized), indicating backfill is
temporarily unavailable. This prevents sending `from_slot=0` which would bypass the backfill
optimization.

### Targeted observability for degraded path

Enhanced `ChainLaserActor::compute_from_slot()` logging to distinguish three scenarios:
- Backfill not supported by endpoint (trace level)
- Backfill skipped due to chain slot still being 0 (debug level)
- Normal backfill operation with computed slot (trace level)

This helps operators debug bootstrap window behavior and identify when degraded backfill is in effect.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of slot computation for uninitialized chain states
  * Fixed optional value processing in stream management operations

* **New Features**
  * Extended endpoint support to enable backfill capabilities for additional gRPC endpoint types

* **Tests**
  * Added comprehensive test coverage for slot computation scenarios

### DIFF
--- a/magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
+++ b/magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
@@ -495,12 +495,12 @@ impl<H: StreamHandle, S: StreamFactory<H>> ChainLaserActor<H, S> {
 
     /// Computes a `from_slot` for backfilling based on the
     /// current chain slot. Returns `None` if backfilling is not
-    /// supported or the slot is still 0.
+    /// supported or the slot is still `0` (i.e. uninitialized).
     fn compute_from_slot(&self) -> Option<u64> {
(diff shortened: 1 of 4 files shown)


GabrielePicco (approved):
LGTM!

PR #1172, opened 2026-04-30
feat: drop guards and cancellation wrappers to "fix" stuck pubkeys

thlorenz:
## Summary

Fix stuck pubkeys in Chainlink account fetching by making pending fetch/clone work cancellation-safe and ownership-aware, and by tracking subscription ownership by reason instead of dropping shared watches opportunistically.

### magicblock-chainlink

- Reworks fetch-cloner pending request deduplication into owned per-pubkey operations with waiter registration, owner completion, cancellation, and timeout handling.
- Adds explicit pending-request error reporting so callers can distinguish owner failures, disappeared owners, cancellations, and timeouts instead of hanging indefinitely.
- Introduces subscription reason ownership for the remote account provider so direct account, delegation record, program data, undelegation tracking, and ATA projection watches do not release each other prematurely.
- Updates fetch-cloner subscription cleanup to release only the reasons acquired by the current operation, including delegation/action dependencies and program data accounts.

## Summary by CodeRabbit

* **New Features**
  * Per-account subscription acquire/release and undelegation-aware subscription tracking
  * Controls to cancel individual/all pending fetch requests and inspect waiter counts
  * Test helpers to block/unblock RPC, pubsub, and clone completion

* **Bug Fixes**
  * Added pending-request error variants for timeout, cancellation, owner disappearance/failure

* **Tests**
  * Expanded concurrency, subscription-lifecycle, and eviction integration tests

* **Chores**
  * Large refactor of fetch/clone and subscription orchestration; .gitignore ignores prompts/


magicblock-chainlink/src/chainlink/fetch_cloner/pending_request_guard.rs
```diff
@@ -0,0 +1,75 @@
...
+pub(super) struct PendingRequestState {
+    pub created_at: Instant,
+    pub waiters: Vec<oneshot::Sender<PendingRequestCompletion>>,
+}
+
+#[derive(Debug, Clone)]
+pub(super) enum PendingRequestCompletion {
+    Success,
+    Failed(String),
+}
```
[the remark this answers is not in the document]
thlorenz: **Verification result:** Already addressed in current code.

- pending_request_guard.rs:18 — `Success(FetchAndCloneResult)` already carries payload (import on line 8).
- fetch_cloner/mod.rs:1962-1973 — owner constructs `PendingRequestCompletion::Success(r.clone())` and forwards to all waiters via `finish_pending_request`, dismissing the guard afterward.
- fetch_cloner/mod.rs:1978-2010 — waiter loop matches `Ok(Ok(PendingRequestCompletion::Success(owner_result)))`, collects `(pubkey, owner_result)`, then merges matching `not_found_on_chain` / `missing_delegation_record` entries into `final_result`.
- Drop path (pending_request_guard.rs:75-79) still sends `Failed("owner future dropped before cleanup")`, which is the correct behavior.
- Tests in fetch_cloner/tests.rs already construct `PendingRequestCompletion::Success(FetchAndCloneResult { ... })` with payload (line 4196).

GabrielePicco (dismissed):

GabrielePicco (dismissed):

GabrielePicco (dismissed):

magicblock-chainlink/src/remote_account_provider/mod.rs
```diff
@@ -710,104 +844,163 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
...
-        );
-
-        // Wait for all accounts to resolve (either from fetch or subscription override)
+        // Setup subscriptions and trigger the fetch only for pubkeys this
+        // call actually claimed. Waiter-only pubkeys already have a
+        // subscription and an in-flight fetch owned by the original
+        // claimer; doing it again here would duplicate work and (for
+        // setup_subscriptions) double-count subscription side effects.
+        if !claimed_pubkeys.is_empty() {
+            self.setup_subscriptions(&claimed_pubkeys).await?;
```
GabrielePicco: This is still an unbounded await after the `fetching_accounts` entries have been inserted. If one `subscribe()` never resolves, this owner future never reaches `fetch()`, never dismisses the guard, and never notifies waiters. The new stale eviction only helps if another request arrives later and crosses `FETCHING_ACCOUNT_STALE_AFTER`; it is not an active cleanup mechanism. This should be wrapped in a hard operation deadline, or better, moved into an owned per-pubkey task that always completes/removes the map entry with Ok/Err/TimedOut.

magicblock-chainlink/src/remote_account_provider/mod.rs
```diff
@@ -857,34 +1050,32 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
...
         // all oneshot receivers get a response (either success or
         // error).
-        let subscription_results = join_all(
-            subscribe_and_fetch
-                .iter()
-                .map(|(pubkey, _)| self.subscribe(pubkey)),
-        )
-        .await;
+        let subscription_results =
+            join_all(pubkeys.iter().map(|pubkey| self.subscribe(pubkey))).await;
```
GabrielePicco: `join_all` makes one stuck subscription block the entire batch. Since each pubkey has already been claimed as pending, a single websocket/RPC subscribe future that never answers can hold all of those pending entries indefinitely. The lower-level subscribe response receive also has no timeout, so the deadline should be the owner task's responsibility, not something waiters discover via a later stale-age check.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -1875,40 +2016,101 @@ where
...
-                if let Err(err) = receiver
-                    .await
-                    .inspect_err(|err| {
-                        warn!(pubkey = %pubkey, error = ?err, "FetchCloner::clone_accounts - RecvError awaiting account, sender dropped without sending value");
-                    })
-                {
-                    // The sender was dropped, likely due to an error in the other request
-                    warn!(
-                        "Failed to receive account from pending request: {err}"
+            match tokio::time::timeout(PENDING_REQUEST_TIMEOUT, receiver).await
```
GabrielePicco: A waiter timing out here does not remove its sender from `PendingRequestState.waiters`, does not cancel the owner, and does not remove the pending request. So callers can wait 60s, fail, and leave the pubkey poisoned until the owner finishes or a later caller triggers 120s stale eviction. That is too slow and still passive. The pending request itself should have a total deadline and complete all waiters exactly once.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -275,6 +321,105 @@ where
...
+                    self.pending_requests.clone(),
+                    pubkey,
+                    generation,
+                ))
+            }
+            Entry::Occupied(mut entry) => {
+                let age = entry.get().created_at.elapsed();
+
+                // Check if entry is stale
+                if age > PENDING_REQUEST_STALE_AFTER {
```
GabrielePicco: This stale eviction is only evaluated when a new caller enters `claim_pending_request`. If no one retries the same pubkey, the stale entry can remain forever. This should not be the mechanism that unsticks production requests. Treat stale eviction as a last-resort assertion/metric, and make the owner operation actively time out and remove its own generation.

magicblock-chainlink/src/remote_account_provider/mod.rs
```diff
@@ -1001,15 +1201,25 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
...
+            // Preserve any existing rollback ownership so the creator can
+            // still roll back this subscription if its batch later fails.
             self.lrucache_subscribed_accounts.add(*pubkey);
-            return Ok(());
+            return Ok(SubscribeResult::AlreadyWatching);
         }
 
-        self.register_subscription(pubkey).await?;
-        Ok(())
+        let evicted = self.register_subscription(pubkey).await?;
```
GabrielePicco: This check-then-await sequence is still race-prone. Two callers can both observe \"not watching\", both enter `register_subscription()`, and then whichever call writes `subscription_rollback_owners` last becomes the sole rollback owner for a shared subscription. If one batch later fails, it can roll back a subscription another successful batch relies on. This wants a per-pubkey subscription state machine/refcount (`Idle`, `Subscribing`, `Active`, `Unsubscribing`) rather than a single rollback token written after an await.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -1875,40 +2016,101 @@ where
...
-                }
-            }
+        // Finish pending requests and dismiss owner guards. On success we
+        // propagate the owner's `FetchAndCloneResult` to all waiters so they
+        // can observe `not_found_on_chain` / `missing_delegation_record`
+        // metadata for the pubkey they were waiting on.
+        for (pubkey, mut guard) in owner_guards {
+            let completion = match &result {
+                Ok(r) => PendingRequestCompletion::Success(r.clone()),
+                Err(err) => PendingRequestCompletion::Failed(
```
GabrielePicco: This reconciliation refetch bypasses `fetch_and_clone_accounts_with_dedup` and also drops the original `mark_empty_if_not_found` semantics by passing `None`. If several waiters fail reconciliation, they can each perform their own fresh fetch, and callers that needed mark-empty behavior may get a different result from the owner request they joined. If this path is needed, it should re-enter the same dedup/semantics-aware operation model rather than calling the raw fetch path.

GabrielePicco (changes requested):
I do not think this should land in its current shape. The PR improves some abort paths, but the production stuck-pubkey failure mode is still mostly handled by passive stale timers. The main issue is that ownership is still tied to caller futures and unbounded awaits, rather than to an owned operation with a hard deadline and guaranteed completion.

The current 30s/60s/120s values are too high, but lowering them alone is not the right fix. A pending pubkey operation should be an owned task with a total budget, cancellation token, generation, and shared completion channel. The owner task should always remove the pending entry and notify waiters with Ok/Err/TimedOut. Stale eviction should be a diagnostic fallback, not a correctness mechanism.

I also think the dedup key is too weak: requests are deduped by pubkey even though `mark_empty_if_not_found`, `min_context_slot`/slot, (e.g. incoming request with higher `min_context_slot`) and fetch semantics affect the result (this may require a huge refactoring in the cloning, so I think we can accept pubkey as keys for now).

Instead of caller-owned pending entries plus stale timers, I believe a better approach is to:

1. Insert `Pending { generation, deadline, waiters, cancel }`.
2. Spawn exactly one owner task for that operation.
3. Run the full lifecycle under `timeout(total_budget, ...)`.
4. The owner task always removes `{pubkey, generation}` and sends a terminal result to every waiter.
5. Waiters only await shared completion; their cancellation removes their waiter, not the owner.
6. Manage subscriptions with per-pubkey state/refcount, not last-writer rollback tokens.

This design makes completion active and bounded by construction.

magicblock-chainlink/src/remote_account_provider/mod.rs
thlorenz: Verified outdated.

magicblock-chainlink/src/remote_account_provider/mod.rs
thlorenz: Verified outdated.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
thlorenz: Verified outdated.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
thlorenz: Verified outdated.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
thlorenz: Verified outdated.

magicblock-chainlink/src/remote_account_provider/mod.rs
thlorenz: Addressed in 1df65a53

magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
```diff
@@ -254,5 +252,28 @@ where
         });
     }
 
+    let ata_pubkeys = atas
+        .iter()
+        .map(|(ata_pubkey, _, _, _)| *ata_pubkey)
+        .collect::<HashSet<_>>();
+    let mut releases = pubkeys_to_subscribe
```
GabrielePicco, line 255: The `DirectAccount` release still filters out ATA pubkeys. Each ATA is added to `pubkeys_to_subscribe` at `ata_projection.rs:107` and then fetched through `task_to_fetch_with_companion` at `ata_projection.rs:121`, which fetches `[ata, eata]` and can add another `DirectAccount` count for the ATA. The final delegated-account release only decrements one direct count, so a projected delegated ATA can remain subscribed even though delegated ATAs should not be watched.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -901,11 +1028,19 @@ where
                     })) => {
                         // We may need to remove temporary subscriptions created
                         // while resolving this update.
-                        let mut subs_to_remove = HashSet::new();
+                        let mut subs_to_remove = Vec::new();
 
-                        // Always unsubscribe from delegation record if it was a new subscription
                         if !was_delegation_record_subscribed {
```
GabrielePicco: This still skips the `DirectAccount` release when the delegation record was already watched. The fetch at `mod.rs:1014` can acquire a new direct ref even for an already-watched record, so the newly acquired ref needs to be balanced independently of the prior watch state.

magicblock-chainlink/src/chainlink/fetch_cloner/delegation.rs
```diff
@@ -214,6 +233,7 @@ where
         Err(_) => None,
     };
 
+    let mut releases = Vec::new();
     if !was_watching_deleg_record
```
GabrielePicco: The helper still releases the delegation-record `DirectAccount` ref only when `!was_watching_deleg_record`. Since `try_get_multi_until_slots_match` can acquire a new direct ref for this fetch, repeated calls can leak refs for watched delegation records.

magicblock-chainlink/src/chainlink/fetch_cloner/program_loader.rs
```diff
@@ -83,4 +121,13 @@ pub(crate) async fn handle_executable_sub_update<T, U, V, C>(
     if let Err(err) = this.cloner.clone_program(loaded_program).await {
         warn!(pubkey = %pubkey, error = %err, "Failed to clone program into bank");
     }
+
+    if acquired_program_data_reason {
+        // Both refs exist for LoaderV3 program-data cleanup.
```
GabrielePicco, line 134: Fetch-error cleanup was added, but the resolver-error branch still returns before `release_program_data_subs` runs. If `ProgramAccountResolver::try_new` fails after the LoaderV3 program-data reason and fetch-owned direct ref were acquired, both temporary refs remain held.

GabrielePicco (commented):
LGTM, with some problems to address

magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
thlorenz: Fixed cccc7c1f

magicblock-chainlink/src/chainlink/fetch_cloner/delegation.rs
thlorenz: Fixed e1b4a476

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
thlorenz: Fixed cba160dd

magicblock-chainlink/src/chainlink/fetch_cloner/program_loader.rs
thlorenz: Fixed 96281d6e

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
GabrielePicco, line 1499: acquires top-level DirectAccount ownership through `try_get_multi`, but the fallible resolver calls return with ? before the normal cleanup at magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs:1742.

GabrielePicco (commented):
LGTM, 

one more problem to address

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
thlorenz: Fixed in eb8930c3

GabrielePicco (approved):
LGTM!

PR #1174, opened 2026-05-01
feat: invoke CommitFinalize by default instead of Commit+Finalize

snawaz:
This PR changes (non-breaking) scheduled commits to use the DLP `CommitFinalize` instruction instead of executing separate `Commit` and `Finalize` instructions. 

- The user-facing behavior stays the same: a scheduled commit still commits the account state to base and finalizes it. 
- For commit-and-undelegate flows, the undelegate step is still preserved after commit-finalize.

## Naming problem

`MagicBlockInstruction::ScheduleCommit` was already a misleading name because it **never** meant _"commit only"_; it always scheduled a commit followed by finalize. But now, with this PR, the codebase feels even more misleading because the underlying DLP operation is now explicitly `CommitFinalize` while we also have `Commit` (which also executes `Finalize` implicitly). 

### Possible fix

Rename the existing `ScheduleCommit` to `ScheduleCommitFinalize` since that is what it actually does (even before this PR). Rename the currently _unused_ `ScheduleCommitFinalize` enum slot to `ScheduleCommit` (or `ScheduleCommitOnly`) and **leave it unimplemented for now because commit-only scheduling semantics are not yet clear and we never supported that**. This keeps the enum shape stable while making the active API naming accurate.

## Summary by CodeRabbit

* **New Features**
  * Added an execution-mode option to choose combined commit+finalize vs separate flows
  * Finalize stage is skipped automatically when no finalize work is needed, reducing overhead

* **Bug Fixes**
  * Consistent error mapping so finalize-related tasks surface the same transaction errors as commit tasks
  * Improved recovery and logging for unfinalized-account and commit-nonce/id error cases

* **Tests**
  * Updated unit/integration tests for single- vs two-stage behavior; increased CPI-limit thresholds and marked two commit-unfinalized-account recovery tests as ignored

### DIFF
--- a/magicblock-committor-service/src/intent_executor/single_stage_executor.rs
+++ b/magicblock-committor-service/src/intent_executor/single_stage_executor.rs
@@ -22,10 +22,7 @@ use crate::{
         IntentExecutionReport,
     },
     persist::{IntentPersister, IntentPersisterImpl},
-    tasks::{
-        commit_task::CommitTask, task_strategist::TransactionStrategy,
-        BaseTaskImpl, FinalizeTask,
-    },
+    tasks::{task_strategist::TransactionStrategy, BaseTaskImpl, FinalizeTask},
     transaction_preparator::TransactionPreparator,
 };
 
@@ -217,12 +214,31 @@ where
             ) => {
                 let optimized_tasks =
                     self.transaction_strategy.optimized_tasks.as_slice();
-                if let Some(BaseTaskImpl::Commit(task)) = err
+                if let Some(task) = err
                     .task_index()
                     .and_then(|index| optimized_tasks.get(index as usize))
                 {
+                    let delegated_account = match task {
+                        BaseTaskImpl::Commit(task) => {
+                            task.committed_account.pubkey
+                        }
+                        BaseTaskImpl::CommitFinalize(task) => {
+                            task.committed_account.pubkey
+                        }
+                        _ => {
+                            error!(
+                                task_index = err.task_index(),
+                                optimized_tasks_len = optimized_tasks.len(),
+                                error = ?err,
+                                "RPC returned unexpected task index"
+                            );
+                            return Ok(ControlFlow::Break(()));
+                        }
+                    };
                     self.handle_unfinalized_account_error(
-                        signature, task, transaction_preparator
+                        signature,
+                        delegated_account,
+                        transaction_preparator,
                     )
                     .await
                 } else {
@@ -262,13 +278,11 @@ where
     async fn handle_unfinalized_account_error<T: TransactionPreparator>(
         &self,
         failed_signature: &Option<Signature>,
-        task: &CommitTask,
+        delegated_account: Pubkey,
         transaction_preparator: &T,
     ) -> IntentExecutorResult<ControlFlow<(), TransactionStrategy>> {
-        let finalize_task: BaseTaskImpl = FinalizeTask {
-            delegated_account: task.committed_account.pubkey,
-        }
-        .into();
+        let finalize_task: BaseTaskImpl =
+            FinalizeTask { delegated_account }.into();
         prepare_and_execute_strategy(
             &self.intent_client,
             &self.authority,
--- a/magicblock-committor-service/src/intent_executor/single_stage_executor.rs
+++ b/magicblock-committor-service/src/intent_executor/single_stage_executor.rs
@@ -214,27 +214,14 @@ where
             ) => {
                 let optimized_tasks =
                     self.transaction_strategy.optimized_tasks.as_slice();
-                if let Some(task) = err
+                if let Some(delegated_account) = err
                     .task_index()
                     .and_then(|index| optimized_tasks.get(index as usize))
-                {
-                    let delegated_account = match task {
-                        BaseTaskImpl::Commit(task) => {
-                            task.committed_account.pubkey
-                        }
-                        BaseTaskImpl::CommitFinalize(task) => {
-                            task.committed_account.pubkey
-                        }
-                        _ => {
-                            error!(
-                                task_index = err.task_index(),
-                                optimized_tasks_len = optimized_tasks.len(),
-                                error = ?err,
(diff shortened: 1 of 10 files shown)


snawaz:
* **#1174** <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1174?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1174?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

magicblock-committor-service/src/intent_executor/two_stage_executor.rs
```diff
@@ -206,10 +203,31 @@ where
...
+                    let delegated_account = match task {
+                        BaseTaskImpl::Commit(task) => {
+                            task.committed_account.pubkey
+                        }
+                        BaseTaskImpl::CommitFinalize(task) => {
+                            task.committed_account.pubkey
+                        }
+                        _ => {
+                            error!(
+                                task_index = ?task_index,
```
taco-paco: This is practically duplicates `else` block below.

This can be avoided simply by
```rust
let optimized_tasks =
    self.transaction_strategy.optimized_tasks.as_slice();
if let Some(delegate_account) = err
    .task_index()
    .and_then(|index| optimized_tasks.get(index as usize))
    .and_then(|el| {
        match el {
            BaseTaskImpl::Commit(task) => {
                Some(task.committed_account.pubkey)
            }
            BaseTaskImpl::CommitFinalize(task) => {
                Some(task.committed_account.pubkey)
            }
            _=> None
        }
    })
{
    self.handle_unfinalized_account_error(
        signature, delegate_account, transaction_preparator
    )
    .await
} else {
    error!(
        task_index = err.task_index(),
        optimized_tasks_len = optimized_tasks.len(),
        error = ?err,
        "RPC returned unexpected task index"
    );
    Ok(ControlFlow::Break(()))
}
```

magicblock-committor-service/src/intent_executor/single_stage_executor.rs
```diff
@@ -217,12 +214,31 @@ where
...
                 {
+                    let delegated_account = match task {
+                        BaseTaskImpl::Commit(task) => {
+                            task.committed_account.pubkey
+                        }
+                        BaseTaskImpl::CommitFinalize(task) => {
+                            task.committed_account.pubkey
+                        }
+                        _ => {
+                            error!(
```
taco-paco: This is practically duplicates `else` block below.

This can be avoided simply by
```rust
let optimized_tasks =
    self.transaction_strategy.optimized_tasks.as_slice();
if let Some(delegate_account) = err
    .task_index()
    .and_then(|index| optimized_tasks.get(index as usize))
    .and_then(|el| {
        match el {
            BaseTaskImpl::Commit(task) => {
                Some(task.committed_account.pubkey)
            }
            BaseTaskImpl::CommitFinalize(task) => {
                Some(task.committed_account.pubkey)
            }
            _=> None
        }
    })
{
    self.handle_unfinalized_account_error(
        signature, delegate_account, transaction_preparator
    )
    .await
} else {
    error!(
        task_index = err.task_index(),
        optimized_tasks_len = optimized_tasks.len(),
        error = ?err,
        "RPC returned unexpected task index"
    );
    Ok(ControlFlow::Break(()))
}
```

magicblock-committor-service/src/tasks/task_strategist.rs
```diff
@@ -128,6 +128,15 @@ impl TaskStrategist {
     ) -> TaskStrategistResult<StrategyExecutionMode> {
         const MAX_UNITED_TASKS_LEN: usize = 22;
 
+        if commit_tasks.is_empty() || finalize_tasks.is_empty() {
```
taco-paco: What do empty `commit_tasks` mean? From what i know this shouldn't be possible and if it is should be an error

taco-paco (changes requested):
Where could I read on task that this PR closes?

From the title it would seem that changes shall come from `magic-program` and not by setting `mode` for building tasks.

Users shall be able to choose the flow: standart, commit+finalize. Now if they choose standard(old) flow we shouldn't implicitly convert it for them.

The change that I see:

1. `process_schedule_commit practically` redirects into `process_schedule_commit_finalize`. We communicate this to users
2. Users still have an option to create standart commit flow via `MagicIntentBundleBuilder`.

Both are supported

magicblock-committor-service/src/intent_executor/two_stage_executor.rs
snawaz: Thanks. Simplified.

magicblock-committor-service/src/intent_executor/single_stage_executor.rs
snawaz: Thanks. Simplified.

magicblock-committor-service/src/tasks/task_strategist.rs
snawaz: It was added temporarily. Now it has been removed. Thanks.

snawaz:
> Where could I read on task that this PR closes?
> 
I've added the details in the description. Please have a look at it.

PS: I accidentally edited your comment while replying to it. (Quite a questionable GitHub UX decision to allow editing someone else's comment by mistake!)

taco-paco (changes requested):
Reading the description the implementation doesn't click for me. 
According to my understanding the only required change is to change `process_schedule_commit`, where it would create CommitFinalize intent instead of old Commit.

This PR forces a conversion within committor service of one Intent type into another.

Imo user still have to be able to create Intent with regular commit/finalize flow, which now they can't. We shouldn't implicetly cast them for it, it is their choice. If they want they shall be able to use `MagicIntentBundleBuilder` to create on old flow intent

snawaz:
Discard in favor of #1191

PR #1175, opened 2026-05-01
fix: cache known-empty eATAs and never subscribe SPL Token program

GabrielePicco:
## Summary

Two related fixes in `fetch_cloner` that together eliminate the dominant source of upstream JSON-RPC traffic from the validator (observed at ~21 not-found fetches/s, peaking around ~111/s = the `400k req/h` we were seeing on the upstream provider — 99.8% of all upstream account fetches with a 0% find rate):

- **Negative cache for known-empty eATAs.** Every subscription update for any SPL ATA used to fire a fresh `try_get_multi_until_slots_match(... ProjectAta)` for the derived eATA. The vast majority of cloned ATAs (wSOL ATAs, memecoin ATAs from active trader/sniper bots) have no corresponding eATA on chain, so each update produced a guaranteed not-found upstream call. We now keep a bounded LRU set of eATA pubkeys that have already been observed as non-existent and short-circuit the fetch on subsequent updates. **Crucially, the cache check happens *after* the (idempotent) `subscribe_to_account(eata)` call**, so the subscription LRU entry keeps getting promoted on every ATA update — if the eATA is created on chain later, the existing eATA-side subscription path (`maybe_build_projected_ata_clone_request_from_eata_sub_update`) projects the ATA without us needing to invalidate the cache.

- **Never subscribe to SPL Token / Token-2022 / Associated-Token-Account programs.** The existing blacklist (`native_program_accounts`) didn't cover these.

### Live evidence (mainnet-usa, before fix)

```
mbv_account_fetches_found_count{origin="project_ata"}     0
mbv_account_fetches_not_found_count{origin="project_ata"} 6,969,110
project_ata share of all upstream fetches                 99.8%
```

Hot accounts in the validator log decode to wSOL ATAs and pump.fun memecoin ATAs across a small cluster of trader wallets — nothing delegation-related, just incidental ATAs that the validator cloned and stayed subscribed to.

## Summary by CodeRabbit

* **New Features**
  * Optimized subscription handling to avoid subscribing to core token programs.
  * Added a bounded negative cache for empty associated token accounts to skip redundant fetches.

* **Tests**
  * Added tests covering caching behavior for empty associated token accounts.
  * Added tests ensuring delegated accounts owned by token programs do not trigger subscriptions.

### DIFF
--- a/magicblock-chainlink/src/chainlink/blacklisted_accounts.rs
+++ b/magicblock-chainlink/src/chainlink/blacklisted_accounts.rs
@@ -77,9 +77,7 @@ pub fn native_program_accounts() -> HashSet<Pubkey> {
     blacklisted_programs
 }
 
-/// Programs that must never be subscribed to via `subscribe_program`.
-/// They own a vast number of accounts on chain and a program-wide
-/// subscription would flood the validator with unrelated updates.
+/// High-cardinality programs skipped for program-wide subscriptions.
 pub fn programs_not_to_subscribe() -> HashSet<Pubkey> {
     let mut programs = HashSet::new();
     programs.insert(TOKEN_PROGRAM_ID);
--- a/magicblock-chainlink/src/chainlink/blacklisted_accounts.rs
+++ b/magicblock-chainlink/src/chainlink/blacklisted_accounts.rs
@@ -1,5 +1,8 @@
 use std::collections::HashSet;
 
+use magicblock_core::token_programs::{
+    ASSOCIATED_TOKEN_PROGRAM_ID, TOKEN_PROGRAM_ID,
+};
 use magicblock_magic_program_api::{self as magic_program};
 use solana_pubkey::Pubkey;
 use solana_sdk_ids::{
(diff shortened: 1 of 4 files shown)


magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -1062,6 +1104,17 @@ where
         })
     }
 
+    fn is_known_empty_eata(&self, eata_pubkey: &Pubkey) -> bool {
+        // `parking_lot::Mutex` does not poison; `lock()` returns the guard
+        // directly so no `unwrap`/`expect` is needed on the hot path.
+        // `get` promotes on hit so frequently revisited entries stay warm.
+        self.known_empty_eatas.lock().get(eata_pubkey).is_some()
```
thlorenz, line 1094: Do we need all this explanation of inner workings of the parking lot mutex and such here?

magicblock-chainlink/src/chainlink/blacklisted_accounts.rs
```diff
@@ -73,3 +76,14 @@ pub fn native_program_accounts() -> HashSet<Pubkey> {
     blacklisted_programs.insert(zk_elgamal_proof_program::ID);
     blacklisted_programs
 }
+
+/// Programs that must never be subscribed to via `subscribe_program`.
+/// They own a vast number of accounts on chain and a program-wide
+/// subscription would flood the validator with unrelated updates.
+pub fn programs_not_to_subscribe() -> HashSet<Pubkey> {
```
thlorenz, line 81: Should we create an issue to add more programs here (investigating a bit more to find problematic programs _before_ we see them in prod)?

thlorenz (approved):
LGTM with minor nit/question.

magicblock-chainlink/src/chainlink/blacklisted_accounts.rs
GabrielePicco: Yes, will keep monitoring and create and issue

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
GabrielePicco: Too verbose indeed. Refactored

lucacillario (approved):
LGTM 👍

PR #1190, opened 2026-05-12
feat: batch cranks

### LINKED ISSUE #1189: feat: enhance crank perfs
### Problem

Cranks are sometimes too slow to trigger

### Proposed Solution

Batch execution of tasks

Dodecahedr0x:
Closes #1189 

## Summary
- Batch crank executions and stop waiting for them to finish before processing more cranks.
- Batch write to SQLite when a batch finishes instead of one write per execution
- Fixes setting the next execution to account for wake-up delay

**Up to 3x speedup for cranks, especially underload**

## Breaking Changes
- [X] None

## Summary by CodeRabbit

* **Documentation**
  * Updated task scheduler docs and README: switched to min-interval setting and expanded performance guidance for batching, concurrency, SQLite tuning, payload handling, and retry/backoff strategies.

* **Dependencies**
  * Bumped solana-pubkey from 3.0 to 4.1.

* **Improvements / Refactor**
  * Scheduler now batches and executes due tasks concurrently, applies batch results atomically, tracks per-task update timestamps to avoid stale changes, records failures for clearer retry logic, and shares RPC/counter resources for reduced cloning and improved throughput.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1190)


magicblock-task-scheduler/src/service.rs
```diff
@@ -145,42 +160,88 @@ impl TaskSchedulerService {
...
+        let (crank_tx, mut crank_rx) = mpsc::unbounded_channel();
+
+        loop {
+            select! {
+                Some(expired) = self.task_queue.next() => {
+                    // A task expired, batch all expired tasks
+                    let first = expired.into_inner();
+                    self.task_queue_keys.remove(&first.id);
+                    let mut batch = vec![first];
+                    let waker = noop_waker();
```
taco-paco: I'd recommend using parent context. Similar can be achieved via

while let Some(expired) = poll_fn(|cx| self.task_queue.poll_expired(cx))
    .now_or_never()
    .flatten()

magicblock-task-scheduler/src/service.rs
```diff
@@ -145,42 +160,88 @@ impl TaskSchedulerService {
...
+                            warn!("Failed to process request ID={}: {e:?}", id);
+                        }
+                        Err(e) => {
+                            error!("Failed to process request: {}", e);
+                            return Err(e);
+                        }
+                    }
+                }
+                _ = self.token.cancelled() => {
+                    break;
```
taco-paco, line 251: Shouldn't we finalize results first prior to existing? Say `executions_left` would become inconsistent on restart as some result could be ready in `crank_rx`

GabrielePicco (approved):

magicblock-task-scheduler/src/service.rs
Dodecahedr0x: Addressed in [2a7ad35](https://github.com/magicblock-labs/magicblock-validator/pull/1190/commits/2a7ad35d04d2429e71d9f58f876039b2e5a56128)

magicblock-task-scheduler/src/service.rs
Dodecahedr0x: Addressed in [48790e5](https://github.com/magicblock-labs/magicblock-validator/pull/1190/commits/48790e5a72f00ff1ca406732f0373afc94bc1123)

magicblock-task-scheduler/src/service.rs
```diff
@@ -145,42 +159,100 @@ impl TaskSchedulerService {
...
+                }
+            }
+        }
+
+        info!("TaskSchedulerService shutdown!");
+        for _ in 0..inflight_batches {
+            let Some((batch, result)) = crank_rx.recv().await else {
```
taco-paco: I think that would do a job without needing new `inflight_batches` variable unless I'm missing something
```suggestion
            while let Some((batch, result)) = crank_rx.recv().await {
```

taco-paco (approved):
LGTM! left a small comment

magicblock-task-scheduler/src/service.rs
Dodecahedr0x: I don't think it works like you are suggesting:
- Just replacing the statement would drain multiple at once, then block forever because of the outter for loop
- Replacing the for loop with this would also block forever for the same reason, None is only returned once the channel is closed

But I can simplify by dropping the channel at shutdown then draining it

PR #1194, opened 2026-05-13
fix: exausted cranks remaining in the DB

GabrielePicco:
## Summary

Fixes task scheduler stale completed-task rows blocking new cranks with the same ID.

  - Treats tasks DB rows as active tasks only.
  - Removes a task from the DB after its final execution instead of leaving it with executions_left = 0.
  - Purges non-active persisted tasks on scheduler startup.
  - Keeps different-authority replacement rejected while an existing task is still active.
  - Updates scheduler integration tests to assert completed tasks are removed, while reschedule and unauthorized replacement behavior remain covered.

## Release Notes

* **Behavior Changes**
  * Completed tasks are now automatically removed from the database after execution finishes.
  * Task IDs can be reused by different authorities once the original task completes.
  * Improved startup cleanup to remove expired tasks with zero remaining executions.

* **Improvements**
  * Enhanced task database maintenance for better resource management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1194)

### DIFF
--- a/magicblock-task-scheduler/src/service.rs
+++ b/magicblock-task-scheduler/src/service.rs
@@ -124,7 +124,7 @@ impl TaskSchedulerService {
         for task in tasks {
             debug!("Task: {:?}", task);
             if !is_valid_task_interval(task.execution_interval_millis)
-                || task.executions_left < 0
+                || task.executions_left <= 0
             {
                 warn!(
                     "Task {} has an invalid parameters: (interval={}, executions_left={}). Skipping.",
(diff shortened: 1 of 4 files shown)


magicblock-task-scheduler/src/service.rs
```diff
@@ -274,7 +278,9 @@ impl TaskSchedulerService {
 
         // Check if the task already exists in the database
         if let Some(db_task) = self.db.get_task(task.id).await? {
-            if db_task.authority != task.authority {
+            if db_task.executions_left > 0
```
Dodecahedr0x: If we remove tasks with 0 executions, how can this happen? I guess it's defensive

Dodecahedr0x (approved):
LGTM

magicblock-task-scheduler/src/service.rs
GabrielePicco: Fair, removed

PR #1196, opened 2026-05-14
fix: unique txs for callbacks with same data and within same blockchash

taco-paco:
## Breaking Changes
- [x] None

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced callback transaction processing to ensure each transaction is uniquely identified, improving overall reliability and preventing potential transaction handling issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1196)

### DIFF
--- a/magicblock-services/src/actions_callback_service.rs
+++ b/magicblock-services/src/actions_callback_service.rs
@@ -1,4 +1,7 @@
-use std::sync::Arc;
+use std::sync::{
+    atomic::{AtomicU64, Ordering},
+    Arc,
+};
 
 use futures_util::future::join_all;
 use magicblock_core::{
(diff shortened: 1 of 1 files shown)


magicblock-services/src/actions_callback_service.rs
```diff
@@ -61,21 +64,32 @@ impl<L: LatestBlockProvider> ActionsCallbackService<L> {
         signature: Option<Signature>,
         result: ActionResult,
     ) -> Vec<Result<VersionedTransaction, CallbackScheduleError>> {
+        /// Counter used to make each callback transaction unique
+        static TX_COUNTER: AtomicU64 = AtomicU64::new(0);
```
snawaz, line 68: If the node restarts, then we risk having non-unique tx?

Can we use combine time in some way, or use random number?

snawaz: ah. Nevermind. If the node restarts, the blockhash will change by that time. So it's an unlikely scenario!

snawaz (approved):
LGTM

Dodecahedr0x (approved):
LGTM

magicblock-services/src/actions_callback_service.rs
taco-paco: Yeap

PR #1201, opened 2026-05-15
feat: normalize gRPC backfill

thlorenz:
## Summary

Normalize gRPC backfill handling so every gRPC endpoint is treated as backfill-capable, and all subscription paths consistently derive `from_slot` from the shared chain slot.

### magicblock-chainlink

- Removed provider-specific gRPC endpoint gating and backfill capability detection. Generic gRPC endpoints with an API key are now accepted instead of only Helius/Triton-shaped URLs.
- Made `ChainSlot::compute_from_slot()` return a concrete slot value, saturating to `0` during bootstrap or low-slot cases.
- Updated account, optimization, and program subscription flows so gRPC requests always set `from_slot` using the chain-slot-derived value.
- Simplified gRPC client behavior now that all gRPC providers are considered backfill-capable from the caller’s perspective.
- Cleaned up stale provider-specific wording and regression tests around optional/missing `from_slot` behavior.

## Summary by CodeRabbit

* **Breaking Changes**
  * gRPC endpoints now require explicit API key validation instead of relying on endpoint allowlists.

* **Refactor**
  * Simplified chain slot and backfill initialization by removing optional configurations. Chain slot tracking is now always required, and backfill support is now unconditionally enabled across all provider implementations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1201)


GabrielePicco (approved):
LGTM!

PR #1232, opened 2026-05-20
feat: coordination mode aware crank

### LINKED ISSUE #1231: feat: crank is not aware of coordination mode
### Problem

The crank service is started even if in replica mode where it sends transactions that are garanteed to fail

### Proposed Solution

Disable the crank service for replicas

Dodecahedr0x:
Closes #1231 

## Summary
Disables the crank service when the node starts in replica mode

## Breaking Changes
- [X] None

## Summary by CodeRabbit

* **New Features**
  * Scheduler now waits for coordination to settle, loads persisted tasks (resetting in-memory scheduling state) and only begins scheduling when in primary mode; in standby/replica mode startup returns immediately.

* **Tests**
  * Added/updated serialized tests to validate startup behavior across coordination states, including standby-return behavior and persisted-task reload.

* **Chores**
  * Added a dev dependency to support serialized test execution.


magicblock-task-scheduler/src/service.rs
```diff
@@ -134,6 +136,20 @@ impl TaskSchedulerService {
     pub async fn start(
         mut self,
     ) -> TaskSchedulerResult<JoinHandle<TaskSchedulerResult<()>>> {
+        if self.is_primary_mode().await {
```
taco-paco: I'd suggest doing it on calling site level(magic_validator.rs). It seems like task scheduler wether works or not, so from point crank service it shouldn't be aware of mode at the moment.

taco-paco (commented):
small comment on design, but never the less this should work as well

bmuddha (approved):
LGTM, but may addressing the concern by @taco-paco is worth it before the merge

PR #1238, opened 2026-05-22
fix: disable Chainlink for replicas

### LINKED ISSUE #1203: bug: disable chainlink and committor service outside of primary mode
### What happened

Currently chainlink sets up subscriptions and attempts clones irregardless of the replication mode. From correctness perspective those actions should only be undertaken in primary mode. Thus it is required to introduce mode awareness (CoordinationMode) into the chainlink, which will keep the chainlink on standby without performing any operation until the node switches to primary. 

If it's possible to disable static subscriptions (like clock and delegation program) for non-primary modes, and then quickly set them up when the mode switch occurs, it should be done, otherwise they can be kept around. 

In replica only mode the chainlink can permanently disabled, since no takeover is possible. 

It's also imperative to double check that committor service also doesn't perform any action in non-primary mode. 

### Commit / version

3b7574952c53ab45e270783599c86b29df73d74c

### Deployment

Local (native)

### OS / Arch

_No response_

### Logs / Links

_No response_

thlorenz:
## Summary

Disable Chainlink-dependent behavior when the validator runs as a replica, while keeping Chainlink enabled for standalone and primary modes.

CLOSES: #1203

**NOTE**: this does not disable the comittor service in those replica modes, but it totally could. That part could be done in a follow up PR if we want to implement it.

### Chainlink

Adds a replication-mode-aware Chainlink wrapper that can run either with the full Chainlink stack or in a disabled mode for replicas. Disabled mode avoids remote account fetching and subscriptions while preserving the account-bank reset behavior needed during replication startup.

### Aperture RPC

Gates RPC paths that require on-chain interaction so replicas serve local account reads but reject primary-only transaction flows with a clear error. This prevents replica nodes from attempting Chainlink-backed fetches or transaction account preparation.

### Validator and replication services

Initializes disabled Chainlink for replica mode and passes a reset-capable Chainlink handle into replication instead of relying on a Chainlink-shaped stub. Replication services are made generic over the account-bank resetter interface so they only depend on the cleanup behavior they need.

### Tests

Adds coverage for mode-aware Chainlink behavior, replica RPC gating, validator replication-mode selection, and the resetter plumbing used by replication services.

## Release Notes

* **New Features**
  * RPC methods now enforce primary validator mode requirements for transaction operations (sendTransaction, simulateTransaction).
  * Added transaction shape validation to reject v0 transactions with address lookup tables and enforce program ID index constraints.

* **Improvements**
  * Enhanced replication mode awareness across validator chainlink operations for better multi-node support.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1238?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


thlorenz:
@coderabbitai review

thlorenz:
@coderabbit resume

magicblock-aperture/src/requests/http/mod.rs
```diff
@@ -361,3 +392,168 @@ pub(crate) mod simulate_transaction;
 
 // Magic Router compatibility methods.
 pub(crate) mod get_delegation_status;
+
+#[cfg(test)]
```
bmuddha: these tests verify something which is obvious from reading the code, they are overly verbose and add very little value, if you think they are absolutely necessary, then please add a proper full coverage in tests directory to respective modules

magicblock-api/src/magic_validator.rs
```diff
@@ -486,6 +479,17 @@ impl MagicValidator {
         latest_block: &LatestBlock,
         accountsdb: &Arc<AccountsDb>,
     ) -> ApiResult<ChainlinkImpl> {
+        if Self::replication_mode_uses_disabled_chainlink(
+            &config.validator.replication_mode,
+        ) {
+            return ChainlinkImpl::disabled(
```
bmuddha, line 484: arguably could've just used Option<T> to indicate disabled state

magicblock-replicator/src/service/context.rs
```diff
@@ -27,7 +28,10 @@ use crate::{
 };
 
 /// Shared state for both primary and replica roles.
-pub struct ReplicationContext {
+pub struct ReplicationContext<R>
```
bmuddha: please remove the generic, it's no longer needed, as we do not perform the reset

magicblock-replicator/src/service/mod.rs
```diff
@@ -57,12 +58,18 @@ const CONSUMER_RETRY_DELAY: Duration = Duration::from_secs(1);
 // =============================================================================
 
 /// Replication service for the selected replication role.
-pub enum Service {
-    Primary(Primary),
-    Replica(Replica),
+pub enum Service<R>
```
bmuddha: ditto, no generic needed, master code doesn't perform the reset or transitions between modes

magicblock-replicator/src/service/primary.rs
```diff
@@ -2,6 +2,7 @@
 
 use std::time::Duration;
 
+use magicblock_chainlink::AccountsBankResetter;
```
bmuddha: ditto: no need for the trait

magicblock-replicator/Cargo.toml
```diff
@@ -13,6 +13,7 @@ bincode = { workspace = true }
 bytes = { workspace = true }
 futures = { workspace = true }
 magicblock-accounts-db = { workspace = true }
+magicblock-chainlink = { workspace = true }
```
bmuddha: no need for the dependency

bmuddha (changes requested):

magicblock-api/src/magic_validator.rs
thlorenz: Not as nice as all callers would have to be aware of both cases vs. here they just call the same methods without knowledge that in the _disabled_ state they are essentially a noop.

magicblock-replicator/src/service/context.rs
thlorenz: Good catch. I originally added the trait when the primary
transition still performed `reset_accounts_bank()`, so the
generic was acting as a narrow seam around that one call.

After the rebase/master changes, the reset is done earlier
in `MagicValidator::start` after ledger replay and before
the replication service is spawned. That means the
replicator no longer needs to know about account-bank reset
at all, and keeping the generic would be stale plumbing /
potentially duplicate the reset. I’ll remove the generic and
keep the reset responsibility in validator startup.

thlorenz: Addressed in 1838a2f2 and 4d87c3d6

magicblock-replicator/src/service/mod.rs
thlorenz: ditto addressed

magicblock-replicator/src/service/primary.rs
thlorenz: ditto addressed

magicblock-replicator/Cargo.toml
thlorenz: ditto addressed

magicblock-aperture/src/requests/http/mod.rs
thlorenz: Addressed in a9c7c7a5..6376c8ce

bmuddha (approved):
LGTM

PR #1240, opened 2026-05-22
Fix eATA projection min-context refetch

GabrielePicco:
## Summary
- Honor an explicit `min_context_slot` when refetching accounts for slot-matching, so projection refetches do not start from a lagging provider chain slot.
- Retry custom minimum-context-slot RPC errors instead of treating them as terminal fetch failures.
- Add regression coverage for an ATA subscription update that arrives ahead of provider chain slot and must still project from the delegated eATA.

## Root Cause
The ATA projection path requested the derived eATA with `min_context_slot` from the ATA subscription update, but the provider refetch path still used the current provider `chain_slot`. When that slot lagged behind the update, the eATA fetch could fail before RPC caught up, causing the base ATA to be cloned instead of the projected delegated ATA.

## Validation
- `cargo test -p magicblock-chainlink test_ata_subscription_update_projects_eata_when_chain_slot_lags -- --nocapture`
- `cargo test -p magicblock-chainlink test_get_accounts_until_slots_match -- --nocapture`
- `cargo test -p magicblock-chainlink --lib`

## Release Notes

* **Bug Fixes**
  * Improved handling of RPC errors related to minimum context slot synchronization requirements
  * Enhanced retry logic for account fetches to respect stricter context slot requirements

* **Tests**
  * Added test coverage for ATA subscription updates when chain slots lag behind expectations
  * Improved error handling in mock client for multi-account fetch operations

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1240?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/magicblock-chainlink/src/chainlink/fetch_cloner/tests.rs
+++ b/magicblock-chainlink/src/chainlink/fetch_cloner/tests.rs
@@ -3091,6 +3091,112 @@ async fn test_delegated_eata_subscription_update_clones_raw_eata_and_projects_at
 }
 
 #[tokio::test]
+async fn test_ata_subscription_update_projects_eata_when_chain_slot_lags() {
+    init_logger();
+    let validator_keypair = Keypair::new();
+    let validator_pubkey = validator_keypair.pubkey();
+    let wallet_owner = random_pubkey();
+    let mint = random_pubkey();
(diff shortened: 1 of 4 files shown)


thlorenz (approved):
LGTM

PR #1242, opened 2026-05-25
Fix same-slot account refetch

GabrielePicco:
## Summary
- retry same-slot account resolution with a single RPC batch after a mixed-source mismatch. This avoids slots never matching due to RPC/Ws races. Observed in production, resulting in "RPC call timeout. Retrying..." errors 
- remove the extra non-waiter refetch that could amplify RPC load
- add structured timeout/slot-mismatch logging and coverage for mixed subscription/fetch results

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RPC fetch reliability with configurable retry logic and timeout handling.
  * Enhanced slot-matching validation with automatic re-fetching when compatibility issues occur.

* **Improvements**
  * Better error messages for account resolution failures.
  * Richer diagnostic logging during account fetch operations, including retry attempts and timing information.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1242?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/magicblock-chainlink/src/remote_account_provider/mod.rs
+++ b/magicblock-chainlink/src/remote_account_provider/mod.rs
@@ -455,8 +455,17 @@ fn next_match_slots_retry(
     Ok(match_slots_retry_delay(config))
 }
 
+fn next_match_slots_rpc_error_retry(
+    retries: &mut u64,
+    start: std::time::Instant,
+    config: &MatchSlotsConfig,
+) -> Result<Duration, String> {
+    next_match_slots_retry(retries, start, config)
+        .map(|delay| delay.max(RPC_FETCH_RETRY_DELAY))
+}
+
 fn match_slots_retry_delay(config: &MatchSlotsConfig) -> Duration {
-    Duration::from_millis(config.retry_interval_ms).max(RPC_FETCH_RETRY_DELAY)
+    Duration::from_millis(config.retry_interval_ms)
 }
 
 impl
@@ -1005,8 +1014,11 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
             {
                 Ok(remote_accounts) => remote_accounts,
                 Err(err) => {
-                    let retry =
-                        next_match_slots_retry(&mut retries, start, &config);
+                    let retry = next_match_slots_rpc_error_retry(
+                        &mut retries,
+                        start,
+                        &config,
+                    );
                     warn!(
                         pubkeys = %pubkeys_str(pubkeys),
                         min_context_slot = ?config.min_context_slot,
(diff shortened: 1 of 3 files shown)


magicblock-chainlink/src/remote_account_provider/mod.rs
```diff
@@ -350,6 +350,9 @@ unsafe impl Sync for ForwardedSubscriptionUpdate {}
 
 // Not sure why helius uses a different code for this error
 const HELIUS_CONTEXT_SLOT_NOT_REACHED: i64 = -32603;
+const RPC_FETCH_MAX_RETRIES: u64 = 3;
```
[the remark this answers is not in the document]
GabrielePicco: Done on purpose

magicblock-chainlink/src/remote_account_provider/mod.rs
```diff
@@ -1567,22 +1651,35 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
...
-        const MAX_RETRIES: u64 = 10;
-        const RPC_CALL_TIMEOUT: Duration = Duration::from_secs(2);
-
         let rpc_client = self.rpc_client.clone();
         let fetching_accounts = self.fetching_accounts.clone();
         let commitment = self.rpc_client.commitment();
         let mark_empty_if_not_found =
             mark_empty_if_not_found.unwrap_or(&[]).to_vec();
         let program_ids = program_ids.map(|ids| ids.to_vec());
+        let method = if pubkeys.len() == 1 {
```
thlorenz: I do not understand why we need this here.
We already have a account fetch origin which states which method triggered the fetch and much more reliably.
We should just make sure that that is included in the logs.

magicblock-chainlink/src/remote_account_provider/mod.rs
```diff
@@ -997,15 +977,34 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
...
-                    Some(fetch_start_slot),
-                )
-                .await?;
+            remote_accounts = match self
+                .fetch_multi_rpc_only(pubkeys, fetch_start_slot, fetch_origin)
+                .await
+            {
+                Ok(remote_accounts) => remote_accounts,
+                Err(err) => {
+                    retries += 1;
```
thlorenz: This retry handling looks very similar to the slot-mismatch
retry path below, but the two branches already differ
subtly: this one sleeps with `max(config.retry_interval_ms,
RPC_FETCH_RETRY_DELAY)`, while the mismatch path sleeps only
`config.retry_interval_ms`. 

We should extract the retry
bookkeeping / limit checks / delay calculation into a shared
helper so both failure modes use the same policy and future
changes don't have to be kept in sync manually.

thlorenz (changes requested):
LGTM, but ideally we'd remove the duplication before merging and address the other comment as well.

magicblock-chainlink/src/remote_account_provider/mod.rs
GabrielePicco: Agreed, removed

magicblock-chainlink/src/remote_account_provider/mod.rs
GabrielePicco: Good point, done here: https://github.com/magicblock-labs/magicblock-validator/pull/1242/changes/74404d9d5046576cae15c5401515ee7fc9025a38

thlorenz (approved):
Thanks for removing the duplication. Good to merge now after addressing the AI comment

PR #1243, opened 2026-05-25
Fix pubsub reconnect failure handling

GabrielePicco:
## Summary
- make pubsub reconnect return `Result` so WARN logs include the concrete reconnect or resubscribe failure
- track connected SubMux clients and send new subscribe/unsubscribe operations only to connected clients while a failed client is reconnecting
- add regression coverage for skipping a disconnected client during reconnect

### DIFF
--- a/magicblock-chainlink/src/submux/mod.rs
+++ b/magicblock-chainlink/src/submux/mod.rs
@@ -323,9 +323,9 @@ where
 
                     // Update connection related metrics
                     let was_connected = {
-                        let mut connected_ids = connected_client_ids
-                            .lock()
-                            .expect("connected_client_ids lock poisoned");
+                        let mut connected_ids = Self::connected_client_ids_lock(
+                            &connected_client_ids,
+                        );
                         connected_ids.remove(&Self::client_key(&client))
                     };
                     if was_connected {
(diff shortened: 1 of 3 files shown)


magicblock-chainlink/src/submux/mod.rs
```diff
@@ -347,11 +369,40 @@ where
         self.clients_lock().clone()
     }
 
+    fn client_key(client: &Arc<T>) -> usize {
```
thlorenz, line 373: Why do we need another id for the clients? Each client [already has an id](https://github.com/magicblock-labs/magicblock-validator/blob/57d900b54f35290e47d8e2791d2ebeb7ac05563b/magicblock-chainlink/src/remote_account_provider/chain_pubsub_client.rs#L63) derived from the provider, etc.
It's much more descriptive than the ptr derived one.

If you are worried about uniqueness we can update the id derivation implementation to make sure we don't get duplicates.

thlorenz (changes requested):
LGTM aside from how the id is derived.

magicblock-chainlink/src/submux/mod.rs
GabrielePicco: Resolved in https://github.com/magicblock-labs/magicblock-validator/pull/1243/changes/4b4b7386154ea3f9b3e39974a9ab97739b6167f4

PR #1249, opened 2026-05-26
Fix: token2022 support

GabrielePicco:
## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added support for Token-2022 (SPL Token Extensions) program alongside standard SPL Token, enabling operations with Token-2022 accounts.
  * Enhanced ATA validation and projection logic with improved delegated account handling.

* **Tests**
  * Expanded test coverage for Token-2022 ATA operations, delegation scenarios, and account projection functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1249?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
+++ b/magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
@@ -4,8 +4,9 @@ use dlp_api::state::DelegationRecord;
 use futures_util::future::join_all;
 use magicblock_accounts_db::traits::AccountsBank;
 use magicblock_core::token_programs::{
-    is_ata, try_derive_ata_address_and_bump, try_derive_eata_address_and_bump,
-    AtaInfo,
+    is_ata, try_derive_ata_address_and_bump_with_token_program,
+    try_derive_eata_address_and_bump, AtaInfo, TOKEN_2022_PROGRAM_ID,
+    TOKEN_PROGRAM_ID,
 };
 use magicblock_metrics::metrics;
 use solana_account::{AccountSharedData, ReadableAccount};
@@ -58,16 +59,22 @@ fn ata_info_from_layout(
 
     let mint = Pubkey::new_from_array(data[0..32].try_into().ok()?);
     let wallet_owner = Pubkey::new_from_array(data[32..64].try_into().ok()?);
-    let (derived_ata, _) =
-        try_derive_ata_address_and_bump(&wallet_owner, &mint)?;
-    if derived_ata != *ata_pubkey {
-        return None;
+    for token_program in [TOKEN_PROGRAM_ID, TOKEN_2022_PROGRAM_ID] {
+        if try_derive_ata_address_and_bump_with_token_program(
+            &wallet_owner,
+            &mint,
+            &token_program,
+        )
+        .is_some_and(|(derived_ata, _)| derived_ata == *ata_pubkey)
+        {
+            return Some(AtaInfo {
+                mint,
+                owner: wallet_owner,
+            });
+        }
     }
 
-    Some(AtaInfo {
-        mint,
-        owner: wallet_owner,
-    })
+    None
 }
 
 /// Resolves ATAs with eATA projection.
--- a/magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
+++ b/magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
@@ -4,9 +4,8 @@ use dlp_api::state::DelegationRecord;
 use futures_util::future::join_all;
 use magicblock_accounts_db::traits::AccountsBank;
 use magicblock_core::token_programs::{
-    is_ata, try_derive_ata_address_and_bump_with_token_program,
-    try_derive_eata_address_and_bump, AtaInfo, TOKEN_2022_PROGRAM_ID,
-    TOKEN_PROGRAM_ID,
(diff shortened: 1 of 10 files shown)


magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
```diff
@@ -58,16 +59,22 @@ fn ata_info_from_layout(
 
     let mint = Pubkey::new_from_array(data[0..32].try_into().ok()?);
     let wallet_owner = Pubkey::new_from_array(data[32..64].try_into().ok()?);
-    let (derived_ata, _) =
-        try_derive_ata_address_and_bump(&wallet_owner, &mint)?;
-    if derived_ata != *ata_pubkey {
-        return None;
+    for token_program in [TOKEN_PROGRAM_ID, TOKEN_2022_PROGRAM_ID] {
```
thlorenz: Would be nice if that logic lives in the same crate/module as `try_derive_ata_address_and_bump_with_token_program`, i.e. `try_derive_ata_address_and_bump` which checks both programs.
Having this as part of chainlink which is already complex is not a good idea IMO.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -912,19 +913,29 @@ where
...
         )
         .and_then(|(wallet_owner, mint)| {
-            try_derive_ata_address_and_bump(&wallet_owner, &mint)
-                .map(|(ata_pubkey, _)| ata_pubkey)
-        });
-        let mut pubkeys_to_clone = vec![pubkey];
-        if let Some(ata_pubkey) = greedy_ata_pubkey {
-            pubkeys_to_clone.push(ata_pubkey);
-        }
+            [TOKEN_2022_PROGRAM_ID, TOKEN_PROGRAM_ID]
```
thlorenz: Here we basicallly are repeating the logic from above that I recommended above to add into a single method inside `magicblock_core::token_programs`.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -1344,26 +1357,42 @@ where
             eata_account.data(),
             deleg_record.owner,
         )?;
-        let (ata_pubkey, _) =
-            try_derive_ata_address_and_bump(&wallet_owner, &mint)?;
-
+        let legacy_ata_pubkey =
```
thlorenz: Again repeating this knowledge about two token programs.

The reusable method could return `(Option<Pubkey>, Option<Pubkey>)` for each token program and here you'd just do 

```rust
let (legacy_ata_pubkey, token_2022_ata_pubkey) = <reusable method>(..);
```

If performance is important then you could add a flag to the method `short_circuit: bool` which returns `None` for the second tuple if the first one is found. That'd be used for the calls higher above.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -1344,26 +1357,42 @@ where
...
+            )?
+            .0;
+        let token_2022_ata_pubkey =
+            try_derive_ata_address_and_bump_with_token_program(
+                &wallet_owner,
+                &mint,
+                &TOKEN_2022_PROGRAM_ID,
+            )?
+            .0;
+
```
thlorenz: All this code above and below is very complex and I'm thinking it should be factored out into a different module entirely instead of living right in the fetch_cloner.

Also it is almost impossible for me to verify these changes since I'm not familiar with the ATA token functionality as much as you are. It'd be nice to add some very detailed comments around how this works (ideally in that separate module).

thlorenz (approved):
LGTM regarding correctness, but the implementation has some repetition and adds eATA specific knowledge to the fetch cloner which isn't ideal.

I explained in the comments how we should improve this before merging.

magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
GabrielePicco: Resolved in f01c43c5: supported ATA derivation now lives in magicblock_core::token_programs and chainlink uses the shared helper.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
GabrielePicco: Resolved in f01c43c5: the greedy path now uses the shared magicblock_core::token_programs helper.

GabrielePicco:
> LGTM regarding correctness, but the implementation has some repetition and adds eATA specific knowledge to the fetch cloner which isn't ideal.
> 
> I explained in the comments how we should improve this before merging.

Thanks @thlorenz, applied the suggestions

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
GabrielePicco: Resolved in f01c43c5: eATA projection now uses the shared helper and preserves the Token-2022-first lookup.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
GabrielePicco: Resolved in f01c43c5: projection code moved into fetch_cloner/ata_projection.rs with comments on the in-bank ATA/layout invariant.

PR #1250, opened 2026-05-27
Reconcile lookup table state on extension errors

GabrielePicco:
## Summary

- Reconcile local lookup table pubkey tracking with the on-chain table before deciding whether an active table can be extended.
- Add remote-only lookup table entries locally with refcount 0 so existing reservations are preserved while capacity checks reflect chain state.

## Root cause

TableMania used its local refcounted pubkey set as the source of truth for lookup table capacity. If the on-chain ALT had been extended outside the current local state, TableMania could attempt to extend a table that was already full on base chain.

## Impact

When local ALT state is stale, a remotely full table is now treated as full before extension, allowing the reservation flow to create a new lookup table for remaining pubkeys.

## Bug Fixes

* Improved pubkey reservation reliability by automatically synchronizing lookup table state with on-chain records before allocating new entries, ensuring consistency and preventing conflicts across operations.
* Enhanced table capacity validation by checking on-chain state when extending existing tables and creating new ones, ensuring accurate tracking and preventing allocation errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1250?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/magicblock-table-mania/src/lookup_table_rc.rs
+++ b/magicblock-table-mania/src/lookup_table_rc.rs
@@ -70,6 +70,19 @@ impl RefcountedPubkeys {
         }
     }
 
+    /// Adds pubkeys already present on chain without creating local
+    /// reservations for them.
+    fn insert_unreserved_many(&mut self, pubkeys: &[Pubkey]) -> usize {
+        let mut inserted = 0;
+        for pubkey in pubkeys {
+            if !self.pubkeys.contains_key(pubkey) {
+                self.pubkeys.insert(*pubkey, AtomicUsize::new(0));
+                inserted += 1;
+            }
+        }
+        inserted
+    }
+
     /// Add a reservation to the pubkey if it is part of this table
     /// - *pubkey* to reserve
     /// - *returns* `true` if the pubkey could be reserved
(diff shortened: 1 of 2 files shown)


thlorenz (approved):
It's ok (after fixing issue the AI pointed out), but not great as this makes everything a lot slower. The original idea was to _not_ have to use the network to update a lookup table which we fully control and for which we should hold all info locally already.

However if this is needed to fix an ongoing issue quickly it's fine.
We should create an issue though to _properly_ address this in the future (without needing the reconciliation).

> TableMania could attempt to extend a table that was already full on base chain

That is exactly the point. The validator should be the only one updating the table, so there should be no way that it would be full already assuming that we are tracking everything correctly locally.
So the real issues that either there's an underlying bug or look up table keys are reused across validator restarts.

Fixing that would be the more correct way to address this.

GabrielePicco:
> It's ok (after fixing issue the AI pointed out), but not great as this makes everything a lot slower. The original idea was to _not_ have to use the network to update a lookup table which we fully control and for which we should hold all info locally already.
> 
> However if this is needed to fix an ongoing issue quickly it's fine. We should create an issue though to _properly_ address this in the future (without needing the reconciliation).
> 
> > TableMania could attempt to extend a table that was already full on base chain
> 
> That is exactly the point. The validator should be the only one updating the table, so there should be no way that it would be full already assuming that we are tracking everything correctly locally. So the real issues that either there's an underlying bug or look up table keys are reused across validator restarts.
> 
> Fixing that would be the more correct way to address this.

- I moved reconciliation only on errors (so this is irrelevant for the happy path)
- I realised the main issue was that the LUT was already closed (or the closing confirmation was lost) and the loop retried it forever since it's impossible to close it after it was already closed and the error was shallowed.

This is now clean and fix both issues

thlorenz (approved):
Nice, that looks like a better fix.
No perf overhead for the happy path anymore.

taco-paco (commented):
LGTM

PR #1251, opened 2026-05-27
Recover pending commit intents on startup

GabrielePicco:
## Summary

- Add a startup background recovery pass for persisted pending commit rows.
- Reconstruct account-only commit/finalize intent bundles from `commit_status` rows and schedule them without inserting duplicate persistence rows.
- Gate recovered intents to rows whose last retry is more than 30 minutes old and accounts that are still delegated on base and ER.
- Use a current local blockhash for recovered `ScheduledCommitSent` fallback signals.
- Simulate signed base transactions before sending so failed transactions do not spend validator lamports.

NOTE:

- The recovery path can change the semantics of persisted pending intent bundles by dropping base actions/callbacks while still committing accounts. This is a correctness issue for supported scheduled intent inputs after restart. This will be solved in https://github.com/magicblock-labs/magicblock-validator/discussions/1246

## Why

If a validator restarts after a base-chain commit/finalize attempt remains pending in persistence, the node currently does not resubmit that persisted work. This leaves already-scheduled account commits dependent on manual intervention.

## Notes

- The recovery path only reconstructs commit/finalize account intents represented by pending `commit_status` rows.
- Normal live scheduling still uses the existing persistence path.
- Recovered scheduling deliberately skips `start_base_intents` to avoid duplicate row insertion.
- Recovery only schedules an intent after age, delegation ownership, and simulation checks pass.

## Summary by CodeRabbit

* **New Features**
  * Added recovery mechanism for pending intents that failed to send initially.
  * Added account delegation status checking across multiple blockchain environments.

* **Bug Fixes**
  * Improved handling of disconnected clients in subscription management to prevent stale data processing.
  * Enhanced client connection tracking with automatic recovery from connection failures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1251?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


magicblock-committor-service/src/persist/commit_persister.rs
```diff
@@ -243,6 +248,15 @@ impl IntentPersister for IntentPersisterImpl {
             .get_commit_statuses_by_id(message_id)
     }
 
+    fn get_pending_commit_statuses(
+        &self,
+    ) -> CommitPersistResult<Vec<CommitStatusRow>> {
+        self.commits_db
+            .lock()
+            .map_err(|err| CommitPersistError::MutexPoisoned(err.to_string()))?
```
taco-paco:
```suggestion
          .expect(POISONED_MUTEX_MSG)
```

Atm we use expect() on locks that return error, that would be out of scope of this PR to add custom handling of poisoned mutexes

magicblock-accounts/src/scheduled_commits_processor.rs
```diff
@@ -344,7 +459,7 @@ struct ScheduledBaseIntentMeta {
     blockhash: Hash,
     payer: Pubkey,
     included_pubkeys: Vec<Pubkey>,
-    intent_sent_transaction: Transaction,
+    intent_sent_transaction: Option<Transaction>,
```
taco-paco, line 481: Not sure `Option` is applicable here, we should execute it anyway.  If on creation infarmation is abasent we should create tx

magicblock-committor-service/src/persist/commit_persister.rs
GabrielePicco: Fair, used expect

magicblock-accounts/src/scheduled_commits_processor.rs
GabrielePicco: The recovered path does still execute a `scheduled_commit_sent` tx. Option is used only because recovered intents are reconstructed from persisted rows and do not have the original sent_transaction/signature available. When it is None, we synthesize a fresh `scheduled_commit_sent` tx with the current blockhash before executing it.

magicblock-committor-service/src/committor_processor.rs
```diff
@@ -187,3 +242,210 @@ impl CommittorProcessor {
...
+        grouped_rows.entry(row.message_id).or_default().push(row);
+    }
+
+    grouped_rows
+        .into_iter()
+        .filter_map(|(message_id, rows)| {
+            if rows
+                .iter()
+                .any(|row| !pending_row_is_old_enough(row, recovery_time))
+            {
```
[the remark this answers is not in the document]
GabrielePicco: Fair comment, but this temporary recover is sufficient for now (considering that it wasn't there before), leaving the proper implementation to MIMD https://github.com/magicblock-labs/magicblock-validator/discussions/1246. We can run this process periodically if needed.

magicblock-accounts/src/scheduled_commits_processor.rs
```diff
@@ -196,9 +351,18 @@ impl ScheduledCommitsProcessorImpl {
...
         mut intent_meta: ScheduledBaseIntentMeta,
+        current_blockhash: Hash,
     ) {
-        let intent_sent_transaction =
-            std::mem::take(&mut intent_meta.intent_sent_transaction);
+        let intent_sent_transaction = intent_meta
+            .intent_sent_transaction
+            .take()
+            .unwrap_or_else(|| {
+                intent_meta.blockhash = current_blockhash;
```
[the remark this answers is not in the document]
GabrielePicco: This is intended

magicblock-committor-service/src/committor_processor.rs
```diff
@@ -187,3 +242,219 @@ impl CommittorProcessor {
...
+                    commit_finalize_and_undelegate_accounts.push(account);
+                } else {
+                    commit_finalize_accounts.push(account);
+                }
+            }
+
+            let mut intent_bundle = MagicIntentBundle::default();
+            if !commit_finalize_accounts.is_empty() {
+                intent_bundle.commit_finalize =
+                    Some(CommitType::Standalone(commit_finalize_accounts));
```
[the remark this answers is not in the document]
GabrielePicco: we currently always commit_finalize

magicblock-accounts/src/scheduled_commits_processor.rs
taco-paco: Why don't we synthesize it when we take data from db? We would avoid `Option` and extra spagetification assosiated with handling it

magicblock-accounts/src/scheduled_commits_processor.rs
```diff
@@ -118,11 +122,150 @@ impl ScheduledCommitsProcessorImpl {
...
+        if intent_bundles.is_empty() {
+            return;
+        }
+        let intent_bundles =
+            self.recoverable_intent_bundles(intent_bundles).await;
+        if intent_bundles.is_empty() {
+            return;
+        }
+
+        let pubkeys_being_undelegated = {
```
taco-paco, line 119: The code below here duplicates `ScheduledCommitsProcessorImpl::process` logic with some minor changes. Let's reuse `process`, if `process` requires modification, let's do it and reuse. For that intents can be flushed into `transaction_scheduler: TransactionScheduler` and then call `process` it will do the job duplicated here

magicblock-accounts/src/scheduled_commits_processor.rs
GabrielePicco: I would not synthesize during DB recovery, because that could create a stale tx before the base commit finishes. The option is clearer imo. Anyway, this code will be superseded by https://github.com/magicblock-labs/magicblock-validator/discussions/1246, so feel free to rework it there.

magicblock-accounts/src/scheduled_commits_processor.rs
GabrielePicco: Fair refactored: https://github.com/magicblock-labs/magicblock-validator/pull/1251/changes/ca36e632ead676b347822b7a35181a0ec0afa208

magicblock-accounts/src/scheduled_commits_processor.rs
```diff
@@ -118,11 +122,151 @@ impl ScheduledCommitsProcessorImpl {
...
+                error!(error = ?err, "Failed to receive pending commit intents");
+                return;
+            }
+        };
+        if intent_bundles.is_empty() {
+            return;
+        }
+        let intent_bundles =
+            self.recoverable_intent_bundles(intent_bundles).await;
+        if intent_bundles.is_empty() {
```
taco-paco, line 170: All the code below is replacable by existing functionality
```
        self.transaction_scheduler.accept_scheduled_base_intent(intent_bundles);
        self.process().await;
```

magicblock-accounts/src/scheduled_commits_processor.rs
```diff
@@ -118,11 +122,151 @@ impl ScheduledCommitsProcessorImpl {
...
+
+    #[instrument(skip(self))]
+    async fn recover_pending_intents(&self) {
+        let intent_bundles = match self
+            .committor
+            .get_pending_intent_bundles()
+            .await
+        {
+            Ok(Ok(intent_bundles)) => intent_bundles,
+            Ok(Err(err)) => {
```
taco-paco: the errors here could be propagatted in calling site and logged there, would reduce code here an below. I see `get_pending_intent_bundles` returns error which should be descriptive enough.

Thus the whole callsite could be shrank to 
```rust
let intent_bundles = match self
            .committor
            .get_pending_intent_bundles()
            .await??
```

magicblock-accounts/src/scheduled_commits_processor.rs
taco-paco: The process historically doesn't have the best api, but to avoid insertion there could be `process_inner()`

Thust the function could would be
```rust
let intents = self.load().await??;
self.process_inner(intents).await?;
```

taco-paco (approved):
Thanks for taking this on! There's couple of things that would be good to do: reduce code repetition, simplify error dispatch and logging via propagation to lowest point and via custom Error variants if needed.

There are also potential race conditions between detached `recover_pending_intents` and ongoing requests, but we can go ahead without addressing this for now.

magicblock-accounts/src/scheduled_commits_processor.rs
GabrielePicco: The normal path re-persists intents via `start_base_intents` before scheduling them. Recovered intents were loaded from existing pending DB rows, and the current code deliberately calls `magicblock-accounts/src/scheduled_commits_processor.rs:191`, whose committor path schedules without re-persisting.

magicblock-accounts/src/scheduled_commits_processor.rs
GabrielePicco: Fair, refactored

PR #1259, opened 2026-05-28
Reduce RPC hot-path requests

GabrielePicco:
## Summary
- Cache latest blockhashes and recent slots in the shared `MagicblockRpcClient` to avoid repeated hot-path RPC calls.
- Seed the slot cache from `getLatestBlockhash` responses and serialize cache refreshes to avoid request stampedes.
- Replace the TableMania ALT finalized-readiness slot polling loop with timed sleeps while keeping finalized account fetches as the readiness check.

## Validation
- `cargo check -p magicblock-rpc-client -p magicblock-table-mania`
- `cargo check -p magicblock-committor-service`
- `cargo test -p magicblock-rpc-client -p magicblock-table-mania`

## Summary by CodeRabbit

* **Performance Improvements**
  * Added in-memory caching for RPC blockhashes and slots to reduce requests and improve latency.
  * Optimized remote-table polling with timed, fixed-interval checks for smoother synchronization.

* **Reliability**
  * New explicit cache invalidation ensures retries use fresh blockhashes, reducing transient failures.
  * Services now accept an optional shared chain-slot signal to improve coordination across components.

* **Logging**
  * Reduced noisy wait-loop logs by emitting periodic status messages less frequently.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1259?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/magicblock-rpc-client/Cargo.toml
+++ b/magicblock-rpc-client/Cargo.toml
@@ -29,6 +29,7 @@ solana-address-lookup-table-interface = { workspace = true, features = [
   "bytemuck",
 ] }
 futures-util = { workspace = true }
+serde_json = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
 
--- a/magicblock-rpc-client/src/lib.rs
+++ b/magicblock-rpc-client/src/lib.rs
@@ -3,7 +3,10 @@
 pub mod utils;
 
 use std::{
-    sync::Arc,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
     time::{Duration, Instant},
 };
 
@@ -35,10 +38,7 @@ use solana_transaction_error::{TransactionError, TransactionResult};
 use solana_transaction_status_client_types::{
     EncodedConfirmedTransactionWithStatusMeta, UiTransactionEncoding,
 };
-use tokio::{
-    sync::{Mutex as TMutex, RwLock},
-    time::sleep,
-};
+use tokio::{sync::Mutex as TMutex, time::sleep};
 use tracing::*;
 
 /// The encoding to use when sending transactions
@@ -241,15 +241,15 @@ const SLOT_CACHE_TTL: Duration = Duration::from_millis(400);
 
 #[derive(Default)]
 struct RpcClientCache {
-    blockhash: RwLock<Option<CachedBlockhash>>,
-    blockhash_refresh: TMutex<()>,
-    slot: RwLock<Option<CachedSlot>>,
-    slot_refresh: TMutex<()>,
+    blockhash: TMutex<Option<CachedBlockhash>>,
+    slot: TMutex<Option<CachedSlot>>,
 }
 
 #[derive(Clone, Copy)]
 struct CachedBlockhash {
     blockhash: Hash,
+    context_slot: Slot,
+    last_valid_block_height: u64,
     fetched_at: Instant,
 }
 
@@ -265,6 +265,7 @@ struct CachedSlot {
 pub struct MagicblockRpcClient {
     client: Arc<RpcClient>,
     cache: Arc<RpcClientCache>,
+    chain_slot: Option<Arc<AtomicU64>>,
 }
 
 impl From<RpcClient> for MagicblockRpcClient {
@@ -279,21 +280,47 @@ impl MagicblockRpcClient {
         Self {
             client,
             cache: Arc::new(RpcClientCache::default()),
+            chain_slot: None,
+        }
+    }
+
+    pub fn new_with_chain_slot(
+        client: Arc<RpcClient>,
+        chain_slot: Arc<AtomicU64>,
+    ) -> Self {
+        Self {
+            client,
+            cache: Arc::new(RpcClientCache::default()),
+            chain_slot: Some(chain_slot),
         }
     }
 
     pub async fn get_latest_blockhash(
         &self,
     ) -> MagicBlockRpcClientResult<Hash> {
-        if let Some(blockhash) = self.cached_blockhash().await {
+        let mut cached = self.cache.blockhash.lock().await;
+        if let Some(blockhash) = Self::fresh_cached_blockhash(&cached) {
             return Ok(blockhash);
         }
 
-        let _guard = self.cache.blockhash_refresh.lock().await;
-        if let Some(blockhash) = self.cached_blockhash().await {
-            return Ok(blockhash);
-        }
+        let (blockhash, context_slot, last_valid_block_height) =
+            self.fetch_latest_blockhash_with_context().await?;
+        *cached = Some(CachedBlockhash {
+            blockhash,
+            context_slot,
+            last_valid_block_height,
+            fetched_at: Instant::now(),
+        });
+        drop(cached);
 
+        self.record_observed_slot(context_slot).await;
+
+        Ok(blockhash)
+    }
+
+    async fn fetch_latest_blockhash_with_context(
+        &self,
+    ) -> MagicBlockRpcClientResult<(Hash, Slot, u64)> {
         let resp: Response<RpcBlockhash> = self
             .client
             .send(RpcRequest::GetLatestBlockhash, json!([self.commitment()]))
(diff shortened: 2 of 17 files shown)


magicblock-rpc-client/src/lib.rs
```diff
@@ -247,24 +276,114 @@ impl From<RpcClient> for MagicblockRpcClient {
...
     ) -> MagicBlockRpcClientResult<Hash> {
-        self.client.get_latest_blockhash().await.map_err(|e| {
-            MagicBlockRpcClientError::GetLatestBlockhash(Box::new(e))
-        })
+        if let Some(blockhash) = self.cached_blockhash().await {
+            return Ok(blockhash);
+        }
+
+        let _guard = self.cache.blockhash_refresh.lock().await;
+        if let Some(blockhash) = self.cached_blockhash().await {
```
thlorenz: This seems to repeat the check from above except above we don't guard the access.
Also needing to take a _guard_ of one resource to access another resource is not good practice. Instead if both resources are that dependent then maybe they should be kept inside a single wrapping resource which will be guarded to access the inner pieces.

magicblock-rpc-client/src/lib.rs
```diff
@@ -247,24 +276,114 @@ impl From<RpcClient> for MagicblockRpcClient {
...
+            return Ok(blockhash);
+        }
+
+        let resp: Response<RpcBlockhash> = self
```
thlorenz, line 324: Why not keep `self.client.get_latest_blockhash()` here instead of manually repeating what's in that method?

magicblock-rpc-client/src/lib.rs
```diff
@@ -247,24 +276,114 @@ impl From<RpcClient> for MagicblockRpcClient {
...
+                RpcClientError::new_with_request(
+                    RpcClientErrorKind::RpcError(RpcError::ParseError(
+                        "Hash".to_string(),
+                    )),
+                    RpcRequest::GetLatestBlockhash,
+                ),
+            ))
+        })?;
+
+        self.cache_slot(resp.context.slot).await;
```
thlorenz: Seems like slot and blockhash are very coupled here, maybe keep them together in a wrapper as well?

magicblock-rpc-client/src/lib.rs
```diff
@@ -247,22 +276,102 @@ impl From<RpcClient> for MagicblockRpcClient {
...
+            });
+        }
+
+        Ok(blockhash)
     }
 
     pub async fn get_slot(&self) -> MagicBlockRpcClientResult<Slot> {
-        self.client
+        if let Some(slot) = self.cached_slot().await {
+            return Ok(slot);
```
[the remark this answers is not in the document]
thlorenz: How was this resolved? Which commit?

magicblock-rpc-client/src/lib.rs
```diff
@@ -247,24 +276,114 @@ impl From<RpcClient> for MagicblockRpcClient {
...
+        self.cache_slot(slot).await;
+        Ok(slot)
+    }
+
+    async fn get_cached_slot(&self) -> MagicBlockRpcClientResult<Slot> {
+        if let Some(slot) = self.cached_slot().await {
+            return Ok(slot);
+        }
+
+        let _guard = self.cache.slot_refresh.lock().await;
```
thlorenz: ditto separate _guard_ (cache) for other resource (cached_slot) which seems like a code smell.

magicblock-rpc-client/src/lib.rs
```diff
@@ -247,24 +276,114 @@ impl From<RpcClient> for MagicblockRpcClient {
...
+    async fn fetch_slot(&self) -> MagicBlockRpcClientResult<Slot> {
         self.client
             .get_slot()
             .await
             .map_err(|e| MagicBlockRpcClientError::GetSlot(Box::new(e)))
     }
 
+    pub async fn clear_cached_blockhash(&self) {
+        let _guard = self.cache.blockhash_refresh.lock().await;
+        let mut cached = self.cache.blockhash.write().await;
```
thlorenz: ditto separate guard

magicblock-rpc-client/Cargo.toml
```diff
@@ -29,6 +29,7 @@ solana-address-lookup-table-interface = { workspace = true, features = [
   "bytemuck",
 ] }
 futures-util = { workspace = true }
+serde_json = { workspace = true }
```
thlorenz, line 32: Afaik this is only needed since we reimplemented code that's already in the solana rpc client crate.

magicblock-table-mania/src/manager.rs
```diff
@@ -648,12 +647,10 @@ impl TableMania {
                     );
                 }
 
-                if let Ok(slot) = self.rpc_client.wait_for_next_slot().await {
-                    if slot - last_slot > 20 {
-                        tracing::Span::current().record("current_slot", slot);
-                        debug!("Still waiting for remote tables");
-                    }
-                    last_slot = slot;
+                sleep(Duration::from_millis(400)).await;
```
thlorenz: If I understand correctly this is to remove polling the RPC over and over for the next slot?
So here we just _assume_ that after 400ms the next slot was reached?

In that case we should take slot jitter into account and wait for at least 450ms (slots don't update exactly after 400ms)

thlorenz: Another option would be to wire the chain slot subscription that chainlink already has to avoid polling and still be able to continue _exactly_ when the next slot was reached.

To me that seems the more correct solution, but may require a lot more changes.

thlorenz (changes requested):
Overall looks good, but some things seem to be unclean and we should evaluate if they can be improved.

magicblock-rpc-client/src/lib.rs
```diff
@@ -247,24 +276,114 @@ impl From<RpcClient> for MagicblockRpcClient {
...
+    }
+
+    async fn fetch_slot(&self) -> MagicBlockRpcClientResult<Slot> {
         self.client
             .get_slot()
             .await
             .map_err(|e| MagicBlockRpcClientError::GetSlot(Box::new(e)))
     }
 
+    pub async fn clear_cached_blockhash(&self) {
```
taco-paco: Why does one have to call this manually? I would imagine simple flow like this
get_latest_blockhash()

1. Check cache
   - Missing -> fetch, store, return
   - Present -> continue
2. Is cached value still fresh?
   - Yes -> return cached value
   - No -> fetch new value, overwrite cache, return

`clear_cached_blockhash` seems like an extra piece which is also error prone as one could forget to trigger that.

magicblock-rpc-client/src/lib.rs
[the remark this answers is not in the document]
GabrielePicco: Fixed in 2151ff90699d343a29e58bc2d7bc789326a2a56e. `get_slot()` now bypasses the cache and fetches from RPC every time, then updates the cache opportunistically.

magicblock-rpc-client/src/lib.rs
GabrielePicco: The second check is intentional double-checked locking: another task may have refreshed the cache while this task was waiting on the refresh mutex. Agree that looks unclean, will refactor

magicblock-rpc-client/src/lib.rs
GabrielePicco: The normal freshness path is already TTL-based. This explicit invalidation is for immediate retry paths after cleanup/patching, where we do not want to sign the next prepared transaction with the same cached hash even if it is still TTL-fresh. Will clarify the names

magicblock-rpc-client/src/lib.rs
GabrielePicco: Solana's Rust RPC client only exposes the parsed blockhash from `get_latest_blockhash()`, while the RPC response also includes `context.slot`. This is dumb, but afaik there is not way to get both without manually implementing it

magicblock-rpc-client/Cargo.toml
GabrielePicco: See this: https://github.com/magicblock-labs/magicblock-validator/pull/1259#discussion_r3322600255

Agree on the ugliness, but afaik there is not way to cleanly do it and I want to avoid 2 RPC calls when the slot information is already in the response

magicblock-rpc-client/src/lib.rs
GabrielePicco: Resolved in d9be235e: cached blockhash now carries the context slot and records it into the shared observed slot.

magicblock-rpc-client/src/lib.rs
GabrielePicco: Resolved in d9be235e: slot cache now uses a single mutex, so the separate refresh guard is gone.

magicblock-rpc-client/src/lib.rs
GabrielePicco: Resolved in d9be235e: same cleanup, no separate refresh guard remains here.

magicblock-table-mania/src/manager.rs
GabrielePicco: Resolved in d9be235e: TableMania now waits on the shared Chainlink-observed slot (added an AtomicU64 to wire chain slot subscription from chainlink), with a 450ms fallback.

magicblock-rpc-client/src/lib.rs
```diff
@@ -394,7 +563,12 @@ impl MagicblockRpcClient {
         slot: Slot,
     ) -> MagicBlockRpcClientResult<Slot> {
         let higher_slot = loop {
-            let next_slot = self.get_slot().await?;
+            let next_slot = if let Some(next_slot) = self.observed_chain_slot()
+            {
+                next_slot
+            } else {
+                self.get_cached_slot().await?
```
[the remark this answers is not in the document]
GabrielePicco: `chain_slot` is always fresh, if the pubsub slot feed is not active, we have way bigger consistency problems

thlorenz (approved):
LGTM now. Thanks for addressing all points I raised.

PR #1263, opened 2026-05-29
refactor: committor service

taco-paco:
refactor(committor-service): simplify code, remove unused and dead code

## Summary

Simplifies `magicblock-committor-service` by removing the Actor pattern (`CommittorService` / `CommittorActor` / `CommittorMessage` / `BaseIntentCommittor` / `CommittorServiceExt`) and replacing it with direct use of `CommittorProcessor` and a new `IntentExecutionService`, aligning the committor service with the same service model already used by task-scheduler and actions_callback. No business logic was changed.

## Breaking Changes
- [x] None. Not really :)

## Test Plan

- Existing integration tests in `test-integration/test-committor-service` updated to use the new `CommittorProcessor` API directly and pass unchanged.
- No new behaviour introduced; this is pure code removal and structural simplification.

## Summary by CodeRabbit

* **Refactor**
  * Shifted to an intent-execution service + processor architecture, simplifying startup/shutdown and removing legacy scheduled-commit and slot-ticker flows; reduced compute-budget and error surface.

* **New Features**
  * Synchronous per-bundle execution API with reliable request/response delivery and result dispatching.
  * Intent execution now notifies downstream clients when commits are sent.

* **Tests**
  * Integration tests updated to exercise the processor-driven execution flow.


magicblock-committor-service/src/service.rs
```diff
@@ -1,570 +1,431 @@
...
-            respond_to: tx,
-            commit_id,
-            pubkey,
-        });
-        rx
-    }
+        // Add metas for intent we schedule
+        let pubkeys_being_undelegated = {
+            let mut intent_metas =
+                self.intents_meta_map.lock().expect(POISONED_MUTEX_MSG);
```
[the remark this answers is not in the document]
taco-paco: Out of scope

magicblock-committor-service/src/service/intent_client.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    async fn accept_scheduled_intents(
+        &self,
+    ) -> Result<Vec<ScheduledIntentBundle>, Self::Error> {
+        // If accounts were scheduled to be committed, we accept them here
+        // and processs the commits
+        let magic_context_acc =
+            self.accounts_db.get_account(&MAGIC_CONTEXT_PUBKEY).expect(
+                "Validator found to be running without MagicContext account!",
+            );
```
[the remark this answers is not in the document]
taco-paco: Have to crash in that case

magicblock-api/src/magic_validator.rs
```diff
@@ -419,10 +421,8 @@ impl MagicValidator {
             exit,
             _metrics: (metrics_service, system_metrics_ticker),
             // NOTE: set during [Self::start]
-            slot_ticker: None,
-            committor_service,
+            intent_execution_service,
```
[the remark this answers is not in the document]
taco-paco: Fixed

magicblock-committor-service/src/service.rs
```diff
@@ -1,570 +1,431 @@
...
+
+            pubkeys_being_undelegated.into_iter().collect::<Vec<_>>()
+        };
+
+        self.process_undelegation_requests(pubkeys_being_undelegated)
+            .await;
+        self.processor
+            .schedule_intent_bundles(intent_bundles)
+            .await?;
+        Ok(())
```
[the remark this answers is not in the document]
taco-paco: Fixed

magicblock-committor-service/src/committor_processor.rs
```diff
@@ -169,6 +153,45 @@ impl CommittorProcessor {
...
+    pub async fn execute_intent_bundles(
+        &self,
+        intent_bundles: Vec<ScheduledIntentBundle>,
+    ) -> CommittorServiceResult<Vec<BroadcastedIntentExecutionResult>> {
+        // Critical section
+        let receivers = {
+            let mut result_listeners = self
+                .pending_result_listeners
+                .lock()
+                .expect(POISONED_MUTEX_MSG);
```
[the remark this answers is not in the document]
taco-paco: Out of scope

magicblock-committor-service/src/committor_processor.rs
```diff
@@ -169,6 +153,45 @@ impl CommittorProcessor {
...
+        };
+
+        self.schedule_intent_bundles(intent_bundles).await?;
+        let results = join_all(receivers.into_iter())
+            .await
+            .into_iter()
+            .collect::<Result<Vec<_>, RecvError>>()?;
+
+        Ok(results)
+    }
```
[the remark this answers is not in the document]
taco-paco: Fixed

magicblock-committor-service/src/committor_processor.rs
```diff
@@ -229,6 +243,62 @@ impl CommittorProcessor {
         Ok(())
     }
 
+    pub async fn execute_intent_bundles(
```
snawaz, line 246: Why is this `execute_intent_bundles` added? 

I think we can remove it as nobody seems to call it (apart from one test)? If so, we can also remove the field `pending_result_listeners`, and the function `dispatcher()` as they work together with `execute_intent_bundles`.

magicblock-api/src/magic_sys_adapter.rs
```diff
@@ -19,13 +23,43 @@ impl MagicSysAdapter {
...
+                .fetch_current_commit_nonces(&pubkeys, min_context_slot)
+                .await;
+            if let Err(err) = sender.send(result) {
+                error!(error = ?err, "Failed to send result back");
+            }
+        });
+        drop(_guard);
+
+        receiver
     }
```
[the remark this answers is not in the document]
snawaz: I kinda agree with rabbit here. why not use the simpler API?

magicblock-committor-service/src/persist/commit_persister.rs
```diff
@@ -465,6 +481,126 @@ impl<T: IntentPersister> IntentPersister for Option<T> {
...
+    }
+    if intent_bundle.is_empty() {
+        return None;
+    }
+
+    Some(ScheduledIntentBundle {
+        id: message_id,
+        slot,
+        blockhash,
+        sent_transaction: Transaction::default(),
```
snawaz, line 570: semantic-change-1 (1): 

`sent_transaction: Transaction::default()` existed earlier as well. So this part is same. However, it is handled differently in the rest of the new code.. and that is where it starts differing.

magicblock-committor-service/src/service.rs
```diff
@@ -1,630 +1,529 @@
...
-    fn stopped(&self) -> WaitForCancellationFutureOwned {
-        self.cancel_token.clone().cancelled_owned()
+impl ScheduledBaseIntentMeta {
+    fn new(intent: &ScheduledIntentBundle) -> Self {
+        Self {
+            slot: intent.slot,
+            blockhash: intent.blockhash,
+            payer: intent.payer,
+            included_pubkeys: intent.get_all_committed_pubkeys(),
+            intent_sent_transaction: intent.sent_transaction.clone(),
```
snawaz: semantic-change-1 (2): 

`sent_transaction` is NOT persisted in the DB, so in the "recovered" intents scenario, `intent.sent_transaction.clone()` will produce an empty transaction. 

In the older (now-deleted) code, tt is handled correctly where `intent_sent_transaction` is declared to be `Option` and here it is initialized conditionally (please compare this with the deleted code).

magicblock-committor-service/src/service.rs
```diff
@@ -1,630 +1,529 @@
...
+        let sent_transaction =
+            mem::take(&mut intent_meta.intent_sent_transaction);
+        let sent_commit = ServiceInner::<R>::build_sent_commit(
+            intent_id,
+            intent_meta,
+            execution_result,
+        );
+        intent_client
+            .notify_commit_sent(sent_transaction, sent_commit)
+            .await?;
```
snawaz, line 367: semantic-change-1 (3):

- Here `mem::take()` is interesting. When `intent_sent_transaction` is a real tx it returns that tx and leaves an empty/default tx in the metadata. When `intent_sent_transaction` is already an empty/default tx (which is possible now for recovered intents) `take()` returns that empty/default tx. It is NOT the root cause; I am merely noting that this path will forward an empty tx for recovered intents.

- **BUG**: the old code synthesized a fresh `scheduled_commit_sent_tx` for recovered intents... but the new code does not synthesize it; it sends `Transaction::default()` instead (see `semantic-change-1 (1)`).


### Possible Fix

We can use the old approach, i.e use `Option` to explicitly handle the recovered path, but that still makes the code less-readable (I took hours to figure that out). So I think it's better to add something like this (the name could be better):

```rust
enum CommitSentNotificationTx { // or maybe ScheduledCommitSentTx 
   Live(Transaction)
   Recovered
}
```

We need a test to ensure recovered scenarios work correctly.

magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
```diff
@@ -308,24 +307,11 @@ where
...
-                if let Err(err) = executor.cleanup().await {
-                    error!(error = ?err, "Failed to cleanup after intent");
-                }
-            });
-        }
+        tokio::spawn(async move {
+            if let Err(err) = executor.cleanup().await {
+                error!(error = ?err, "Failed to cleanup after intent");
+            }
+        });
```
snawaz, line 314: semantic-change-2:

It is another semantic change compared to the master. We seem to (partially) reverse the behavior we ensured in #1291. 

Or maybe... we do the same thing but now with the help of `close_buffers` flag?

Anyway, it may be intentional semantic change if it is at all. Please clarify the intended behavior.

snawaz (changes requested):
It's an awesome PR. reducing 2000+ lines while simplifying the service model is not an easy feat!!

I posted two major comments tagged `semantic-change-1` and `semantic-change-2`.

- `semantic-change-1` looks like a real recovered-intent behavior regression.. and if so, it **likely needs code + test changes**.

- `semantic-change-2` may only need clarification, assuming the partial cleanup on failure is _intentional_ and _safe_.

magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
taco-paco: Yes, now executor has a flag and correctly cleans up everything.

magicblock-committor-service/src/committor_processor.rs
taco-paco: This is migration of old CommittorServiceExt logic. I think removing it is out of scope here. It allows to await in place scheduled intent results.

magicblock-committor-service/src/service.rs
taco-paco: Thanks, missed this during refactoring.
Fixed in a185476bb

magicblock-committor-service/src/service.rs
taco-paco: Fixed in a185476bb

magicblock-committor-service/src/persist/commit_persister.rs
taco-paco: Fixed in a185476bb

magicblock-api/src/magic_sys_adapter.rs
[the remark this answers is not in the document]
taco-paco: For some reason i though it wasn't allowed in this context.
Fixed in https://github.com/magicblock-labs/magicblock-validator/commit/a185476bb76c3c273712ec6fc74928e3e43c52d2

snawaz (approved):
Thanks for fixing the regression!

Looks awesome now.

Let's merge! 🚀

PR #1264, opened 2026-05-29
fix: drop stale subscription updates

thlorenz:
## Summary

Prevent stale forwarded subscription updates from recreating local bank state for accounts that are no longer watched by the remote account provider.

### magicblock-chainlink

- Drop subscription updates in the fetch cloner when the provider no longer watches the account, while preserving the existing delegated-account discovery path.
- If a stale update arrives for an account that is still present locally, route cleanup through the existing account-removal listener instead of evicting directly from the subscription-update task.
- Serialize the final watched-account check and eviction submission for the same pubkey so a concurrent re-subscription cannot race with stale-account cleanup.
- Keep unsubscribe behavior and the SubMux forwarder hot path unchanged.
- Add regression coverage for stale updates, removal enqueueing, watched-again skip behavior, and same-pubkey eviction/subscription serialization.

## Summary by CodeRabbit

* **New Features**
  * Subscription updates now carry an explicit source (account vs program), improving update routing and handling.

* **Bug Fixes**
  * Stop processing stale forwarded subscription updates for unwatched accounts. If the account still exists, a best-effort removal notification is enqueued (failures logged) and eviction is gated to avoid races with re-watching.

* **Tests**
  * Added tests and helpers validating forwarded-update handling, removal-notification enqueueing, guarded eviction with per-key locking, and re-watch race scenarios.


bzawisto (approved):
LGTM

PR #1268, opened 2026-06-01
fix: retry failed program subscriptions

thlorenz:
## Summary

Fix program subscription bookkeeping so transient subscription failures remain retryable instead of being recorded as successful locally.

### magicblock-chainlink

- Record SubMux program subscriptions only after the underlying fan-out subscription succeeds.
- Keep the Laser program stream manager's previous subscribed-program set when an incremental stream update fails, allowing the same program subscription to be retried.
- Add regression coverage for retrying failed program subscriptions in both SubMux and Laser stream-manager paths.
- Extend existing test mocks with targeted failure hooks needed to validate retry behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected subscription state rollback so failed subscription updates no longer leave stale program IDs.

* **Tests**
  * Added tests verifying retries and that subsequent attempts write updated subscription requests.
  * Enhanced test mocks to simulate and observe forced failures and count subscription/write attempts.


thlorenz:
@coderabbitai review

thlorenz:
@coderabbitai resume

magicblock-chainlink/src/submux/mod.rs
```diff
@@ -1057,21 +1057,25 @@ where
...
         )
         .process(self.connected_clients_snapshot())
-        .await
+        .await?;
+
+        self.program_subs
+            .lock()
+            .expect("program_subs lock poisoned")
+            .insert(program_id);
+        Ok(())
```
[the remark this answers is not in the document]
thlorenz: Stop pointing out expect calls on locks. We use those all over the codebase. It's ok.

magicblock-chainlink/src/submux/mod.rs
```diff
@@ -1057,21 +1057,25 @@ where
...
 
         AccountSubscriptionTask::SubscribeProgram(
             program_id,
             self.required_program_subscription_confirmations(),
         )
         .process(self.connected_clients_snapshot())
-        .await
+        .await?;
+
+        self.program_subs
```
GabrielePicco: `program_subs` is not updated until after the fanout await, a client that reconnects during this await can snapshot program_subs before the insertion while it was also excluded from the original `connected_clients_snapshot()`. 
That client is then marked connected without this program subscription and will not be resubscribed until another reconnect, so program update coverage can silently drop during concurrent subscribe/reconnect races; consider tentatively recording the desired program before awaiting and rolling it back on error.

GabrielePicco (commented):
LGTM in general, but I believe this introduce a reconnect race that can leave a reconnected client missing the program subscription. Should be fixed before merging

magicblock-chainlink/src/submux/mod.rs
thlorenz: Fixed in [fd10fed](https://github.com/magicblock-labs/magicblock-validator/pull/1268/commits/fd10fed74e7bf9128765e7d79ed60d11a0dccb1d)

GabrielePicco (approved):
LGTM!

PR #1269, opened 2026-06-01
Allow same-slot delegated clone override

GabrielePicco:
## Summary

- Allow a same-slot delegated clone to override an existing not delegated account clone.
- Keep duplicate delegated-account clone protection when the destination is already delegated.
- Add clone-account regression coverage for both the allowed plain-to-delegated transition and the rejected duplicate delegated update.

 ## Note on Scope

  This PR fixes the observed `InvalidWritableAccount` regression by allowing a same-slot delegated clone to override an existing plain ATA clone. In the failing flow, Chainlink first cloned the raw ATA as `delegated=false`, then later synthesized
  the delegated ATA from the companion eATA/delegation state at the same `remote_slot`. The previous clone guard rejected that projection as a duplicate.

  This is a pragmatic hotfix, not the ideal long-term boundary. The underlying issue is projection provenance: eATA-derived ATA state is currently emitted through the generic `CloneAccount` path, which makes a locally synthesized projection look
  like raw RPC account state. A follow-up should keep `CloneAccount` semantics strict and add a projection-specific materialization/update path, or an explicit projection source/version marker, for same-slot eATA projections.

### DIFF
--- a/programs/magicblock/src/clone_account/common.rs
+++ b/programs/magicblock/src/clone_account/common.rs
@@ -237,7 +237,9 @@ pub fn set_account_from_fields(
     // In the same slot, we do not expect an account to be delegated,
     // undelegated and then delegated. So if we see the same account being
     // delegated twice (ore more) in the same slot, we consider such cases
-    // as "duplicates".
+    // as "duplicates". A delegated clone is still allowed to override a
+    // same-slot plain clone, which can happen when an eATA projection catches
+    // up after the underlying ATA was cloned without delegation metadata.
     //
     // Under current design, we clone with "CONFIRMED" commitment and thus:
     //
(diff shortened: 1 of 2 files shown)


programs/magicblock/src/clone_account/common.rs
```diff
@@ -247,7 +249,10 @@ pub fn set_account_from_fields(
     //
     // Given this, same slot re-delegation is impossible therefore this this totally acceptable.
     //
-    if fields.delegated && acc.remote_slot() == fields.remote_slot {
+    if fields.delegated
+        && acc.delegated()
```
snawaz, line 253: ah, I see. This is interesting. 🤔

thlorenz (approved):
This LGTM in general, however I'm not sure regarding the limitation.
Wouldn't this miss undeletations of an account in the same slot on purposeand lead to the validator thinking it can write to the account even though it has been undelegated?

So basically, we favor same slot updates that lead to the account being delegated but are ok with possibly missing updates for the opposite?

GabrielePicco:
> This LGTM in general, however I'm not sure regarding the limitation. Wouldn't this miss undeletations of an account in the same slot on purposeand lead to the validator thinking it can write to the account even though it has been undelegated?
> 
> So basically, we favor same slot updates that lead to the account being delegated but are ok with possibly missing updates for the opposite?

This is the minimal fix that addresses the problems we are seeing in production.

It should be safe, as any delegation before an undelegation in the same slot would fail. Therefore, if we see a delegation, it must be more recent than the undelegation, while the opposite is not necessarily true.

This needs more careful thinking, but as a hotfix, it should be safe to ship.

PR #1273, opened 2026-06-02
fix: preserve streams on optimize failure

thlorenz:
## Summary

Fixes stream optimization so failed ChainLaser resubscriptions do not leave accounts without their existing active streams.

### magicblock-chainlink

- Updates the ChainLaser stream manager to build replacement optimized streams before changing the current account stream topology.
- Keeps the previous unoptimized/current stream set intact when an optimization subscribe call fails, avoiding a temporary drop of active account coverage.
- Adds mock stream-factory support for targeted subscribe failures and a regression test covering partial optimization failure followed by a successful retry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved failure handling during stream optimization to ensure existing stream configuration is preserved if errors occur during updates.

* **Tests**
  * Added test coverage for failure recovery scenarios during stream rebuilding operations.


GabrielePicco (approved):
LGTM!

PR #1277, opened 2026-06-02
Handle oversized single-stage committor transactions

GabrielePicco:
## Summary
- classify RPC sendTransaction responses for oversized versioned transactions as split-recoverable limit errors
- reuse the existing two-stage committor fallback when single-stage execution is too large
- add focused coverage for the oversized transaction classifier


taco-paco:
That works. Will address the properly in follow-ups

PR #1278, opened 2026-06-02
fix: execute post-delegation actions after clone

GabrielePicco:
## Summary
- Move post-delegation actions behind clone completion by emitting a dedicated top-level executor instruction immediately after clone or final clone-continue.
- The executor validates the immediately previous top-level Magic clone sibling through the instruction sysvar before invoking actions, so actions run once and only after the delegated account has been cloned. Note: direct invoke from clone is not viable as there is re-entracy if post-delegations actions commit/undelegated
- Depend on the [magicblock-svm solana-svm crate fix](https://github.com/magicblock-labs/magicblock-svm/tree/fix/post-action-executor-privileged-access/svm) from [magicblock-svm#24](https://github.com/magicblock-labs/magicblock-svm/pull/24) to keep that exact clone -> PostAct transaction shape privileged.
- NOTE: this makes post-delegation actions fully replicable as self-contained in a single transaction

## Breaking Changes
- [x] None

## Summary by CodeRabbit

* **New Features**
  * Post-delegation actions can be embedded in clone flows and executed via a new executor program; clone instructions carry explicit action payloads.
  * Task assembly can include a standalone-action noop instruction to support single-action flows.

* **Bug Fixes**
  * Stronger validation and ordering for delegated clones to prevent invalid or duplicate action-bearing clones and ordering regressions.

* **Tests**
  * Expanded unit and integration tests covering delegated cloning, post-delegation execution, and related invariants.


programs/magicblock/src/clone_account/common.rs
```diff
@@ -169,6 +174,162 @@ pub fn adjust_authority_lamports(
     Ok(())
 }
 
+const INSTRUCTIONS_SYSVAR_IDX: u16 = 2;
```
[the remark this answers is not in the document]
GabrielePicco: ix layout is fixed

magicblock-committor-service/src/tasks/utils.rs
```diff
@@ -65,13 +68,32 @@ impl TransactionUtils {
...
     ) -> TaskStrategistResult<VersionedTransaction> {
         let budget_instructions = Self::budget_instructions(
             Self::tasks_compute_units(tasks),
             compute_unit_price,
             Self::tasks_accounts_size_budget(tasks),
         );
-        let ixs = Self::tasks_instructions(&authority.pubkey(), tasks);
+        let mut ixs = Self::tasks_instructions(&authority.pubkey(), tasks);
+        if let Some(nonce) = standalone_action_nonce {
+            ixs.push(Self::standalone_action_noop_instruction(nonce));
```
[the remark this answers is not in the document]
GabrielePicco: compute units are over budgeted, so this is fine

magicblock-account-cloner/src/lib.rs
```diff
@@ -185,16 +181,26 @@ impl ChainlinkCloner {
...
+            pubkey, offset, data, is_last, actions,
         )
     }
 
     fn cleanup_ix(pubkey: Pubkey) -> Instruction {
         InstructionUtils::cleanup_partial_clone_instruction(pubkey)
     }
 
+    fn post_delegation_action_ix(
+        pubkey: Pubkey,
```
snawaz: What is this `pubkey` supposed to represent? A more descriptive name would help readability.

magicblock-magic-program-api/src/instruction.rs
```diff
@@ -357,3 +357,20 @@ pub enum CallbackInstruction {
     /// - **2..n** `[]`        Accounts required by the embedded instructions
     ExecuteCallback { instruction: Instruction },
 }
+
+/// Instruction(s) for the post-delegation action executor builtin-program.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
+pub enum PostDelegationActionInstruction {
```
snawaz: The name is confusing. 

- `PostDelegationActionExecutorInstruction` is less confusing? It's too long though. 
- `PostActProgramInstruction`? It also borrows `PostAct` part from the program-id itself. I kinda liked it.

programs/magicblock/src/clone_account/common.rs
```diff
@@ -169,6 +174,163 @@ pub fn adjust_authority_lamports(
...
+                invoke_context,
+                "Post-delegation action for {} uses validator authority as signer",
+                pubkey
+            );
+            return Err(
+                MagicBlockProgramError::PostDelegationActionUsesValidatorAuthority
+                    .into(),
+            );
+        }
+    }
```
snawaz, line 280: ❤️

programs/magicblock/src/clone_account/common.rs
```diff
@@ -234,25 +397,20 @@ pub fn set_account_from_fields(
...
-    //
+    // A delegated clone is still allowed to override a plain clone, which can
+    // happen when an eATA projection catches up after the underlying ATA was
+    // cloned without delegation metadata. Once the local target is delegated or
+    // undelegating, another delegated clone would re-run post-delegation
+    // actions and is rejected.
     if fields.delegated
-        && acc.delegated()
-        && acc.remote_slot() == fields.remote_slot
+        && fields.owner == dlp_api::id()
```
snawaz, line 408: concern: 

Is `fields.owner == dlp_api::id()` valid? I thought in case of `ATA`, we only change the special bit to indicate `delegated` but the owner remains unchanged, right?

snawaz (commented):
I skimmed through it. Haven't finished reviewing it yet. 

So far it looks good, especially the idea of executing actions inside a top-level ix. Looks clean! 

**It works on my machine!**

I posted one comment with the `concern` tag. 🤔 

Review still in progress.

magicblock-magic-program-api/src/instruction.rs
```diff
@@ -357,3 +357,20 @@ pub enum CallbackInstruction {
...
+    /// Executes post-delegation actions immediately after a matching delegated
+    /// clone instruction in the same transaction.
+    ///
+    /// # Account references
+    /// - **0.**   `[SIGNER]`  Validator authority
+    /// - **1.**   `[WRITE]`   Delegated clone target
+    /// - **2.**   `[]`        Instructions sysvar
+    /// - **3..n** `[]`        Accounts required by the embedded instructions
+    Execute {
+        pubkey: Pubkey,
```
snawaz: It is supposed to be `cloned_account`?

magicblock-account-cloner/src/lib.rs
GabrielePicco: Good point, I’ll rename it to delegated_account_pubkey for clarity.

magicblock-magic-program-api/src/instruction.rs
GabrielePicco: Fair, renamed to `PostDelegationActionExecutorInstruction`.

magicblock-magic-program-api/src/instruction.rs
GabrielePicco: Yes, it’s the cloned account pubkey, reanamed to `cloned_account_pubkey` for clarity

programs/magicblock/src/clone_account/common.rs
GabrielePicco: Yes, for ATA/eATA projection the owner should remain the token program owner. fields.owner == dlp_api::id() is not a valid way to identify delegated ATAs. In this path fields.delegated is the delegation signal; `fields.owner == dlp_api::id()` only means the incoming clone still looks DLP-owned, which is a different case (e.g. the delegated fees vault) and should not be used as the ATA condition.

magicblock-account-cloner/src/lib.rs
```diff
@@ -146,6 +114,39 @@ impl ChainlinkCloner {
...
+
+        Self::short_vec_len(tx.signatures.len())
+            + tx.signatures.len() * 64
+            + 3
+            + Self::short_vec_len(message.account_keys.len())
+            + message.account_keys.len() * 32
+            + 32
+            + Self::short_vec_len(message.instructions.len())
+            + instructions_size
+    }
```
snawaz: It seems to be missing the header size?

BTW, `bincode::serialize_size()` does the same thing? Maybe we can use that.

magicblock-magic-program-api/src/instruction.rs
```diff
@@ -357,3 +357,20 @@ pub enum CallbackInstruction {
...
+
+/// Instruction(s) for the post-delegation action executor builtin-program.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
+pub enum PostDelegationActionExecutorInstruction {
+    /// Executes post-delegation actions immediately after a matching delegated
+    /// clone instruction in the same transaction.
+    ///
+    /// # Account references
+    /// - **0.**   `[SIGNER]`  Validator authority
+    /// - **1.**   `[WRITE]`   Delegated clone target
```
snawaz: I don't see `Delegated clone target` being used as writable in the top-level ix. Is mutability actually required here?

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -485,6 +486,77 @@ where
...
+
+        self.ensure_delegation_action_dependencies(
+            request.pubkey,
+            request.account.remote_slot(),
+            &request.delegation_actions,
+        )
+        .await?;
+
+        Ok(self.clone_account_with_ownership(request).await?)
+    }
```
snawaz, line 521: Skipping this function along with `normalize_unresolved_dlp_clone_request()`. Lots of things unclear to me, but it works on my machine, so I trust it _for now_.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -596,8 +668,16 @@ where
...
+                let bank_slot = in_bank.remote_slot();
+                let update_slot = account.remote_slot();
+                let same_slot_delegated_refresh = bank_slot == update_slot
+                    && account.delegated()
+                    && (!in_bank.delegated() || in_bank.undelegating());
+                if bank_slot > update_slot
+                    || (bank_slot == update_slot
+                        && !same_slot_delegated_refresh)
+                {
+                    Some(bank_slot)
```
snawaz, line 681: Q: why `(bank_slot == update_slot && !same_slot_delegated_refresh)` is considered out-of-order-slot?

programs/magicblock/src/clone_account/common.rs
snawaz: ah, I see. I realized it's an additional check and the original check is moved below. I think you're checking delegation-through-DLP (which is different from ATA delegation). 🤔

programs/magicblock/src/clone_account/common.rs
```diff
@@ -234,25 +399,20 @@ pub fn set_account_from_fields(
...
+        && !is_validator_magic_fee_vault(pubkey)
     {
+        return Err(
+            MagicBlockProgramError::DelegatedCloneOwnerIsDelegationProgram
+                .into(),
+        );
+    } else if fields.delegated && (acc.delegated() || acc.undelegating()) {
```
snawaz: concern: 

Are we missing `&& acc.remote_slot() == fields.remote_slot` check here? Otherwise it should not be an error (i.e redelegation is allowed as long as it happens on different slots)?

magicblock-account-cloner/src/lib.rs
```diff
@@ -146,6 +114,39 @@ impl ChainlinkCloner {
...
+    fn short_vec_len(mut len: usize) -> usize {
+        let mut size = 1;
+        while len >= 0x80 {
+            len >>= 7;
+            size += 1;
+        }
+        size
+    }
+
+    fn transaction_size(tx: &Transaction) -> usize {
```
bmuddha: 🤯 I somehow doubt that this custom rolled impl is way more performant than bincode/wincode `serialized_size`, it will return an error which can be matched against in case if u16::MAX is overflowed.

magicblock-magic-program-api/src/instruction.rs
GabrielePicco: Good catch. The executor itself only reads the delegated flag from account 1. Writability is only required when an embedded post-delegation action marks that account writable, and append_action_accounts already upgrades duplicate metas in that case. So we can make the base cloned-account meta readonly and let action metas upgrade it when needed.

magicblock-committor-service/src/intent_executor/mod.rs
```diff
@@ -222,11 +222,12 @@ where
             .await?;
 
             // Standalone actions executed in single stage
-            let strategy = TaskStrategist::build_strategy(
+            let mut strategy = TaskStrategist::build_strategy(
                 commit_tasks,
                 &self.authority.pubkey(),
                 persister,
             )?;
+            strategy.standalone_action_nonce = Some(intent_bundle.id);
```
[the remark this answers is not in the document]
GabrielePicco: not critical and avoiding and nonce is a temporary hotfix, will be removed with an in-protocol nonce

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
GabrielePicco: Good question. “out-of-order” is a bit overloaded here. The intent is to ignore non-advancing subscription updates: if the bank already has the account at the same slot, then a normal update at that slot is treated as stale/duplicate and should not overwrite local state.

- same_slot_delegated_refresh is the exception. In the undelegate/redelegate same-slot path, the bank can still hold a plain or undelegating version while the subscription update carries the delegated state at the same slot, so we must allow that update through.  
- So the rule is: bank_slot > update_slot is truly out-of-order; bank_slot == update_slot && !same_slot_delegated_refresh is a duplicate/stale same-slot update. I can rename/comment this to make that distinction clearer.

programs/magicblock/src/clone_account/common.rs
GabrielePicco: Good catch. Fixed and added a test for it

magicblock-account-cloner/src/lib.rs
```diff
@@ -636,33 +607,33 @@ impl Cloner for ChainlinkCloner {
...
-                ClonerError::FailedToCloneRegularAccount(
-                    request.pubkey,
-                    Box::new(err),
-                )
-            })?;
-
-            self.send_actions_tx(actions_tx).await?;
+            let tx = self.build_small_account_tx(&request, blockhash);
+            let tx_size = Self::transaction_size(&tx);
+            if tx_size > MAX_INLINE_TRANSACTION_SIZE
```
bmuddha, line 614: technically we can have transactions which are greater than 64KB, by structuring it in such a way, that the transaction only has u8::MAX pubkeys, and u16::MAX instructions with u16::MAX  bytes each. This way it will serialize no problem and can be sanitized and executed

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
snawaz: Makes sense now. Thanks. 

I guess the variable `out_of_order_slot` needs to be renamed. That naming is always confusing. Maybe `non_advancing_slot` name, along with your comment in the code itself, would be helpful. These are edge-cases and touching these parts breaks things in a way that is hard to reproduce/catch (in tests especially).

snawaz (approved):
LGTM.

magicblock-account-cloner/src/lib.rs
GabrielePicco: Agreed, removed

magicblock-account-cloner/src/lib.rs
GabrielePicco: Noted, I will follow on

PR #1281, opened 2026-06-03
feat: added vrf ephemeral test queue, delegation record and metadata for mb-test-validator

jonasXchen:
## Summary
Adds a feature-gated ephemeral test queue (Sc9MJUng…) that is exempt from the per-request VRF fee — the same treatment as the canonical DEFAULT_EPHEMERAL_QUEUE — so a locally-created delegated queue can be served inside an Ephemeral Rollup without a fee-payer escrow. The exemption is centralized in is_fee_exempt_ephemeral_queue() and applied consistently across request_randomness, provide_randomness, and purge_expired_requests; the test queue is only compiled in under the ephemeral-test-queue Cargo feature, so production builds are unchanged.

## Breaking Changes
-  None
(Feature is off by default. Without ephemeral-test-queue, only DEFAULT_EPHEMERAL_QUEUE/5hBR5… is exempt — identical to current behavior. The three fee call-sites were refactored to a shared helper with no behavior change; the SDK gains additive constants.)

## Test Plan
Both build configs compile: cargo build-sbf --manifest-path program/Cargo.toml (production) and … --features ephemeral-test-queue.
Verified end-to-end locally: solana-test-validator (with the feature build + the pre-delegated Sc9MJ… queue/record/metadata) + ephemeral-validator + vrf-oracle (default paywJ identity) → roll-dice-delegated request was fulfilled in the ER (player state updated), with no fee charged and no Feepayer … was modified without being delegated / Sum of account balances … do not match errors.
Production fee path is unchanged: without the feature, is_fee_exempt_ephemeral_queue only matches DEFAULT_EPHEMERAL_QUEUE.

### DIFF
--- a/.github/packages/npm-package/mbTestValidator.ts
+++ b/.github/packages/npm-package/mbTestValidator.ts
@@ -87,6 +87,15 @@ function runMbTestValidator(): void {
     "--account",
     "FRqXJqfCi3o6gF3Yqnkx1gKA3YnbRDJbBs6hKpme3NHJ",
     p("FRqXJqfCi3o6gF3Yqnkx1gKA3YnbRDJbBs6hKpme3NHJ.json"),
+    "--account",
+    "Sc9MJUngNbQXSXGP3F67KvKwVnhaYn6kcioxXNVowYT",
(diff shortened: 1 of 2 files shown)


GabrielePicco (approved):
LGTM!

PR #1283, opened 2026-06-03
feat: bump filtering version

Dodecahedr0x:
## Summary
Bump filtering service version.

## Summary by CodeRabbit

* **Chores**
  * Updated query filtering service dependency across supported platforms (darwin-arm64, darwin-x64, linux-arm64, and linux-x64).


GabrielePicco:
@Dodecahedr0x approved, but this should be automated in the script like for the other packages

GabrielePicco (approved):

PR #1285, opened 2026-06-03
fix: use wire size (1232), not encoded size (1644), for tx fit checks!

snawaz:
- Use the serialized transaction wire size for committor fit checks because **Solana sendability** is capped by `PACKET_DATA_SIZE (1232 bytes)`, **not the RPC base64 string limit (1644)**.

- **bug**: the old encoded-length check was imprecise: raw sizes like 1231, 1232, and 1233 all encode to 1644 bytes, so we could accept a 1233-byte transaction that cannot be sent.

PR #1286, opened 2026-06-03
Preflight oversized committor transactions before delivery prep

GabrielePicco:
## Summary
- Reject oversized assembled committor transactions with `FailedToFitError` before they are sent.
- Split delivery preparation so the real ALT-backed transaction layout is checked before buffer accounts are created.
- Route pre-send `FailedToFitError` through the existing single-stage split fallback and recognize the shorter RPC `base64 encoded too large` error message.

## Root cause
Oversized single-stage transactions could still reach delivery preparation, or be classified only after `sendTransaction`, because the final assembled transaction size was not checked at the same boundary used for actual submission.

## Release Notes

* **Bug Fixes**
  * Enhanced detection of oversized transactions to catch additional error scenarios.
  * Improved error recovery handling for transaction preparation edge cases.

* **Refactor**
  * Optimized transaction preparation process with better sequencing of internal operations.

### DIFF
--- a/magicblock-committor-service/src/intent_executor/error.rs
+++ b/magicblock-committor-service/src/intent_executor/error.rs
@@ -51,6 +51,7 @@ impl InternalError {
                     RpcClientErrorKind::RpcError(
                         RpcError::RpcResponseError { code: -32602, message, .. }
                     ) if message.contains("VersionedTransaction too large")
+                        || message.contains("base64 encoded too large")
                 )
         }
 
@@ -433,13 +434,15 @@ mod tests {
 
     use super::{InternalError, TransactionStrategyExecutionError};
 
-    fn send_transaction_too_large_error() -> TransactionStrategyExecutionError {
+    fn send_transaction_error(
+        message: impl Into<String>,
+    ) -> TransactionStrategyExecutionError {
         let rpc_error = RpcClientError {
             request: Some(RpcRequest::SendTransaction),
             kind: Box::new(RpcClientErrorKind::RpcError(
                 RpcError::RpcResponseError {
                     code: -32602,
-                    message: "base64 encoded solana_transaction::versioned::VersionedTransaction too large: 1684 bytes (max: encoded/raw 1644/1232)".to_string(),
+                    message: message.into(),
                     data: RpcResponseErrorData::Empty,
                 },
             )),
@@ -454,7 +457,18 @@ mod tests {
 
     #[test]
     fn send_transaction_too_large_triggers_single_stage_split() {
-        let err = send_transaction_too_large_error();
+        let err = send_transaction_error(
+            "base64 encoded solana_transaction::versioned::VersionedTransaction too large: 1684 bytes (max: encoded/raw 1644/1232)",
+        );
+
+        assert!(err.is_single_stage_split_limit_error());
+    }
+
+    #[test]
+    fn send_transaction_short_too_large_message_triggers_single_stage_split() {
+        let err = send_transaction_error(
+            "Invalid Request: ErrorObject { code: InvalidParams, message: \"Invalid Request: base64 encoded too large\", data: None }",
+        );
 
         assert!(err.is_single_stage_split_limit_error());
     }
--- a/magicblock-committor-service/src/intent_executor/mod.rs
+++ b/magicblock-committor-service/src/intent_executor/mod.rs
@@ -326,7 +326,7 @@ where
         // We can recover that Error by splitting execution
         // in 2 stages - commit & finalize
         // Otherwise we return error
-        let execution_err = match res {
+        let (execution_err, cleanup_lookup_tables) = match res {
             Err(IntentExecutorError::FailedToFinalizeError {
                 err,
                 commit_signature: _,
@@ -334,17 +334,28 @@ where
             }) if !committed_pubkeys.is_empty()
                 && err.is_single_stage_split_limit_error() =>
             {
-                Some(err)
+                (Some(err), true)
             }
             Err(IntentExecutorError::FailedFinalizePreparationError(
                 TransactionPreparatorError::FailedToFitError,
-            )) if !committed_pubkeys.is_empty() => None,
+            )) if !committed_pubkeys.is_empty() => (None, true),
+            Err(IntentExecutorError::FailedFinalizePreparationError(
+                TransactionPreparatorError::PreflightFailedToFitError,
+            )) if !committed_pubkeys.is_empty() => (None, false),
             res => {
                 let signature = res.as_ref().ok().copied();
                 single_stage_executor
                     .execute_callbacks(signature, res.as_ref().map(|_| ()));
-                let transaction_strategy =
+                let mut transaction_strategy =
                     single_stage_executor.consume_strategy();
+                if matches!(
+                    &res,
+                    Err(IntentExecutorError::FailedFinalizePreparationError(
+                        TransactionPreparatorError::PreflightFailedToFitError
+                    ))
+                ) {
+                    transaction_strategy.lookup_tables_keys.clear();
+                }
                 execution_report.dispose(transaction_strategy);
                 return res.map(ExecutionOutput::SingleStage);
             }
@@ -355,7 +366,11 @@ where
         // Note that this not necessarily will pass at the end due to the same reason
         let strategy = single_stage_executor.consume_strategy();
         let (commit_strategy, finalize_strategy, cleanup) =
-            handle_cpi_limit_error(&self.authority.pubkey(), strategy);
+            handle_cpi_limit_error(
+                &self.authority.pubkey(),
+                strategy,
+                cleanup_lookup_tables,
+            );
         execution_report.dispose(cleanup);
         if let Some(execution_err) = execution_err {
             execution_report.add_patched_error(execution_err);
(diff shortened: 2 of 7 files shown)


GabrielePicco:
@coderabbit review

magicblock-committor-service/src/intent_executor/mod.rs
```diff
@@ -334,8 +334,11 @@ where
             }) if !committed_pubkeys.is_empty()
                 && err.is_single_stage_split_limit_error() =>
             {
-                err
+                Some(err)
             }
+            Err(IntentExecutorError::FailedFinalizePreparationError(
+                TransactionPreparatorError::FailedToFitError,
+            )) if !committed_pubkeys.is_empty() => None,
```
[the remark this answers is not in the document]
GabrielePicco: Fixed in ba21952d.

magicblock-committor-service/src/intent_executor/error.rs
```diff
@@ -51,6 +51,7 @@ impl InternalError {
                     RpcClientErrorKind::RpcError(
                         RpcError::RpcResponseError { code: -32602, message, .. }
                     ) if message.contains("VersionedTransaction too large")
+                        || message.contains("base64 encoded too large")
```
snawaz, line 54: I wonder where this message `"base64 encoded too large"` come from. While sending tx, the wire data is not encoded, as it does not make sense to encode the data because encoding makes the data _bigger_.

So I'm kinda puzzled.

magicblock-committor-service/src/transaction_preparator/mod.rs
```diff
@@ -90,27 +92,42 @@ impl TransactionPreparator for TransactionPreparatorImpl {
...
 
-        let message =
+        let tx =
             TransactionUtils::assemble_tasks_tx_with_standalone_action_nonce(
                 authority,
                 &tx_strategy.optimized_tasks,
                 self.compute_budget_config.compute_unit_price,
                 &lookup_tables,
                 tx_strategy.standalone_action_nonce,
+            )?;
```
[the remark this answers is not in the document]
GabrielePicco: cleanup no longer releases lookup-table keys that were never reserved.

magicblock-committor-service/src/intent_executor/mod.rs
```diff
@@ -326,22 +326,36 @@ where
...
                 finalize_signature: _,
             }) if !committed_pubkeys.is_empty()
                 && err.is_single_stage_split_limit_error() =>
             {
-                err
+                (Some(err), true)
             }
+            Err(IntentExecutorError::FailedFinalizePreparationError(
+                TransactionPreparatorError::FailedToFitError,
+            )) if !committed_pubkeys.is_empty() => (None, true),
```
taco-paco, line 341: If this error is handled by splitting into two stage flow, I'd assume we want to add it into patched report. Anything that is recoverd would be there, why do we return None here?

magicblock-committor-service/src/intent_executor/mod.rs
```diff
@@ -352,9 +366,15 @@ where
         // Note that this not necessarily will pass at the end due to the same reason
         let strategy = single_stage_executor.consume_strategy();
         let (commit_strategy, finalize_strategy, cleanup) =
-            handle_cpi_limit_error(&self.authority.pubkey(), strategy);
+            handle_cpi_limit_error(
+                &self.authority.pubkey(),
+                strategy,
+                cleanup_lookup_tables,
```
taco-paco, line 372: Additionally the tests are missing for new logic. To validate this case please add test into `test-integration/test-committor-service/tests/test_intent_executor.rs`

magicblock-committor-service/src/transaction_preparator/delivery_preparator.rs
```diff
@@ -75,21 +75,39 @@ impl DeliveryPreparator {
...
+        authority: &Keypair,
+        strategy: &mut TransactionStrategy,
+        persister: &Option<P>,
+    ) -> DeliveryPreparatorResult<Vec<AddressLookupTableAccount>> {
+        let lookup_tables = self
+            .prepare_lookup_tables_for_delivery(authority, strategy)
+            .await?;
+        self.prepare_tasks_for_delivery(authority, strategy, persister)
+            .await?;
         Ok(lookup_tables)
```
taco-paco, line 111: What is the reason for split up here?
 Now instead of preparing in parallel we prepare ALTs first and then buffers. 

Both of them need to be prepared and one doesn't depend on another, those should be prepared in parallel as before

magicblock-committor-service/src/transaction_preparator/mod.rs
```diff
@@ -90,27 +92,42 @@ impl TransactionPreparator for TransactionPreparatorImpl {
...
+                }
+                TaskStrategistError::SignerError(err) => {
+                    TransactionPreparatorError::SignerError(err)
+                }
+            })?;
         }
 
-        // Pre tx preparations. Create buffer accs + lookup tables
+        // Prepare lookup tables before buffers so real ALT layout is checked
+        // before creating buffer accounts.
```
taco-paco, line 107: Could you elaborate on what is ALT layout check? Why buffer preparation should be concerned with it?

Buffer preparation is agnostic to ALTs and don't use it, additionally buffer accounts are deterministically defined from commit id and will be the same anyway. If they're in ALTs they will be there in next run.

Buffer preparation also handles already prepared state. For now I don't see reason for this split as it will just make preparation slower

magicblock-committor-service/src/intent_executor/mod.rs
```diff
@@ -352,9 +366,15 @@ where
...
+            );
         execution_report.dispose(cleanup);
-        execution_report.add_patched_error(execution_err);
+        if let Some(execution_err) = execution_err {
```
taco-paco, line 375: If we got here it means we handled some error by splitting into two stage, that error should be added in report.

magicblock-committor-service/src/transaction_preparator/mod.rs
```diff
@@ -90,27 +92,42 @@ impl TransactionPreparator for TransactionPreparatorImpl {
                 self.compute_budget_config.compute_unit_price,
                 &dummy_lookup_tables,
                 tx_strategy.standalone_action_nonce,
-            )?;
+            )
+            .map_err(|err| match err {
+                TaskStrategistError::FailedToFitError => {
```
taco-paco, line 97: This highlights my doubts on need of new `PreflightFailedToFitError` variant.

 `FailedToFitError` is error returned "preflight". TaskStrategist returns it preflight, new `assemble_tasks_tx_with_standalone_action_nonce` returns it preflight.

What is a reason for another enum variant?

Open to renaming it if it is confusing.

magicblock-committor-service/src/intent_executor/mod.rs
```diff
@@ -326,22 +326,36 @@ where
...
             {
-                err
+                (Some(err), true)
             }
+            Err(IntentExecutorError::FailedFinalizePreparationError(
+                TransactionPreparatorError::FailedToFitError,
+            )) if !committed_pubkeys.is_empty() => (None, true),
+            Err(IntentExecutorError::FailedFinalizePreparationError(
+                TransactionPreparatorError::PreflightFailedToFitError,
+            )) if !committed_pubkeys.is_empty() => (None, false),
```
taco-paco, line 344: Why we return false here while true for `FailedToFitError`?

magicblock-committor-service/src/intent_executor/mod.rs
taco-paco: In case we don't want some errors in report for whatever reason, we could add `fn reportable(&self) -> bool` on Error which would be a cleaner solution

magicblock-committor-service/src/intent_executor/mod.rs
taco-paco: sane [as](https://github.com/magicblock-labs/magicblock-validator/pull/1286/changes#r3353912518)

magicblock-committor-service/src/intent_executor/mod.rs
```diff
@@ -352,9 +366,15 @@ where
         // Note that this not necessarily will pass at the end due to the same reason
         let strategy = single_stage_executor.consume_strategy();
         let (commit_strategy, finalize_strategy, cleanup) =
-            handle_cpi_limit_error(&self.authority.pubkey(), strategy);
+            handle_cpi_limit_error(
```
taco-paco, line 369: Let's rename function for clarity as it not only handle cpi limit error now

taco-paco (changes requested):
Thanks for the fix! 
Overall seems like the main point was
```rust
        if serialize_and_encode_base64(&tx).len() > MAX_ENCODED_TRANSACTION_SIZE
        {
            return Err(TaskStrategistError::FailedToFitError);
        }
```

But PR included other changes on delivery preparation. Hope in the future we could make PRs more atomic. 

Added comments on both matters.

Additionally as new patching logic appeared it would be good add tests in corresponding file that checks exactly patching correctness and recovery of intents. See
`test-integration/test-committor-service/tests/test_intent_executor.rs`

magicblock-committor-service/src/transaction_preparator/delivery_preparator.rs
taco-paco: can be discussed in this [thread](https://github.com/magicblock-labs/magicblock-validator/pull/1286#discussion_r3353878158)

GabrielePicco:
Closing as this PR seems the proper fix: https://github.com/magicblock-labs/magicblock-validator/pull/1285 rather than the defensive logic applied here

PR #1287, opened 2026-06-03
Handle ALT extend invalid instruction data

GabrielePicco:
## Summary

Problem: Table Mania retried the same active lookup table when an ALT extend transaction failed with `InstructionError(2, InvalidInstructionData)`. That could keep replaying the same rejected extend path and let committor preparation fail instead of moving the reservation forward.

Solution: classify `InvalidInstructionData` at the existing-table ALT extend instruction, skip the normal retry path for that table, and create a fresh lookup table with a minimal one-key init payload. Other extend failures still use the existing bounded retry behavior.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced lookup table extension with improved error detection and resilience mechanisms
  * Added automatic retry and fallback strategies when table extension encounters specific errors

* **Chores**
  * Updated dependencies for error handling capabilities

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4285,6 +4285,7 @@ dependencies = [
  "solana-signer",
  "solana-slot-hashes 3.0.1",
  "solana-transaction",
+ "solana-transaction-error 3.2.0",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
--- a/magicblock-table-mania/Cargo.toml
+++ b/magicblock-table-mania/Cargo.toml
@@ -30,6 +30,7 @@ solana-transaction = { workspace = true }
 solana-compute-budget-interface = { workspace = true }
 solana-instruction = { workspace = true }
 solana-message = { workspace = true }
+solana-transaction-error = { workspace = true }
 tokio = { workspace = true }
 
 [features]
--- a/magicblock-table-mania/src/error.rs
+++ b/magicblock-table-mania/src/error.rs
@@ -35,4 +35,67 @@ impl TableManiaError {
             _ => None,
         }
     }
+
+    pub fn is_sent_transaction_invalid_instruction_data_at(
+        &self,
+        instruction_index: u8,
+    ) -> bool {
+        use magicblock_rpc_client::MagicBlockRpcClientError;
+        use solana_instruction::error::InstructionError;
+        use solana_transaction_error::TransactionError;
+
+        match self {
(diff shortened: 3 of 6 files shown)


GabrielePicco:
@CodeRabbit review

thlorenz (approved):
LGTM with a minor concern. You can decide if that is an issue or not:

This fallback only skips the failed table for the current reservation attempt. Since the table is not marked non-extendable, it can be selected again later if it becomes `active_tables.last()` again, e.g. after the fallback table is released while the original table still has reservations. Should we persistently quarantine/mark the table as no-longer-extendable after `InvalidInstructionData`, while still allowing it to satisfy existing lookups?

If it is ok to try to extend the same table again in the future and then end up in the same fallback then it is not an issue, but wanted to point it ou.

* * *

Step-by-step details of the issue:

1. `reserve_new_pubkeys` only attempts to extend `active_tables.last()` when that table is not full.
   - See `magicblock-table-mania/src/manager.rs:300-303`.

2. When extending Table A fails with `InvalidInstructionData`, the PR sets `force_new_table = true`.
   - See `magicblock-table-mania/src/manager.rs:317-324`.

3. `force_new_table` is only a local variable for the current loop iteration.
   - It is initialized on each iteration at `magicblock-table-mania/src/manager.rs:294`.
   - Nothing is stored on Table A to indicate that future extends should be skipped.

4. The fallback then creates Table B and pushes it to `active_tables`.
   - See `magicblock-table-mania/src/manager.rs:376-377`.

5. Immediately after this, future reservations will try Table B first because it is now `active_tables.last()`.

6. However, Table A remains active and is not marked full/non-extendable.
   - Fullness is based only on local pubkey count: `magicblock-table-mania/src/lookup_table_rc.rs:350-357`.

7. If Table B is later released while Table A still has reservations, `release_pubkeys` can remove B from `active_tables` and keep A.
   - See `magicblock-table-mania/src/manager.rs:552-563`.

8. At that point, `active_tables` may become `[A]` again.

9. Since A is still not full locally and is again `active_tables.last()`, a new reservation can try to extend A again.

10. So the fallback avoids retrying A only in the immediate flow, but does not persistently prevent future attempts to extend A.

GabrielePicco:
> mark the table as no-longer-extendable after `InvalidInstructionData`

Fair point, added a `table.mark_non_extendable()` on this case

PR #1290, opened 2026-06-04
fix: reject short account responses

thlorenz:
## Summary

Reject malformed multi-account RPC responses when the result count does not match the requested
account count.

### magicblock-chainlink

- Validates multi-account RPC response lengths before converting results into remote accounts.
- Returns the existing structured account-resolution error for malformed RPC-only fetch results.
- Uses the existing pending-request error notification path for background fetches so all waiters are resolved and pending fetch state is cleaned up.
- Adds regression coverage for shortened multi-account responses by extending the RPC test mock with an opt-in truncation mode.

## Bug Fixes

* Improved error handling for blockchain RPC responses to verify returned account data matches the number of accounts requested. The system now detects and reports mismatches, preventing unresponsive behavior when RPC servers return incomplete or truncated results.


GabrielePicco (approved):
LGTM!

PR #1291, opened 2026-06-04
fix(committor): race-condition on cleanup

taco-paco:
## Summary

Cleaning buffers on failure isn't safe due to potential race condition:
Assume pubkey set A being committed
Intent1 fails and cleans up, another Intent2 with set A executes right away
That could lead for Intent2 init of buffers executing prior of Intent1 buffer cleanup
With same set A buffers will have same address

Due to this PR cleans up only once intent succeeded. This is safe for buffers as commit_nonce advances and next buffers address will be different

## Breaking Changes
- [ ] None

## Bug Fixes
* Strengthened buffer cleanup reliability by refining execution timing to occur only after successful intent bundle processing, preventing potential race conditions when multiple operations target the same buffers simultaneously
* Enhanced error diagnostics for buffer cleanup failures by including specific buffer address information in error logs, improving troubleshooting capabilities

### DIFF
--- a/magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
+++ b/magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
@@ -283,6 +283,7 @@ where
 
         // Execute an Intent
         let result = executor.execute(intent.clone(), persister).await;
+        let execution_succeeded = result.inner.is_ok();
         let _ = result.inner.as_ref().inspect_err(|err| {
             error!(intent_id = intent.id, error = ?err, "Failed to execute intent bundle");
         });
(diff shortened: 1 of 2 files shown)


snawaz:
@taco-paco : My wild guess is that this issue is related with `InvalidDiff` issue in which case `CommitDiffFromBuffer` is invoked and the `buffer_account` is empty! So looks like it is cleaned-up before DLP ix is invoked?

So could it be related? 🤔 

image

ref: https://explorer.solana.com/tx/4fTNkzG47hX5JmitcQXdHEoDqPxxDMTXoW7bX5fx86xsnAUceaKHpnPBovs9V6b6wbErfC6wYVML4cDmoQLkdzhC?customUrl=https%3A%2F%2Fmainnet-tee-as.magicblock.app

GabrielePicco (approved):
LGTM!

PR #1293, opened 2026-06-04
fix(scheduler): remove block subscription in the scheduler

bmuddha:
## Summary by CodeRabbit

* **Refactor**
  * Restructured slot transition handling in the transaction scheduler. Slot updates are now driven exclusively by the primary-mode slot ticker, with replica-mode slots managed through existing block command handlers. Removed legacy subscription-based block production update mechanism.


GabrielePicco (approved):
LGTM!

PR #1300, opened 2026-06-05
feat: schedule automatic undelegation on actions and AML check failure

Dodecahedr0x:
## Summary
Automatically schedules the undelegation of AML-flagged accounts. Currently, if an account fails the AML check, it is just not cloned, meaning it stays owned by the DLP on mainnet but can't be used in the ER. This PR automatically undelegates it.

## Breaking Changes
- [X] None

## Summary by CodeRabbit

* **New Features**
  * Added undelegation scheduling capability for cloned accounts
  * Integrated AML risk assessment system into account operations
  * New post-delegation action instruction for triggering undelegation

* **Tests**
  * Enhanced integration test suite to cover AML risk scenarios and undelegation workflows
  * Expanded test infrastructure to support risk service integration


snawaz:
@Dodecahedr0x : can you please add some description as to when/why undelegation should be scheduled and automated.

magicblock-api/src/magic_validator.rs
```diff
@@ -99,6 +101,29 @@ use crate::{
 type InnerChainlinkImpl = ProdInnerChainlink<ChainlinkCloner>;
 
 type ChainlinkImpl = ProdChainlink<ChainlinkCloner>;
+/// Bridges chainlink's [`UndelegationScheduler`] to the committor service so a
+/// delegated clone rejected by AML is undelegated on the base layer.
+struct CommittorUndelegationScheduler(Arc<CommittorService>);
+
+#[async_trait::async_trait]
+impl UndelegationScheduler for CommittorUndelegationScheduler {
```
taco-paco: `magic_validator.rs` is already massive, let's extract this into separate file

taco-paco (changes requested):
I think this PR digresses from current `committor` flow, and imo there's no need in this deviation.
Delegation is an account state property, therefore it has to go through builtin program and modify delegated flags. 

This makes flow similar to regular Intent scheduling one, in fact you do schedule an Intent youself.
1. Validator detect account is malicious
2. Validator sends TX in magic-program(new entrypoint, or an old one with commit)
3. magic-program mutates account states correctly and schedules undelegation(`mark_account_as_undelegated `)
4. CommittorService picks it up and executes as in current flow

This gets rid of new `schedule_undelegation` functionality completely as it will seamlessly integrates into current Committor logic. All is needed is new entrypoint in magic-program and TX sender

Current state bypasses all of that flow to which it belongs. This additionally lead to an amigious state where in AccountsDB account is delegated but on Base it is not. Maybe cloner will fix this(which I'm not sure about) after some updates, but this leaves window for an attack where in meantime it could be used by user.

magicblock-committor-service/src/committor_processor.rs
```diff
@@ -263,6 +275,69 @@ impl CommittorProcessor {
     }
 }
 
+pub(crate) fn undelegation_intent_bundle(
```
taco-paco: This has to be created by `magic-program` which additionally would call `mark_account_as_undelegated`

magicblock-committor-service/src/committor_processor.rs
```diff
@@ -263,6 +275,69 @@ impl CommittorProcessor {
...
+        id: intent_id(),
+        slot: remote_slot,
+        blockhash: Default::default(),
+        sent_transaction: Default::default(),
+        payer: Default::default(),
+        intent_bundle,
+    }
+}
+
+fn intent_id() -> u64 {
```
taco-paco: Intent ids are assigned by `magic-program`

Dodecahedr0x:
@taco-paco I completely reworked the flow to use the regular commit path. However, it now clones and undelegates the account in a single transaction.

GabrielePicco (changes requested):
I think this PR now covers the same high-level case as #1352

A couple of details from #1352 still look worth carrying over here before we rely on this PR as the replacement:

  1. **Chunked clone pending state.** For large accounts, `needs_undelegation` currently puts `ScheduleUndelegation` in the same final tx after `CloneAccountContinue(is_last=true, actions=[])`. But `process_clone_account_continue` removes the pubkey from `PENDING_CLONES`
  immediately when `actions` is empty. If the scheduling instruction fails after that, `CleanupPartialClone` can no longer clean up the account. We should either keep the clone pending until `ScheduleUndelegation` succeeds and remove it in that processor, or split scheduling into a separate tx and skip cleanup on schedule-only failure. Given this PR’s sibling-executor design, keeping pending until the scheduler succeeds seems like the better fit.

  2. **Local undelegating flag.** `process_schedule_cloned_account_undelegation` schedules the base intent, but does not appear to mark the local cloned account as `undelegating`. The #1352 path reused `process_schedule_commit`, which marks the account undelegating and locks it locally. We should do the same here so the ER state immediately reflects that the account is leaving.

  I’d also port/adapt the small/large transaction-shape tests from #1352 to `needs_undelegation`, so we explicitly cover the small clone + schedule path and the chunked final-continue + schedule path.

Dodecahedr0x:
@GabrielePicco 

[e569134](https://github.com/magicblock-labs/magicblock-validator/pull/1300/commits/e569134066c600c4f1e53461abb7ec8f206389dd) adds equivalent tests for transaction shapes.

[226441d](https://github.com/magicblock-labs/magicblock-validator/pull/1300/commits/226441d9a9dd41bd2e8a266499c222e64082a171) fixes marking the account as undelegating.

[61f7cfe](https://github.com/magicblock-labs/magicblock-validator/pull/1300/commits/61f7cfe44a491d5c71058afa63304830cff16ed8) fixes the pending clone issues by passing the need for undelegation in the instruction.

GabrielePicco (approved):
LGTM!

PR #1314, opened 2026-06-12
Move validator unregister out of shutdown path

GabrielePicco:
## Summary
- Send the validator domain-registry unregister transaction immediately after the shutdown signal while RPC is still serving.
- Keep transaction confirmation at confirmed commitment, but run it as a best-effort background log task.
- Avoid waiting for confirmation during shutdown; after a successful send, the worst case is sent but unconfirmed in local logs.

## Why
The user-visible outage begins when RPC is cancelled, not when shutdown preparation starts. Awaiting the send before shutdown preparation keeps the transaction from being cancelled when the Tokio runtime drops, while confirmation no longer gates perceived downtime.

## Summary by CodeRabbit

* **New Features**
  * Configurable RPC commitment for blockchain operations.
  * Public APIs to submit unregistration without waiting and to submit-and-confirm in background (returns a task handle).

* **Improvements**
  * Shutdown flow now initiates background unregistration and avoids blocking on confirmation.
  * Transaction sending split into confirmed and non-confirmed paths for clearer, non-blocking behavior.
* **Behavior**
  * Unregistration now de-duplicates concurrent attempts and logs background confirmation progress.

### DIFF
--- a/magicblock-api/src/domain_registry_manager.rs
+++ b/magicblock-api/src/domain_registry_manager.rs
@@ -1,7 +1,10 @@
-use std::{io, time::Duration};
+use std::{io, sync::Arc, thread, time::Duration};
 
-use anyhow::{anyhow, Context};
+use anyhow::Context;
 use borsh::BorshDeserialize;
+use magicblock_rpc_client::{
+    MagicBlockSendTransactionConfig, MagicblockRpcClient,
+};
 use mdp::{
     consts::ER_RECORD_SEED,
     instructions::{sync::SyncInstruction, version::v0::SyncRecordV0},
@@ -18,11 +21,14 @@ use solana_sdk_ids::system_program;
 use solana_signature::Signature;
 use solana_signer::Signer;
 use solana_transaction::Transaction;
-use tokio::time::sleep;
-use tracing::info;
+use tracing::{error, info};
+
+const UNREGISTER_CONFIRMATION_TIMEOUT: Duration = Duration::from_secs(8);
+const UNREGISTER_CONFIRMATION_INTERVAL: Duration = Duration::from_millis(400);
 
 pub struct DomainRegistryManager {
-    client: RpcClient,
+    client: Arc<RpcClient>,
+    rpc_client: MagicblockRpcClient,
 }
 
 impl DomainRegistryManager {
@@ -34,8 +40,13 @@ impl DomainRegistryManager {
         url: impl ToString,
         commitment: CommitmentConfig,
     ) -> Self {
+        let client = Arc::new(RpcClient::new_with_commitment(
+            url.to_string(),
+            commitment,
+        ));
         Self {
-            client: RpcClient::new_with_commitment(url.to_string(), commitment),
+            client: client.clone(),
+            rpc_client: MagicblockRpcClient::new(client),
         }
     }
 
@@ -142,11 +153,6 @@ impl DomainRegistryManager {
     ) -> Result<Signature, Error> {
         let (pda, _) = Self::get_pda(&payer.pubkey());
 
-        // Verify existence to avoid failed tx costs
-        let _ = self
-            .fetch_validator_info(&pda)
-            .await?
-            .ok_or(Error::NoRegisteredValidatorError)?;
         self.send_instruction_without_confirmation(
             payer,
             pda,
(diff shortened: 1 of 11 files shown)


magicblock-validator/src/main.rs
```diff
@@ -156,9 +156,19 @@ async fn run() {
...
     api.prepare_ledger_for_shutdown();
     let stop_step = Instant::now();
     api.stop().await;
+    if let Some(handle) = unregister_handle {
+        if handle.is_finished() {
+            if let Err(err) = handle.await {
+                error!(error = ?err, "Unregister task failed");
+            }
+        } else {
+            debug!("Unregister still running; not waiting during shutdown");
```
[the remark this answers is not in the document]
GabrielePicco: send is confirmed now

magicblock-validator/src/main.rs
```diff
@@ -156,9 +156,21 @@ async fn run() {
         .await;
     }
 
+    let unregister_handle = api.start_unregister_validator_on_chain().await;
```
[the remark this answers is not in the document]
GabrielePicco: Fair, OK for now as very unlikely, but will be handled in the smart contract (e.g. with a nonce or timestamp to skip in case a new registration frontrun de-registration)

magicblock-api/src/magic_validator.rs
```diff
@@ -681,17 +682,75 @@ impl MagicValidator {
...
-        .inspect(|_| info!("Unregistered validator on chain"))
+        .inspect(|signature| {
+            info!(%signature, "Sent validator unregister transaction")
+        })
+    }
+
+    pub async fn start_unregister_validator_on_chain(
+        &self,
+    ) -> Option<tokio::task::JoinHandle<()>> {
+        if self.config.chain_operation.is_none()
```
thlorenz, line 690: Lots of added logic here not sure if this would be better in some other module instead of adding more code to this already big module here?

magicblock-api/src/domain_registry_manager.rs
```diff
@@ -180,28 +236,73 @@ impl DomainRegistryManager {
...
+        signature: Signature,
+    ) -> Result<(), Error> {
+        let manager = DomainRegistryManager::new_with_commitment(
+            url,
+            CommitmentConfig::confirmed(),
+        );
+        manager.confirm_signature(&signature).await
+    }
+
+    async fn confirm_signature(
```
thlorenz: This method is not at all specific to the domain registry manager.
I think it would be better to have this insight, the magic RPC client or some other transaction specific utility crate.

And it turns out we already [have a whole module for that there](https://github.com/magicblock-labs/magicblock-validator/tree/master/magicblock-rpc-client/src/signature_confirmer.rs)

thlorenz (approved):
LGTM after addressing the nits I pointed out.

magicblock-api/src/domain_registry_manager.rs
```diff
@@ -130,6 +136,27 @@ impl DomainRegistryManager {
         Ok(())
     }
 
+    async fn send_unregister(
+        &self,
+        payer: &Keypair,
+    ) -> Result<Signature, Error> {
+        let (pda, _) = Self::get_pda(&payer.pubkey());
+
+        // Verify existence to avoid failed tx costs
```
Dodecahedr0x: AFAIK, a primary validator crashes if it fails to register. If so, this can be assumed to be true since the coordination mode check is already there

magicblock-api/src/magic_validator.rs
```diff
@@ -681,17 +682,75 @@ impl MagicValidator {
...
+            || !matches!(self.config.lifecycle, LifecycleMode::Ephemeral)
+            || !CoordinationMode::current().needs_onchain_interactions()
+        {
+            return None;
+        }
+
+        let rpc_url = self.config.rpc_url().to_owned();
+        let step_start = Instant::now();
+        let signature =
```
Dodecahedr0x: Why not spawn a task with the send and confirm, instead of awaiting for the send before shutting down? In both case, an error during the sending is not handled

Dodecahedr0x (approved):
LGTM, small comments

magicblock-api/src/magic_validator.rs
GabrielePicco: This ensure send is done before shutdown, while confirmation is not blocking

PR #1319, opened 2026-06-14
fix: correct ledger replay startup

thlorenz:
## Summary

Fix ledger replay correctness during validator startup by making replay start at the first
block actually available in the ledger and by waiting for queued replay transactions to finish
before reporting replay completion.

**NOTE**: I found those issues while trying to make validator authority override during startup work.
          We've since given up on that feature, but the fixes are still important so I supply them via this
          PR.

### Ledger replay start slot

Ledger replay now clamps the blockhash warmup scan to the ledger's first available slot. This prevents replay from stopping immediately when the requested warmup range starts before the first persisted block, which can happen when the ledger does not contain slot 0 or has been pruned/compacted.

If the ledger has no blocks, replay is skipped and startup continues from the requested starting slot.

### Replay completion semantics

Ledger replay now waits for the scheduler to drain all previously queued replay transactions before `process_ledger` returns. Replay transaction submission is fire-and-forget, so queueing all transactions was not enough to guarantee that the replayed account state had actually been applied.

The scheduler now supports an ordered drain barrier that is acknowledged only once it is idle, making replay completion mean that the queued replayed transactions have executed and committed.

## Summary by CodeRabbit

* **New Features**
  * Added drain synchronization to the transaction scheduler, allowing callers to wait until replayed transactions fully complete before continuing.
* **Bug Fixes**
  * Improved ledger processing to correctly compute replay scan start based on ledger retention, and to wait for replay drain after replay finishes.
* **Behavior Changes**
  * Scheduler now processes drain requests as an ordered “idle barrier,” guaranteeing no queued work is left unfinished when proceeding.


magicblock-ledger/src/blockstore_processor/mod.rs
```diff
@@ -167,17 +167,38 @@ pub async fn process_ledger(
 ) -> LedgerResult<u64> {
     // Since transactions may refer to blockhashes that were present when they
     // ran initially we ensure that they are present during replay as well
-    let blockhashes_only_starting_slot =
+    let requested_blockhash_start =
         full_process_starting_slot.saturating_sub(max_age);
+    // The ledger may not have a persisted block at slot 0 (or at any slot below
```
bmuddha: that scenario is impossible, accountsdb cannot be reset to slot 0 or to any slot which is not present in the ledger, in fact it's impossible to reset accountsdb slot without removing the database entirely, but this implies a total state loss without any way to recover it. The comment is thus confusing and the logic is redundant.

bmuddha (approved):
LGTM, with minor nit

magicblock-ledger/src/blockstore_processor/mod.rs
thlorenz: I checked the startup path again, and I don’t think the scenario is impossible.

`accountsdb.reset` and `ledger.reset` are independent config options. If `accountsdb.reset = true` while `ledger.reset = false`, `AccountsDb::new` removes only `accountsdb/main` and initializes a fresh DB whose storage header slot starts at `0`. The existing ledger is kept as-is, and it may not contain slot `0` - normally the first produced block is slot `1`, and older slots may also have been pruned.

In that case `maybe_process_ledger()` calls `process_ledger()` with `full_process_starting_slot = accountsdb.slot()`, i.e. `0`. Without clamping the scan start to `ledger.get_lowest_slot()`, replay starts by trying `ledger.get_block(0)`, gets `None`, breaks immediately, and replays nothing.

So agreed that the comment wording could be clearer it’s not a slot-only reset of AccountsDb, it’s a fresh AccountsDb after wiping the DB directory  (and I fixed that in aade62b4),  but the logic itself is needed for the supported _reset AccountsDb but keep ledger_ configuration.

PR #1320, opened 2026-06-14
chore: remove replay authority override

thlorenz:
## Summary

Remove the unused ledger replay validator-authority override feature.

### Configuration

Removes `ledger.replay-authority-override` from `LedgerConfig` and the example config. This also removes the corresponding environment variable support for `MBV_LEDGER__REPLAY_AUTHORITY_OVERRIDE` through normal config deserialization.

The replica-mode authority override remains supported via `validator.replication-mode.replica.authority-override`.

### Ledger restore tests

Removes integration test helpers and duplicate restore test variants that existed only to exercise the replay override path. The remaining ledger restore tests now use the standard validator setup paths.

## Release Notes

* **Chores**
  * Removed the optional validator authority override setting used during ledger replay.
  * Updated ledger restore and replay tests/utilities to stop accepting or using authority override inputs, and to consistently use standard validator initialization.
  * Adjusted related helper signatures and test assertions to match the simplified restore/replay workflow.


bmuddha (approved):
LGTM

PR #1301, opened 2026-06-06
fix: prevent ephemeral accounts from being evicted

snawaz:
Currently ephemeral accounts are being evicted:

> EvictAccount: evicted ‘29Bdguts6QpejFEoLCnW7MojGzW3syHQ5rxQJZ6ak7tv’

After this PR (though, ideally, the prevention should be on the caller side):

image

PR #1216, opened 2026-05-19
fix: delegated account evictions

thlorenz:
## Summary

Prevent delegated accounts from being capacity-evicted while preserving the subscription cleanup path for accounts that are safely delegated and not undelegating.

**Note:** Avoiding delegated-account eviction races adds meaningful orchestration complexity:
cleanup must only drop direct delegated-account subscriptions when safe, while undelegation
must retain or restore protected subscription ownership. This PR handles that coordination, but
**we may want to revisit whether dropping delegated-account subscriptions is still worth the
complexity** now that gRPC subscriptions are working reliably.

### magicblock-chainlink

Adds capacity eviction protection for delegated accounts by allowing the remote account provider to consult the local account bank before evicting subscriptions. Delegated accounts that are currently undelegating are treated as protected so they remain subscribed while undelegation is in progress.

Delegated-account cleanup is routed through a silent direct-subscription release path. This lets the fetch cloner remove direct subscription ownership for accounts that are already delegated without disrupting undelegation-driven subscription ownership.

The subscription LRU now supports filtered capacity eviction so protected entries can be skipped while still freeing capacity from eligible subscriptions. Diagnostics and errors were expanded to make failed capacity eviction attempts easier to understand.

### Tests

Adds coverage for silent delegated direct cleanup, protected capacity eviction, delegated fetch-cloner behavior, and the interaction between delegated cleanup and undelegation protection.

## Summary by CodeRabbit

* **New Features**
  * Runtime-configurable eviction protection to keep delegated/undelegating accounts from being evicted; LRU insertion respects protected entries and reports when no evictable capacity exists.
  * Exposed undelegation-tracking registration so tracking can be acquired externally.

* **Bug Fixes**
  * Improved delegated-account cleanup to remove only direct subscriptions while preserving undelegation tracking and avoiding unintended unsubscriptions; removal notifications skip protected accounts.

* **Tests**
  * Added extensive tests covering delegated cleanup, eviction-prioritization, and subscription-release semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/magicblock-labs/magicblock-validator/pull/1216?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1323, opened 2026-06-14
fix(chainlink): prevent ephermal account from being evicted

### LINKED ISSUE #1302: bug:  prevent ephemeral accounts from being evicted
Currently `EvictAccount` instruction is being invoked with ephemeral accounts as well.

> EvictAccount: evicted ‘29Bdguts6QpejFEoLCnW7MojGzW3syHQ5rxQJZ6ak7tv’

A [hotfix][1] is implemented that ensures the instruction fails. However, **it should be done properly such that `EvictAccount` is not invoked to begin with**.

[1]: https://github.com/magicblock-labs/magicblock-validator/pull/1301

0xzrf:
## Summary

Prevent validator-native ephemeral accounts from entering chainlink's subscription eviction pipeline, so `EvictAccount` is never submitted for them. This implements the proper caller-side fix for [#1302](https://github.com/magicblock-labs/magicblock-validator/issues/1302), following the same layered approach used for delegated accounts in [#1216](https://github.com/magicblock-labs/magicblock-validator/pull/1216).

Changes in `magicblock-chainlink`:

- Add `should_schedule_bank_eviction()` and use it in the removal listener and stale subscription-update path to skip delegated, undelegating, and ephemeral bank state.
- Extend `CapacityEvictionProtection` with an `ephemeral` flag so LRU capacity eviction skips live ephemeral entries (predicate uses `ephemeral && owner != default` so post-eviction tombstones and normal cloned accounts are unaffected).
- Skip pubsub/LRU registration entirely when acquiring a subscription for a live ephemeral account.
- Add unit/integration tests covering LRU filtering, capacity eviction, subscription acquire, stale updates, and removal-listener behavior.

## Breaking Changes

- None

## Test Plan

- `cargo test -p magicblock-chainlink --lib ephemeral`
- `cargo test -p magicblock-chainlink --lib test_should_schedule_bank_eviction`

New tests:

- `test_should_schedule_bank_eviction` — helper covers normal, delegated, undelegating, and ephemeral accounts
- `test_subscribe_account_removals_skips_ephemeral_and_evicts_normal` — removal listener skips ephemeral, still evicts normal cloned accounts
- `test_ephemeral_account_is_protected_from_capacity_eviction` — LRU evicts a non-protected candidate instead of a live ephemeral entry
- `test_acquire_subscription_skips_ephemeral_account` — no pubsub or LRU entry for ephemeral acquire
- `test_stale_subscription_update_does_not_notify_removal_for_ephemeral` — stale account-sub updates do not enqueue removal for ephemeral; still do for normal accounts
- `test_lru_cache_skips_ephemeral_protected_candidate_and_evicts_next` — LRU filter unit test

Closes #1302 (caller-side fix; on-chain guard from #1301 remains as defense in depth)

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined account eviction and removal logic for ephemeral accounts
  * Updated subscription handling to properly assess account eligibility for removal notifications

* **Tests**
  * Added comprehensive test coverage for ephemeral account behavior during capacity eviction
  * Added test coverage for subscription update scenarios with protected accounts

### DIFF
--- a/magicblock-chainlink/src/remote_account_provider/mod.rs
+++ b/magicblock-chainlink/src/remote_account_provider/mod.rs
@@ -342,11 +342,12 @@ pub(crate) enum SubscriptionReleaseMode {
 pub(crate) struct CapacityEvictionProtection {
     pub delegated: bool,
     pub undelegating: bool,
+    pub ephemeral: bool,
 }
 
 impl CapacityEvictionProtection {
     pub fn is_protected(self) -> bool {
-        self.delegated || self.undelegating
+        self.delegated || self.undelegating || self.ephemeral
     }
 }
 
@@ -861,6 +862,7 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
             CapacityEvictionProtection {
                 delegated: false,
                 undelegating: false,
+                ephemeral: false,
             },
         )
     }
@@ -1466,6 +1468,10 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
         reason: SubscriptionReason,
     ) -> RemoteAccountProviderResult<()> {
         // 1. First realize subscription
+        if self.capacity_eviction_protection_for(pubkey).ephemeral {
+            trace!(pubkey = %pubkey, "Skipping subscription for ephemeral account");
+            return Ok(());
+        }
         self.pubsub_client.subscribe(*pubkey, None).await?;
 
         // 2. Add to LRU cache
--- a/magicblock-chainlink/src/remote_account_provider/mod.rs
+++ b/magicblock-chainlink/src/remote_account_provider/mod.rs
@@ -1467,11 +1467,6 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
         pubkey: &Pubkey,
         reason: SubscriptionReason,
     ) -> RemoteAccountProviderResult<()> {
-        // 1. First realize subscription
-        if self.capacity_eviction_protection_for(pubkey).ephemeral {
-            trace!(pubkey = %pubkey, "Skipping subscription for ephemeral account");
-            return Ok(());
-        }
         self.pubsub_client.subscribe(*pubkey, None).await?;
 
         // 2. Add to LRU cache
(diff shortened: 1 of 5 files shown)


magicblock-chainlink/src/remote_account_provider/mod.rs
```diff
@@ -1466,6 +1468,10 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
         reason: SubscriptionReason,
     ) -> RemoteAccountProviderResult<()> {
         // 1. First realize subscription
+        if self.capacity_eviction_protection_for(pubkey).ephemeral {
```
thlorenz: This still records subscription ownership after `register_subscription` no-ops for an ephemeral account. That makes the first acquire look successful with ownership but no pubsub/LRU entry. A second acquire then takes the `existing` branch above and adds the pubkey to `lrucache_subscribed_accounts` without ever subscribing, so `is_watching()` can become true for an account that is not actually subscribed. Release has the same mismatch: it may try to unsubscribe a subscription that was never created and restore the phantom ownership on failure.

Could we make the skip path explicit to the caller, or move the ephemeral check before ownership is recorded, so skipped ephemeral acquires leave no ownership and no LRU state? The existing `test_acquire_subscription_skips_ephemeral_account` only checks the first acquire, so it misses this second-acquire and release mismatch.

This is basically equivalent to what [coderabbit complained about as well](https://github.com/magicblock-labs/magicblock-validator/pull/1323/changes#r3410108808)

thlorenz (changes requested):
Thanks for adding protections around ephemeral accounts and covering the main LRU/removal paths. I found one blocking issue in the subscription skip path: returning `Ok(())` from `register_subscription` still causes `acquire_subscription_with_mode` to record ownership, so a later acquire can put the ephemeral account into the LRU without any pubsub subscription, and release can leave phantom ownership behind. That breaks the intended "skip pubsub/LRU registration entirely" behavior and can make `is_watching` inaccurate for ephemeral accounts.

0xzrf:
hey @thlorenz 
I've made the appropriate changes as reviewed

Can you review them?
Thanks.

thlorenz:
Hi, in the meantime, we fixed this with another [PR](https://github.com/magicblock-labs/magicblock-validator/pull/1353) which is very similar to yours, but it's a lot less lines of code because it updates existing test to ensure this fixes instead of creating entirely new tests.

I hope you understand that we prefer that smaller PR.
Keep up the good work and don't be demotivated by this, it's a normal part of development to find two solutions to the same problem and have to choose one.

0xzrf:
@thlorenz yes, a project should always do what's best for them, and a +20 diff is definitely a much more idiomatic solution here

Looking forward to keep contributing!
Thanks for reviewing the PR

PR #1328, opened 2026-06-15
fix: silent overflow for WS port

bzawisto:
## Summary
The fix closes a startup bug where setting the RPC listen port to the maximum value (65535) would break the validator. Because the WebSocket port is derived as "listen + 1," that value had no valid result - it led to unexpected random port in release builds. 

## Test Plan
Unit tests

## Release Notes

* **Bug Fixes**
  * Enhanced aperture listen port validation to prevent invalid configurations and runtime errors
  * Improved WebSocket port derivation logic to safely handle edge cases

* **Tests**
  * Added comprehensive tests for aperture listen port configuration validation


GabrielePicco (approved):
LGTM!

PR #1329, opened 2026-06-15
Fix DLP program update discovery

GabrielePicco:
## Summary
- Forward DLP-owned program subscription updates to delegated account discovery without classifying them by account data bytes in the pubsub actors.
- Drop internal DLP record updates only after greedy discovery misses, so metadata records are not cloned from program subscriptions.
- Add regressions for delegated application data that begins with an internal record discriminator and for real internal DLP record updates.

## Root Cause
The pubsub actors filtered DLP program updates using only the account data prefix. Delegated application accounts can contain bytes that match internal DLP record discriminators, which caused valid delegated-account updates to be dropped before discovery could fetch the derived delegation-record PDA.

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Enhanced processing of subscription updates from programs to properly filter internal DLP program accounts and prevent unnecessary cloning operations.
  * Updated delegation record forwarding logic to ensure correct handling of DLP account updates.

* **Tests**
  * Added comprehensive test coverage for internal DLP delegation record discovery and subscription handling scenarios.


magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -654,6 +654,19 @@ where
             return;
         }
 
+        if matches!(update.source, SubscriptionSource::Program)
+            && update.account.is_owned_by_delegation_program()
+            && update.account.fresh_account().is_some_and(|account| {
+                is_internal_dlp_account_data(account.data())
+            })
```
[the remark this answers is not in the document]
GabrielePicco: This is ok, because not delegated account are subscribed, and delegated accounts don't have updates

magicblock-chainlink/src/chainlink/fetch_cloner/tests.rs
```diff
@@ -3228,6 +3228,164 @@ async fn test_discovered_dlp_owned_account_without_delegation_record_is_ignored(
...
+) {
+    init_logger();
+    let validator_keypair = Keypair::new();
+    let validator_pubkey = validator_keypair.pubkey();
+    let account_pubkey = random_pubkey();
+    let account_owner = random_pubkey();
+    const CURRENT_SLOT: u64 = 100;
+
+    let mut app_data = vec![0; DelegationRecord::size_with_discriminator()];
+    app_data[..8].copy_from_slice(&100u64.to_le_bytes());
```
thlorenz, line 4255: This test basically shows how a user could manufacture an account that looks like a delegation record. Then that account could be delegated, and now the check returns positive, even though it is not really a delegation record.

This was not introduced in this PR, but I'm not sure if this is desired.

thlorenz (approved):
I am approving this, even though I don't fully understand it's usefulness.
Account updates received via program subs are already only passed on if we also have a direct subscription to that account.
Since we are never subscribed to delegation records for a long time, they should not be passed on in the first place.
At least this is how it's supposed to work and I'm not sure if this got broken somehow along the way.

The only reason for this to exist that I see is that for this short time where we subscribe to a delegation record we will then receive an update and want to ignore it.
However, this means that we might miss an important update because the direct subscriptions might not work sufficiently. Same risk as for any other account where we also use program subs to backup the direct subs.

GabrielePicco:
> I am approving this, even though I don't fully understand it's usefulness. Account updates received via program subs are already only passed on if we also have a direct subscription to that account. Since we are never subscribed to delegation records for a long time, they should not be passed on in the first place. At least this is how it's supposed to work and I'm not sure if this got broken somehow along the way.
> 
> The only reason for this to exist that I see is that for this short time where we subscribe to a delegation record we will then receive an update and want to ignore it. However, this means that we might miss an important update because the direct subscriptions might not work sufficiently. Same risk as for any other account where we also use program subs to backup the direct subs.

 Thanks, this is the intended boundary change. The old behavior was “only forward if directly subscribed,” but this PR deliberately makes DLP-owned program updates an exception so FetchCloner can attempt greedy discovery first.                                              
                                                                                                                                                                                                                                                                                 
The important part is placement: `is_internal_dlp_account_data` is only a byte-shape check, not proof that the pubkey is an internal DLP PDA. So using it in the pubsub actor before discovery can drop a legitimate delegated user account. In this PR we first derive/fetch the `delegation record` for the updated pubkey; only if that discovery misses do we drop internal-looking DLP program-sub updates as DLP internal noise.                                                                                                                              
                                                                                                                                                                                                                                                                         
Direct account-sub updates are unchanged. So if we are directly subscribed to a delegation record, the account-sub path can still process it. The new drop only applies to the program-sub copy after discovery failed.

magicblock-chainlink/src/chainlink/fetch_cloner/tests.rs
GabrielePicco: Agreed, and that test is meant to show exactly that limitation. A user account can have application data that parses like a delegation record prefix. That does not make the account a delegation record, which is enforced

PR #1332, opened 2026-06-16
feat(rpc): add shared base-layer blockhash registry

### LINKED ISSUE #1258: Add a base-layer blockhash registry
## Goal

Add a base-layer blockhash registry.

## Why

The current behavior or architecture does not provide this outcome.

## Scope

**Problem**
Many services make Base layer request for latest blockhash, that result in too many requests

**Proposed Solution**
Create a singleton that would serve a purpose of providing consumers a valid blockhash, thus minimizing RPC usage

0xzrf:
## Summary

Shared `BaseLayerBlockhash` registry refreshed every 10s at validator startup; consumers read lock-free instead of calling `getLatestBlockhash` per transaction. Closes [#1258](https://github.com/magicblock-labs/magicblock-validator/issues/1258). Ephemeral blockhash path unchanged.

## Changes

- **`magicblock-rpc-client`**: `BaseLayerBlockhash` + background refresher; `MagicblockRpcClient::new_with_registry`; registry-first `get_latest_blockhash()` with legacy fallback; `invalidate_cached_blockhash()` clears registry too.
- **`magicblock-api`**: spawn refresher in `try_from_config`; pass registry to committor, claim-fees, domain registry.
- **`magicblock-committor-service`**: `ChainConfig.blockhash_registry`; single `new_with_registry` call in committor processor.
- **`magicblock-validator-admin`**: claim-fees reads from registry, RPC fallback on miss.

## Breaking Changes
- None (operators)

No config, CLI, disk, or ephemeral RPC changes. Internal API only: `claim_fees`, `ClaimFeesTask::start`, `DomainRegistryManager` constructors/static helpers, and `ChainConfig` gain an optional `BaseLayerBlockhash` param (defaults to `None` in test helpers).

## Test Plan

- [x ] Existing tests pass (committor tests still use registry-less clients)

No new unit tests — wiring only; stale-blockhash retries still go through existing invalidation paths.

## Summary by CodeRabbit

* **Performance Improvements**
  * Implemented shared blockhash caching with automatic background refresh to reduce RPC load.
  * Transaction construction now prioritizes cached blockhashes, improving speed and reducing network dependency.
  * Validator operations including registration, unregistration, and fee claiming now benefit from optimized blockhash handling.

### DIFF
--- a/magicblock-committor-service/src/committor_processor.rs
+++ b/magicblock-committor-service/src/committor_processor.rs
@@ -71,22 +71,12 @@ impl CommittorProcessor {
         );
         let rpc_client = Arc::new(rpc_client);
         let websocket_uri = chain_config.websocket_uri.clone();
-        let magic_block_rpc_client = match (chain_slot, websocket_uri) {
-            (Some(chain_slot), websocket_uri) => {
-                MagicblockRpcClient::new_with_chain_slot_and_websocket(
-                    rpc_client,
-                    chain_slot,
-                    websocket_uri,
-                )
-            }
-            (None, Some(websocket_uri)) => {
-                MagicblockRpcClient::new_with_websocket(
-                    rpc_client,
-                    Some(websocket_uri),
-                )
-            }
-            (None, None) => MagicblockRpcClient::new(rpc_client),
-        };
+        let magic_block_rpc_client = MagicblockRpcClient::new_with_registry(
+            rpc_client,
+            chain_slot,
+            chain_config.websocket_uri.clone(),
+            chain_config.blockhash_registry.clone(),
+        );
 
         // Create TableMania
         let gc_config = GarbageCollectorConfig::default();
--- a/magicblock-rpc-client/src/lib.rs
+++ b/magicblock-rpc-client/src/lib.rs
@@ -275,7 +275,7 @@ pub struct BaseLayerBlockhash {
 impl BaseLayerBlockhash {
     /// Lock-free read of the cached blockhash, if the refresher has
     /// populated it at least once and it hasn't been invalidated.
-    pub fn get(&self) -> Option<CachedBlockhash> {
+    fn get(&self) -> Option<CachedBlockhash> {
         self.inner.load().as_deref().copied()
     }
 
--- a/magicblock-rpc-client/src/lib.rs
+++ b/magicblock-rpc-client/src/lib.rs
@@ -12,6 +12,7 @@ use std::{
     time::{Duration, Instant},
 };
 
+use arc_swap::ArcSwapOption;
 use futures_util::future::try_join_all;
 use serde_json::json;
 use signature_confirmer::{SignatureConfirmer, SignatureConfirmerConfig};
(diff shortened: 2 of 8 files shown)


magicblock-rpc-client/src/lib.rs
```diff
@@ -254,14 +256,108 @@ struct BlockhashCache {
...
+    /// Lock-free read of the cached blockhash, if the refresher has
+    /// populated it at least once and it hasn't been invalidated.
+    pub fn get(&self) -> Option<CachedBlockhash> {
+        self.inner.load().as_deref().copied()
+    }
+
+    /// Returns the cached base-layer blockhash, if available.
+    pub fn latest_blockhash(&self) -> Option<Hash> {
+        self.get().map(|cached| cached.blockhash)
+    }
```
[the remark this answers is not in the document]
0xzrf: I don't think that will be the case since the blockhash will be updated in every customisable time

thlorenz:
@copilot resolve the merge conflicts in this pull request

magicblock-rpc-client/src/lib.rs
```diff
@@ -254,14 +256,108 @@ struct BlockhashCache {
...
+
+impl BaseLayerBlockhash {
+    /// Lock-free read of the cached blockhash, if the refresher has
+    /// populated it at least once and it hasn't been invalidated.
+    fn get(&self) -> Option<CachedBlockhash> {
+        self.inner.load().as_deref().copied()
+    }
+
+    /// Returns the cached base-layer blockhash, if available.
+    pub fn latest_blockhash(&self) -> Option<Hash> {
```
thlorenz, line 283: This should apply a freshness check before exposing the registry value.
If `refresh_once` starts failing after a value has been stored, the old value remains in `ArcSwapOption` and every caller will keep signing with it forever. That affects both `MagicblockRpcClient::get_latest_blockhash()` and direct users like claim fees/domain registry, since they treat any registry value as valid and never fall back to RPC. 
Once the blockhash ages past Solana's validity window, those transactions can fail continuously until a successful refresh or process restart.

Can we expire the cached value here (or in `get()`) based on `fetched_at`, and return `None` when it is too old so callers fall back to a direct fetch? The TTL should be longer than the 10s refresh interval but comfortably below blockhash expiry.

magicblock-committor-service/src/committor_processor.rs
```diff
@@ -71,22 +71,12 @@ impl CommittorProcessor {
         );
         let rpc_client = Arc::new(rpc_client);
         let websocket_uri = chain_config.websocket_uri.clone();
```
thlorenz, line 73: This local is no longer used after the constructor simplification, and `cargo check` reports it as an unused variable. 
Please remove it to keep the crate warning-free.

```
warning: unused variable: `websocket_uri`
  --> magicblock-committor-service/src/committor_processor.rs:73:13
   |
73 |         let websocket_uri = chain_config.websocket_uri.clone();
   |             ^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_websocket_uri`
   |
   = note: `#[warn(unused_variables)]` (part of `#[warn(unused)]`) on by default
```

thlorenz (changes requested):
I found one blocking freshness issue: the registry can keep serving an expired blockhash indefinitely if the refresher stops succeeding, which would make every registry-backed transaction fail until the process restarts or a retry path invalidates it.

 I also noted one small cleanup for an unused variable.

PR #1334, opened 2026-06-16
feat: Adding backpressure to subscription update reader

bzawisto:
## Summary
The listener drained the bounded update channel into an unbounded set of clone/fetch tasks. Now a Semaphore caps in-flight tasks — a permit is taken before reading the channel and released when the task finishes, so a full pile stops draining the channel and backpressure reaches the WS layer. The 5_000 channel size is now shared as const.

This is rather unlike to happen since the consumer should be fast enough. There's additional metric added which is useful to detect if reader becomes saturated. 

## Summary of Changes

- **Refactor**
  - Improved subscription update processing to better cap concurrency, drain completed work promptly, and ensure all spawned tasks finish cleanly on shutdown.
- **New Features**
  - Added a Prometheus gauge metric to track the number of in-flight subscription updates, with helper functions to increment and decrement it.
- **Chores**
  - Centralized the default subscription update limit into a shared internal constant for consistent configuration.


bzawisto:
@CodeRabbit review

thlorenz:
@coderabbitai resume

thlorenz (approved):
Looks good. The subscription update listener now bounds concurrent processing with a semaphore while
 preserving the existing bounded channel, and the in-flight gauge is registered and balanced by a drop
 guard around task processing.

GabrielePicco (approved):

PR #1336, opened 2026-06-16
fix: check locks early

Dodecahedr0x:
## Summary
Check how many accounts the transaction contains before doing any work.

## Breaking Changes
- [X] None


bmuddha (approved):
LGTM

PR #1337, opened 2026-06-17
fix: various correctness related fixes

bmuddha:
- bound the number on gSS request
- do not start slot ticker (actions drain) in non-primary mode
- apply replica authority override in all instruction processors
- limit websocket message size to 1MB
- serve duplicate signature subscritpions
- prevent race conditions cloning large accounts
- chainlink: prevent race conditions during cloning

## Summary

* **New Features**
  * Added a 1 MiB maximum WebSocket message size after upgrade.
* **Bug Fixes**
  * Enforced a limit of 256 signatures per signature-status request.
  * Improved WebSocket signature subscriptions for cached results and corrected cleanup of expired one-shot subscribers.
  * Prevented duplicate concurrent clone submissions by coordinating clone work per key and added additional cloning race coverage.
* **Chores**
  * Updated validator authority checks and related authorization/validation logic to use the effective validator authority.

### DIFF
--- a/magicblock-aperture/src/state/subscriptions.rs
+++ b/magicblock-aperture/src/state/subscriptions.rs
@@ -35,7 +35,7 @@ pub(crate) type AccountSubscriptionsDb =
 /// Manages subscriptions to accounts owned by a specific program. Maps a program `Pubkey` to its subscribers.
 pub(crate) type ProgramSubscriptionsDb =
     Arc<scc::HashMap<Pubkey, UpdateSubscribers<ProgramAccountEncoder>>>;
-/// Manages one-shot subscriptions for transaction signature statuses. Maps a `Signature` to its subscriber.
+/// Manages one-shot subscriptions for transaction signature statuses.
 pub(crate) type SignatureSubscriptionsDb =
     Arc<scc::HashMap<Signature, UpdateSubscriber<TransactionResultEncoder>>>;
 /// Manages subscriptions to all transaction logs.
@@ -178,13 +178,32 @@ impl SubscriptionsDb {
         signature: Signature,
         chan: WsConnectionChannel,
     ) -> (SubscriptionID, Arc<AtomicBool>) {
-        let encoder = TransactionResultEncoder;
-        let subscriber = self
+        let mut subscriber = self
             .signatures
             .entry_async(signature)
             .await
-            .or_insert_with(|| UpdateSubscriber::new(Some(chan), encoder));
-        (subscriber.id, subscriber.live.clone())
+            .or_insert_with(|| {
+                UpdateSubscriber::new(None, TransactionResultEncoder)
+            });
+        subscriber.txs.insert(chan.id, chan.tx);
+        let id = subscriber.id;
+        let live = subscriber.live.clone();
+        (id, live)
+    }
+
+    /// Removes one pending signature subscription.
+    pub(crate) async fn remove_signature_subscriber(
+        &self,
+        signature: &Signature,
+        conid: ConnectionID,
+    ) {
+        let Some(mut entry) = self.signatures.get_async(signature).await else {
+            return;
+        };
+        entry.txs.remove(&conid);
+        if entry.txs.is_empty() {
+            let _ = entry.remove();
+        }
     }
 
     /// Sends a notification to a signature subscriber and removes the subscription.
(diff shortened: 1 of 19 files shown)


GabrielePicco (approved):
LGTM!

magicblock-aperture/src/state/subscriptions.rs
```diff
@@ -178,13 +178,32 @@ impl SubscriptionsDb {
...
+        let mut subscriber = self
             .signatures
             .entry_async(signature)
             .await
-            .or_insert_with(|| UpdateSubscriber::new(Some(chan), encoder));
-        (subscriber.id, subscriber.live.clone())
+            .or_insert_with(|| {
+                UpdateSubscriber::new(None, TransactionResultEncoder)
+            });
+        subscriber.txs.insert(chan.id, chan.tx);
```
thlorenz, line 188: This still collapses duplicate `signatureSubscribe` requests from the same websocket connection. Because the map is keyed only by `ConnectionID`, a second subscription for the same signature on the same connection overwrites the first sender, returns the same subscription id, and adds another expirer entry for the same `(signature, conid)`. 
When the older expirer entry fires it can remove the connection entry and cancel the newer subscription before its own TTL, so F-040 is only fixed for duplicate subscriptions from different connections.

Can we track each signature subscription by its returned `SubscriptionID` as well, or otherwise allow multiple entries per connection, so every subscribe call gets an independent id and lifecycle?

programs/magicblock/src/schedule_task/process_execute_task.rs
```diff
@@ -55,7 +55,8 @@ pub(crate) fn process_execute_crank(
             transaction_context,
             VALIDATOR_IDX,
         )?;
-        if validator_pubkey != &validator_authority_id() {
+        let validator_authority = effective_validator_authority_id();
```
thlorenz, line 58: Switching the processors to `effective_validator_authority_id()` is incomplete without updating the internal instruction builders that create these calls. For example, `InstructionUtils::execute_task_instruction`, `accept_scheduled_commits_instruction`, `modify_accounts_instruction`, clone/finalize helpers, etc. still add `validator_authority_id()` as the signer account. With an `authority_override` configured, those internally generated transactions will carry the real validator identity account while the processor now expects the override pubkey, so scheduled task execution and scheduled commit/ticker flows will fail with `IncorrectAuthority` instead of being fixed.

Can this change also update the instruction construction/signing path, or add override-mode tests around the scheduled task and scheduled commit execution paths to prove the generated instructions pass the new validation?

thlorenz (changes requested):
I found a couple of blocking gaps that leave the fixes incomplete: duplicate signature subscriptions still collapse per connection, and the authority override validation is not paired with the internal instruction builders that create those validator-authorized instructions.

Separately, I am not a huge fan of combining all these unrelated fixes into one PR. The websocket limits/subscription behavior, chainlink clone race handling, slot ticker lifecycle, and Magic Program authority changes are independent and would be much easier to review, correlate and safely backport as separate PRs.

Also there is still a coderabbit comment that needs to be resolved

magicblock-aperture/src/state/subscriptions.rs
bmuddha: that would be incorrect from protocol spec standpoint, since a connection cannot subscribe for the same update multiple times, agave follows the same common sense

programs/magicblock/src/schedule_task/process_execute_task.rs
bmuddha: That's intended, since in replica mode none of those helpers can be invoked: 1. Replica cannot generate transactions, only replay them 2. Replica might not have access to keypair to sign the transaction, thus making the use of those helpers impossible anyway.

thlorenz (approved):
LGTM

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -444,51 +446,152 @@ where
...
+    ) -> ClonerResult<Signature> {
+        let program_id = program.program_id;
+        let remote_slot = program.remote_slot;
+
+        loop {
+            if self
+                .accounts_bank
+                .get_account(&program_id)
+                .is_some_and(|account| account.remote_slot() >= remote_slot)
+            {
```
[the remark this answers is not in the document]
bmuddha: that's intended

PR #1338, opened 2026-06-17
Enable sBPF v3 runtime feature

GabrielePicco:
## Summary
- Activate the Agave sBPF v3 deployment/execution feature in the processor SVM environment.
- Persist the feature account alongside the existing runtime feature accounts.

## Why
Cloned Devnet programs can now require sBPF v3. Without this feature active locally, LoaderV4 rejects those ELF files during deploy verification with an unsupported sBPF version error surfaced as invalid account data.


bmuddha (approved):
LGTM

PR #1348, opened 2026-06-19
feat: enable sub-compaction

taco-paco:
## Summary

Enable prallelization in compaction using sub-compactions > 1.

## Breaking Changes
- [ ] None

## Summary by CodeRabbit

* **Chores**
  * Tuned database concurrency to better match system workload priority, improving throughput under load.
* **Bug Fixes**
  * Updated transaction validation limits to use a runtime-derived maximum instead of a fixed constant, improving compatibility with supported transaction shapes and reducing out-of-range rejections.


bmuddha (approved):
LGTM

PR #1350, opened 2026-06-19
Fix primary on-chain registration

GabrielePicco:
## Summary
- Treat standalone and primary replication modes as validators that manage on-chain registration.
- Spawn the existing primary on-chain setup for ephemeral primary nodes after replication starts.
- Start recurring fee claims for primary validators using the same primary/standalone eligibility predicate.
- Allow primary nodes through unregister eligibility while replicas remain excluded.

## Root cause
Primary replication mode started the replication service but skipped `spawn_primary_onchain_setup`, which was only called for standalone validators. After enabling primary registration, the recurring fee-claim task also needed to use the same primary/standalone eligibility instead of the older standalone-only gate.

## Validation
- `cargo test -p magicblock-api magic_validator::tests`
- `cargo check -p magicblock-api`


bmuddha (approved):
lgtm

PR #1353, opened 2026-06-22
fix: skip ephemeral account evictions

### LINKED ISSUE #1302: bug:  prevent ephemeral accounts from being evicted

thlorenz:
## Summary

Prevent chainlink from submitting `EvictAccount` for local ephemeral accounts when account-removal notifications are processed.

CLOSES: #1302

### magicblock-chainlink

- Extends the removal preflight to treat ephemeral bank accounts as protected state, alongside delegated and undelegating accounts.
- Documents the key invariant directly in the code: ephemeral accounts are local-validator state and must not update on base chain, so dropping the remote subscription cannot make the local ephemeral bank state stale.
- Adds regression coverage for delegated, undelegating, ephemeral, and normal account-removal behavior so only normal accounts are evicted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ephemeral bank accounts are now protected from eviction, ensuring proper account lifecycle management alongside delegated accounts.

### DIFF
--- a/magicblock-chainlink/src/chainlink/mod.rs
+++ b/magicblock-chainlink/src/chainlink/mod.rs
@@ -288,19 +288,24 @@ impl<T: ChainRpcClient, U: ChainPubsubClient, V: AccountsBank, C: Cloner>
 
         task::spawn(async move {
             while let Some(pubkey) = removed_accounts_rx.recv().await {
-                // Pre-flight check: skip if delegated/undelegating
+                // Pre-flight check: skip if delegated/undelegating/ephemeral
                 // (the processor enforces this too, but this avoids
-                // the overhead of building and submitting a doomed tx)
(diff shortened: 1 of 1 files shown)


snawaz (approved):
Thanks. 

Looks good to me!

bzawisto (approved):

PR #1354, opened 2026-06-22
fix: prevent subscription reconcile race

### LINKED ISSUE #1015: enhancement: prevent unsub/sub race condition
Right now we do unsubscribe accounts that we only query but don't clone at the end of the task that needed that information.
One example are delegation records.

However there is a potential race condition such that if a delegation record is cloned + subscribed to in the meantime we would still unsubscribe from it in the end.

More details and a plan to fix this issue are in the comments.

Since the cases for which this matters are very rare this is a lower priority problem.

thlorenz:
## Summary

Fix a race where subscription reconciliation could unsubscribe an account while a normal subscription registration was between the pubsub subscribe and LRU tracking steps.

CLOSES: #1015

### magicblock-chainlink

The subscription reconciler now coordinates with the existing per-pubkey subscription transition locks. When reconciliation detects drift, it locks the affected pubkey, rechecks the current LRU/pubsub state, and only then performs a repair. This prevents stale reconciliation snapshots from undoing an in-progress registration while preserving the reconciler's ability to clean up real pubsub-only drift and resubscribe LRU-owned accounts missing from pubsub.

Added regression coverage for the pubsub-before-LRU race by pausing the mock pubsub client after insertion but before subscription acquisition completes. Added follow-up coverage proving lock-aware reconciliation still removes truly stale pubsub subscriptions and restores missing pubsub subscriptions for accounts still owned by the LRU.

Test/dev-only helpers were kept out of the main provider implementation flow, and the lock helper documents why creating a missing per-pubkey lock during reconciliation is intentional.

### DIFF
--- a/magicblock-chainlink/src/remote_account_provider/chain_pubsub_client.rs
+++ b/magicblock-chainlink/src/remote_account_provider/chain_pubsub_client.rs
@@ -319,6 +319,10 @@ pub mod mock {
         pending_program_subscribe_failures: Arc<Mutex<usize>>,
         reconnectable: Arc<Mutex<bool>>,
         subscribe_blocked: Arc<Mutex<bool>>,
+        subscribe_pause_after_insert: Arc<Mutex<bool>>,
+        subscribe_insertions: Arc<AtomicU64>,
+        subscribe_insert_notify: Arc<Notify>,
+        subscribe_continue_notify: Arc<Notify>,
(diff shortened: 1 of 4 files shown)


thlorenz:
@coderabbitai resume

bzawisto:
LGTM

bzawisto (approved):

PR #1362, opened 2026-06-26
fix: omit reconnecting pubsub clients from reconciliation

thlorenz:
## Summary

Fix subscription reconciliation while pubsub clients are reconnecting:

- Make SubMux reconciliation snapshots consider only currently connected inner clients.
- Skip noisy reconciliation repair when no pubsub clients are connected.
- Add regression coverage for reconnecting clients and the all-disconnected reconciliation window.
- Document the Chainlink/SubMux reconciliation behavior for future agents.

### magicblock-chainlink

SubMux now reports subscription union/intersection based on connected clients only. Reconnecting clients no longer make the reconciler think healthy subscriptions are missing, and reconciler-triggered repair continues to fan out only to connected clients.

The pubsub client abstraction now exposes whether reconciliation is meaningful. SubMux uses this to report that reconciliation should be skipped when every inner client is disconnected or reconnecting, preventing full-LRU false positives while reconnect loops resubscribe from the authoritative trackers.

The subscription reconciler now checks that availability before taking pubsub snapshots. If no connected client is available, it leaves repair to reconnect handling for that tick instead of logging or attempting to resubscribe every LRU account.

### Tests and docs

Added unit coverage for:

- ignoring reconnecting clients in SubMux reconciliation snapshots;
- skipping reconciler repair when no SubMux clients are connected.

Updated the Chainlink agent documentation to capture the connected-client-only reconciliation behavior and the reconnect catch-up responsibility.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subscription reconciliation to only account for subscriptions from currently connected clients.
  * Reconciliation now avoids union/intersection noise when no live reconciliation snapshot is available, skipping repair/noisy reporting for that cycle.
  * Reconnecting/disconnected clients no longer affect reconciliation until they complete reconnect and rejoin the connected set.
* **Tests**
  * Added reconciliation tests covering reconnecting-only clients and the no-connected-clients scenario, plus updated timing assertions for connection-state transitions.

### DIFF
--- a/.agents/context/crates/magicblock-chainlink.md
+++ b/.agents/context/crates/magicblock-chainlink.md
@@ -321,6 +321,8 @@ Removal handling is serialized with same-pubkey subscription transitions via `ev
 
 If subscription metrics are enabled, a background task periodically runs `subscription_reconciler::reconcile_subscriptions` to compare the LRU with actual pubsub-client subscriptions, update metrics, and notify removal for subscriptions that vanished.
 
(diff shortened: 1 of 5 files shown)


bzawisto:
LGTM

bzawisto (approved):

PR #1363, opened 2026-06-26
feat: Establish nats connection in the background

bzawisto:
## Summary
In case of warm restart - nats connection is establish by replicator in the background (separate thread). If preserves the same terminating behaviour in case of failure. When there's an empty account db it performs blocking connect on main thread as it needs to pull the snapshot anyway before moving forward.

## Summary by CodeRabbit

* **New Features**
  * Replication startup now supports both immediate and deferred broker connection, improving bootstrap handling for fresh and already-running nodes.
  * Message handling now consistently tracks whether acknowledgements are required for different replication events.

* **Bug Fixes**
  * Shutdown draining now reports both drained and acknowledged messages, giving clearer visibility into replication cleanup.
  * Added more detailed timing logs for key replication startup and shutdown steps.

### DIFF
--- a/magicblock-api/src/magic_validator.rs
+++ b/magicblock-api/src/magic_validator.rs
@@ -58,7 +58,7 @@ use magicblock_program::{
     validator::{self, validator_authority},
     TransactionScheduler as ActionTransactionScheduler,
 };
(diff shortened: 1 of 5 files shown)


bmuddha (approved):
LGTM

PR #1368, opened 2026-06-29
feat: persist and recover committor intent execution

### LINKED ISSUE #1262: Gracefully shut down the committor processor
## Goal

Gracefully shut down the committor processor.

## Why

The current behavior or architecture does not provide this outcome.

## Scope

**Problem**
On shutdown all ongoing intents should ideally be finalized and results processed, aborting amid way could lead to errors

**Proposed Solution**
Wait until all results arrive - exit worker then

### LINKED ISSUE #1702: Persist and recover committor intent execution stages
## Goal

A restart shouldn't mean losing pending intents or repeating transactions that already succeeded. Record execution progress in outbox accounts so the committor can pick up where it left off.

## Why

The tricky part is knowing what happened before the restart. An intent might not have started, a transaction might still be pending, or the commit might have succeeded while finalization is unfinished.

A confirmation timeout doesn't tell us which happened. Before retrying, check the previous transaction's status so we don't submit the same work twice.

## Scope

Replace the committor persister with outbox-backed recovery, following [MIMD-0025](https://github.com/magicblock-labs/magicblock-validator/discussions/1246) and [MIMD-0027](https://github.com/magicblock-labs/magicblock-validator/discussions/1416).

Keep enough information to resume each stage safely:

- Record accepted intents and execution-stage changes, including pending signatures and blockhashes.
- Recover both single-stage and two-stage execution, checking pending transactions before replacing an attempt.
- Validate stage transitions. If only finalization needs retrying, retain the confirmed commit signature.
- Clean up completed intents so recovery doesn't pick them up again.

Before rollout, we still need to decide how to recover work stored by the old persister. Shutdown coordination is tracked separately in #1262.

taco-paco:
## What changed

Record accepted intents and their execution stages in outbox accounts so a restart doesn't mean starting over. The committor can resume pending work using the transaction signatures and blockhashes retained by stage-specific executors.

This replaces the committor persister with outbox acceptance, stage updates, recovery scans, and cleanup of completed intents, following [MIMD-0025](https://github.com/magicblock-labs/magicblock-validator/discussions/1246) and [MIMD-0027](https://github.com/magicblock-labs/magicblock-validator/discussions/1416).

Closes #1702

## Impact

The persistence format changes. Before rollout, we still need to decide how to recover work stored by the old persister. Shutdown coordination is tracked separately in #1262.

## Reviewer notes

A transaction may have succeeded even if confirmation timed out. Check how stage persistence is ordered relative to submission, and how recovery checks the previous signature before replacing an attempt.

If commit succeeded but finalization didn't finish, the retry must keep the confirmed commit signature. Stage transitions must remain valid, and completed outboxes must be cleaned up so recovery doesn't pick them up again.


magicblock-committor-service/src/intent_engine/intent_channel.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub fn schedule(
+        &self,
+        intent_bundles: Vec<OutboxIntentBundle>,
+    ) -> Result<(), IntentScheduleError> {
+        // If db not empty push el-t there
+        // This means that at some point channel got full
+        // Worker first will clean-up channel, and then DB.
+        // Pushing into channel would break order of commits
+        // Lock shall be held across to avoid races
+        let db = self.db.lock().expect(POISONED_MSG);
```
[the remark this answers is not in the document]
taco-paco: Out of scope

test-integration/test-schedule-intent/tests/test_outbox_flow.rs
```diff
@@ -0,0 +1,1152 @@
...
+    // Simulates shutdown/crash after successful execution
+    // Validator could sent tx and then crash
+    // On restart validator will extract outbox bundles
+    let outbox_bundle = test_env
+        .outbox_client()
+        .outbox_reader()
+        .fetch_outbox_intent(outbox_bundle.inner.id)
+        .await
+        .expect("fetch failed")
+        .expect("outbox bundle not found");
```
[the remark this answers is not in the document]
taco-paco: > hat cleanup step can hide the very post-execution transition this PR is adding

What for example?

test-integration/test-schedule-intent/tests/common/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+        .unwrap();
+
+    let ix = dlp_api::instruction_builder::top_up_ephemeral_balance(
+        payer.pubkey(),
+        payer.pubkey(),
+        Some(LAMPORTS_PER_SOL / 2),
+        Some(1),
+    );
+    ctx.send_and_confirm_instructions_with_payer_chain(&[ix], &payer)
+        .unwrap();
```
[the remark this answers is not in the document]
taco-paco: That's fine in test and it will fail anyway

magicblock-committor-service/src/intent_engine.rs
```diff
@@ -0,0 +1,87 @@
+pub mod db;
+pub mod intent_channerl;
```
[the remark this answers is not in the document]
taco-paco: Fixed

magicblock-committor-service/src/intent_engine/db.rs
```diff
@@ -0,0 +1,141 @@
...
+        drop(queue);
+
+        let intent_pda = outbox_intent_pda(id);
+        let account = self
+            .accounts_db
+            .get_account(&intent_pda)
+            .ok_or(Error::IntentNotFoundError(id))?;
+
+        let outbox_intent = OutboxIntentBundle::try_from_bytes(account.data())?;
+        Ok(Some(outbox_intent))
```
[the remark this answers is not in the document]
taco-paco: That is fine, keeping it in is also incorrect, from the fact that we kept it in it won't become valid all of a sudden. The most correct approach is to return popped data in Error but this is not needed as of right now

magicblock-committor-service/src/intent_executor/cleanup_handle.rs
```diff
@@ -0,0 +1,49 @@
...
+        let cleanup_futs = self.junk.iter().map(|to_cleanup| {
+            self.transaction_preparator.cleanup_for_strategy(
+                &self.authority,
+                &to_cleanup.optimized_tasks,
+                &to_cleanup.lookup_tables_keys,
+                close_buffers,
+            )
+        });
+
+        try_join_all(cleanup_futs).await.map(|_| ())
```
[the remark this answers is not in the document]
taco-paco: Done

magicblock-committor-service/src/intent_engine/intent_channel.rs
```diff
@@ -0,0 +1,126 @@
...
+        let res = iter.try_for_each(|el| self.sender.try_send(el));
+        match res {
+            Ok(_) => Ok(()),
+            Err(TrySendError::Closed(_)) => {
+                Err(IntentScheduleError::ChannelClosed)
+            }
+            Err(TrySendError::Full(el)) => {
+                let leftovers = std::iter::once(el).chain(iter).collect();
+                db.store_intent_bundles(leftovers)
+                    .map_err(IntentScheduleError::from)
```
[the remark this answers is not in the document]
taco-paco: Fixed

magicblock-committor-service/src/intent_executor/single_stage_intent_executor.rs
```diff
@@ -0,0 +1,178 @@
...
+        let flow = check_pending_signature(
+            &self.ctx.intent_client,
+            &self.pending_signature,
+        )
+        .await?;
+        match flow {
+            // Intent was executed - return
+            ControlFlow::Break(()) => {
+                Ok(ExecutionOutput::SingleStage(self.pending_signature))
+            }
```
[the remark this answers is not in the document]
taco-paco: Fixed

magicblock-committor-service/src/intent_executor/strategy_executor/utils.rs
```diff
@@ -0,0 +1,585 @@
...
+            el,
+            BaseTaskImpl::Commit(_) | BaseTaskImpl::CommitFinalize(_)
+        )
+    });
+    let (mut commit_stage_tasks, mut finalize_stage_tasks) = (vec![], vec![]);
+    for (i, el) in strategy.optimized_tasks.into_iter().enumerate() {
+        if Some(i) <= last_commit_ind {
+            commit_stage_tasks.push(el);
+        } else {
+            finalize_stage_tasks.push(el);
```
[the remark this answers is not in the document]
taco-paco: This is an expected bahaviour, CommitFinalize is faster task that is executed in Commit stage

magicblock-committor-service/src/intent_executor/strategy_executor/utils.rs
```diff
@@ -0,0 +1,585 @@
...
+            }
+        } else {
+            // Already timed out; see comment above.
+            executor.execute_callbacks(None, Err(ActionError::TimeoutError));
+        }
+    }
+
+    // Timeout concerns only callbacks
+    // We still need to drive intent to completion
+    executor.execute().await
```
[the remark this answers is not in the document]
taco-paco: If tx was already sent `ExecutionState::pending_signature` will be set to that tx signature, as result we will query rpc and get a tx result

magicblock-committor-service/src/intent_executor/strategy_executor/patcher.rs
```diff
@@ -0,0 +1,416 @@
...
+    async fn patch(
+        &mut self,
+        err: &TransactionStrategyExecutionError,
+        strategy: &mut TransactionStrategy,
+        report: &mut IntentExecutionReport,
+    ) -> IntentExecutorResult<ControlFlow<(), TransactionStrategy>> {
+        if self.committed_pubkeys.is_empty() {
+            // No patching is applicable if intent doesn't commit accounts
+            return Ok(ControlFlow::Break(()));
+        }
```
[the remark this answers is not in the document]
taco-paco: If Intent only with actions fails we shouldnt patch it

magicblock-committor-service/src/intent_executor/two_stage_intent_executor.rs
```diff
@@ -0,0 +1,259 @@
...
+        let pubkeys = intent.get_all_committed_pubkeys();
+
+        let mut execution_report = IntentExecutionReport::default();
+        let result = self.execute_inner(intent, &mut execution_report).await;
+
+        if is_undelegate {
+            self.ctx
+                .task_info_fetcher
+                .reset(ResetType::Specific(&pubkeys));
+        }
```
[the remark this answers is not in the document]
taco-paco: Fixed

magicblock-committor-service/src/outbox/outbox_client.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        Ok(())
+    }
+
+    fn outbox_reader(&self) -> Self::OutboxReader {
+        const CAPACITY: NonZeroUsize = NonZeroUsize::new(1000).unwrap();
+        InternalOutboxIntentBundlesReader::new(
+            self.accounts_db.clone(),
+            CAPACITY,
+        )
```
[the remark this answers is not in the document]
taco-paco: This is compiler time error

magicblock-committor-service/src/service.rs
```diff
@@ -178,29 +173,45 @@ where
...
-            })?;
-        if bundles.is_empty() {
-            return Ok(());
-        }
+    async fn reschedule_intents(
+        &self,
+    ) -> Result<(), IntentExecutionServiceError> {
+        /// Number of intents rescheduled at once
+        const RESCHEDULE_CHUNK_SIZE: NonZeroUsize =
+            NonZeroUsize::new(1000).unwrap();
```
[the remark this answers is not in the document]
taco-paco: Compile-time error - fine. Actually good

magicblock-committor-service/src/utils.rs
```diff
@@ -1,40 +1,25 @@
...
-            update_status.clone(),
-        ) {
-            error!(pubkey = %pubkey, error = ?err, "Failed to persist status");
-        }
-    });
+pub fn serialized_transaction_size(
+    transaction: &impl SerializableTransaction,
+) -> usize {
+    // SAFETY: runs on transactions we already serialize before sending.
+    usize::try_from(bincode::serialized_size(transaction).unwrap()).unwrap()
```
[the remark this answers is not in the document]
taco-paco: Out of scope. Just a copy-pasted code

magicblock-metrics/src/metrics/mod.rs
```diff
@@ -419,6 +419,14 @@ lazy_static::lazy_static! {
         "Number of signatures included in batched getSignatureStatuses requests"
     ).unwrap();
 
+    static ref RPC_CLIENT_SIGNATURE_HISTORY_DURATION_SECONDS: Histogram = Histogram::with_opts(
+        HistogramOpts::new(
+            "rpc_client_signature_history_duration_seconds",
+            "Time in seconds spent on getSignatureStatuses-with-history (restart recovery)"
+        )
+        .buckets(vec![0.05, 0.1, 0.25, 0.5, 1.0, 2.0, 5.0, 10.0])
+    ).unwrap();
```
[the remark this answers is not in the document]
taco-paco: That is fine as fails right on start

mimd-0025.md
```diff
@@ -0,0 +1,293 @@
...
+
+Deallocates the `OutboxIntentBundle` ephemeral account. Called after L1 execution is confirmed.
+
+**Accounts:**
+```
+[validator_auth (signer), closing_pda (writable)]
+```
+
+**Logic:**
+1. Wipe and deallocate `closing_pda`
```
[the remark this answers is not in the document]
taco-paco: Outdated, removed

magicblock-committor-service/src/intent_executor/two_stage_intent_executor.rs
```diff
@@ -0,0 +1,259 @@
...
+        )
+        .await?;
+
+        let finalized_stage =
+            finalize_strategy_executor.done(finalize_signature);
+        Ok(ExecutionOutput::TwoStage {
+            commit_signature: finalized_stage.commit_signature,
+            finalize_signature: finalized_stage.finalize_signature,
+        })
+    }
```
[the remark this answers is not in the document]
taco-paco: Fixed

magicblock-committor-service/src/outbox/outbox_intent_bundles_reader.rs
```diff
@@ -0,0 +1,327 @@
...
+        // SAFETY: OrderedIntent is #[repr(transparent)] over OutboxIntentBundle —
+        // identical memory layout, zero-copy cast
+        let mut items: Vec<OutboxIntentBundle> = unsafe {
+            let mut v = std::mem::ManuallyDrop::new(heap.into_vec());
+            Vec::from_raw_parts(
+                v.as_mut_ptr() as *mut OutboxIntentBundle,
+                v.len(),
+                v.capacity(),
+            )
+        };
```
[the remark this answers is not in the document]
taco-paco: That adds unnecessary overhead on allocation of new vector + moving. No need to do it if the code above fundamentally safe

mimd-0025.md
```diff
@@ -0,0 +1,293 @@
...
+
+Additionally, just before sending a transaction to L1, the validator records the transaction signature back into the `OutboxIntentBundle` via `SetIntentExecutionStage`. This lets the restart logic distinguish "prepared but not yet sent" from "sent — check the signature on L1."
+
+**No stage is executed twice.** Because `SetIntentExecutionStage` is committed to `AccountsDB` before the L1 transaction is submitted, any restart that finds `status = Executing(...)` will always query L1 for the stored signature first. A confirmed & successful tx signature advances or closes the intent. The write-before-send ordering is the invariant that makes this hold.
+
```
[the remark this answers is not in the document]
taco-paco: Valid, addressed

programs/magicblock/src/intent_bundles/outbox/process_scheduled_commit_sent.rs
```diff
@@ -128,19 +129,73 @@ pub fn process_scheduled_commit_sent(
...
+        .error_message
+        .map(|_| Err(InstructionError::Custom(INTENT_FAILED_CODE)))
+        .unwrap_or(Ok(()))?;
+
+    // Close Outbox intent
+    close_outbox_account_cpi(
+        invoke_context,
+        validator_authority_id,
+        expected_pda,
+    )
```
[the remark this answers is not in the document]
taco-paco: Out of scope, just moved code

magicblock-magic-program-api/src/instruction.rs
```diff
@@ -57,27 +58,31 @@ pub enum MagicBlockInstruction {
...
     ///
-    /// We implement it this way so we can log the signature of this transaction
-    /// as part of the [MagicBlockInstruction::ScheduleCommit] instruction.
-    /// Args: (intent_id, bump) - bump is needed in order to guarantee unique transactions
-    ScheduledCommitSent((u64, u64)),
+    /// # Account references
+    /// - **0.** `[WRITE, SIGNER]` Validator Authority (receives rent refund from close)
+    /// - **1.** `[WRITE]`         Outbox intent PDA to close, seeds: `["outbox-intent", intent_id.to_le_bytes()]`
+    /// - **2.** `[WRITE]`         Ephemeral vault
+    ScheduledCommitSent(u64),
```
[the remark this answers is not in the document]
taco-paco: Done

test-integration/test-tools/src/transactions.rs
```diff
@@ -57,6 +57,7 @@ pub fn send_transaction(
 ) -> Result<Signature, client_error::Error> {
     let blockhash = rpc_client.get_latest_blockhash()?;
     tx.try_sign(signers, blockhash)?;
+    println!("trtr: {}", tx.signatures[0]);
```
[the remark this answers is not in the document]
taco-paco: Removed

magicblock-committor-service/src/intent_engine/intent_channel.rs
```diff
@@ -0,0 +1,131 @@
...
+                Poll::Ready(item.map(Ok))
+            } else {
+                // Some(T) always will be returned here as per check above
+                let el = db.pop_intent_bundle();
+                Poll::Ready(el.transpose())
+            }
+        } else {
+            let item = ready!(this.stream.poll_next(cx));
+            Poll::Ready(item.map(Ok))
+        }
```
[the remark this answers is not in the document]
taco-paco: Fixed

programs/magicblock/src/intent_bundles/outbox/process_accept_scheduled_commits.rs
```diff
@@ -0,0 +1,259 @@
...
+    let Some(num_accept_intents) =
+        num_ix_accounts.checked_sub(INTENT_PDAS_OFFSET as usize)
+    else {
+        ic_msg!(
+            invoke_context,
+            "AcceptScheduledCommits ERR: not enough accounts to accept intents ({}), need validator authority, magic context, vault, and at least one outbox intent PDA",
+            num_ix_accounts
+        );
+        return Err(InstructionError::MissingAccount);
+    };
```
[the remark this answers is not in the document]
taco-paco: Fixed

programs/magicblock/src/magic_context.rs
```diff
@@ -53,26 +53,46 @@ impl MagicContext {
...
         };
+
         let mut len = [0; mem::size_of::<u64>()];
         len.copy_from_slice(raw_len);
-        u64::from_le_bytes(len) != 0
+        u64::from_le_bytes(len)
+    }
+
+    pub fn has_scheduled_intents(data: &[u8]) -> bool {
+        Self::scheduled_intents_len(data) != 0
```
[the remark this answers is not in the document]
taco-paco: Fixed

magicblock-committor-service/src/outbox/outbox_client.rs
```diff
@@ -0,0 +1,229 @@
...
+    ) -> Result<
+        Vec<ScheduledIntentBundle>,
+        (Vec<ScheduledIntentBundle>, Self::Error),
+    > {
+        // If accounts were scheduled to be committed, we accept them here
+        // and processs the commits
+        let magic_context_acc =
+            self.safe_get_account(&MAGIC_CONTEXT_PUBKEY).expect(
+                "Validator found to be running without MagicContext account!",
+            );
```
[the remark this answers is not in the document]
taco-paco: Migration of old code

snawaz:
image

Wow, that's a lot of changes. Does this PR also include **heavy refactoring**?

Should we strive for smaller PRs so reviews can be done more thoroughly?

taco-paco:
> Wow, that's a lot of changes. Does this PR also include heavy refactoring?
> Should we strive for smaller PRs so reviews can be done more thoroughly?

It does include refactoring. A lot of code that would be otherwise duplicated in different IntentExecutors was extracted into utils. Patching logic was extracted into `patcher.rs`. Most of them don't have any business logic changes.

The main logic shall be easily reviewable because of centralization in single general functions: `stage_execution_loop`, `execute_single_stage_flow`,  `execute_two_stage_flow`.

I agree that would be better to split it off, but it wasn't really planned and is something that happened in the middle of writing code so it would be hard to extract those changes into separate PRs. Will try to forsee this next time

magicblock-committor-service/src/service.rs
```diff
@@ -158,17 +157,15 @@ where
                 }
                 _ = interval.tick() => {
                     let accept_result = self
```
thlorenz, line 100: This worker accepts scheduled intents without checking that the validator is in primary mode. 

Since `magicblock-api` still starts this service during normal startup and the new `AcceptScheduleCommits`/`SetIntentExecutionStage` handlers do not guard on `CoordinationMode::Primary`, replicas can mutate MagicContext/outbox state and start base-layer intent execution even though `CoordinationMode::Replica` is documented as having no side effects. 

Can we gate this accept/schedule loop on primary mode, or otherwise keep the service disabled for replicas?

programs/magicblock/src/intent_bundles/outbox/process_scheduled_commit_sent.rs
```diff
@@ -128,19 +129,73 @@ pub fn process_scheduled_commit_sent(
...
+            );
+            return Err(InstructionError::Custom(
+                custom_error_codes::UNABLE_TO_UNLOCK_SENT_COMMITS,
+            ));
+        }
+    };
+
+    // Log data
+    log_sent_commit(invoke_context, &commit);
+    commit
```
thlorenz: For failed intent execution this returns before closing the outbox PDA, but the code has already removed the entry from `SENT_COMMITS`. 

The committor reports both successes and failures through this path, so a failed intent leaves the durable outbox account open while dropping the local failure report. 
After restart it can be rediscovered and retried instead of staying terminally failed. 

Can we record the failure and close/consume the outbox atomically, or close first and then surface a durable failure status?

magicblock-committor-service/src/service.rs
```diff
@@ -158,17 +157,15 @@ where
...
-                            error!("Failed to accept intents: {}", err);
-                            continue;
-                        }
-                    };
+                    let intent_bundles = accept_result.unwrap_or_else(|(accepted_intents, err)| {
+                        error!("Failed to accept intents: {}", err);
+                        accepted_intents
+                    });
 
+                    let intent_bundles= intent_bundles.into_iter().map(OutboxIntentBundle::accepted).collect();
```
thlorenz: With the new durable outbox flow, accepted outbox bundles can be fed back into the execution scheduler from both fresh accepts and recovery/rescheduling. The scheduler still only tracks executing intents through committed pubkeys, so standalone action bundles, where `get_all_committed_pubkeys()` is empty, are never recorded as executing and duplicate intent IDs are not caught while the first action is in flight. 

Since standalone actions have external base-layer side effects, can we track executing intent IDs independently of committed pubkeys before scheduling these accepted outbox bundles?

thlorenz (changes requested):
I found several lifecycle issues that look blocking before merge and commented in place except for the below two:

#### Outside-diff 1

**Location:** `magicblock-committor-service/src/intent_engine/intent_execution_engine.rs:281-290`

I could not find a good in-diff anchor for this one. The behavior appears to predate the PR, but it is still worth calling out if we want the new durable outbox design to close this lifecycle gap:

The intent scheduler completes and releases committed pubkeys even when `result.inner` is an error. 
In a two-stage flow, the commit stage may already have been sent/recorded while the finalize stage failed. 
Releasing the keys at that point lets later intents touching the same accounts execute before the failed intent is retried, finalized, or marked terminal, which conflicts with per-account intent serialization and recovery safety. 

The engine should keep those account keys blocked or move the intent into a durable terminal/retry state before unblocking dependent intents.

#### Outside-diff 2

**Location:** `magicblock-committor-service/src/tasks/task_strategist.rs:140-156`

I could not find a fair in-diff anchor for this one either. The task-count limit was not introduced by this PR, but it still seems critical for the commit pipeline:

The strategist limits single-stage packing by task count only, while each commit/finalize/undelegate task advertises 120,000 CU and `assemble_tasks_tx_with_uniqueness_nonce` sums those values into one compute-budget limit. A 22-task single-stage transaction can request 2,640,000 CU, and large two-stage commit batches can exceed Solana's 1,400,000 CU transaction cap as well. 

The strategist should enforce the transaction-level CU cap when selecting single-stage and two-stage task groups, or split stages so every produced transaction stays below the cap.

programs/magicblock/src/intent_bundles/outbox/process_scheduled_commit_sent.rs
taco-paco: In the world without network errors we have intents that can't fail. If intent failed due to Action error - we patch it and execute until it succeeds. Other than that Commit, Finalize, Undelegate can't fail, if they do we have a problem but we shouldn't close intent. 
So not closing intent here is a design choice. If say intent failed due to unknown network reason we should not close it but have it reloaded upon restart or retried until it succeeds.

magicblock-committor-service/src/service.rs
taco-paco: > are never recorded as executing and duplicate intent IDs are not caught while the first action is in flight.

What do you mean by "duplicate intent IDs"? there can't be duplicate intent ids as each intent is +1 from previous one. All OutboxIntent are mapped from ScheduledIntents so thier ids can't contain duplicates

taco-paco: Restart doesn't do anything as MagicContext and OutboxItntents are accounts in Accounts DB and new intents will just get +1 intent id, still no duplicates

magicblock-committor-service/src/service.rs
taco-paco: Valid, thanks, fixed 01c6b28b

magicblock-magic-program-api/src/outbox.rs
```diff
@@ -0,0 +1,136 @@
...
+        commit: Signature,
+        finalize: PendingTransaction,
+    },
+}
+
+impl ExecutionStage {
+    pub fn apply_stage_transition(
+        &mut self,
+        stage: ExecutionStage,
+    ) -> Result<(), &'static str> {
```
snawaz: better return Enum error? `str` are hard to match on!

magicblock-magic-program-api/src/outbox.rs
```diff
@@ -0,0 +1,136 @@
...
+            Self::TwoStage(value) => value.pending_transaction(),
+        }
+    }
+}
+
+impl TwoStageProgress {
+    fn apply_stage_transition(
+        &mut self,
+        stage: TwoStageProgress,
+    ) -> Result<(), &'static str> {
```
snawaz: better return Enum error? `str` are hard to match on!

programs/magicblock/src/intent_bundles/outbox/process_accept_scheduled_commits.rs
```diff
@@ -0,0 +1,266 @@
...
+                AccountMeta::new(sponsor, true),
+                AccountMeta::new(pda, true),
+                AccountMeta::new(EPHEMERAL_VAULT_PUBKEY, false),
+            ],
+            data: MagicBlockInstruction::CreateEphemeralAccount { data_len }
+                .try_to_vec()
+                .map_err(|_| InstructionError::InvalidInstructionData)?,
+        },
+        &[pda],
+    )
```
snawaz: perf: Can we avoid the self-CPI here (in a follow-up PR)?

Since this is invoking the same program, it may be cheaper and simpler to refactor the account-creation logic into a shared helper (say, **`create_ephemeral_account_unchecked`**) and call it directly from both paths. A shared-helper would avoid lots of overhead associated with CPI (instruction construction, serialization, account meta allocation, account remapping, signer/writable privilege checks, stack push/pop, and dispatch back through the processor).

It's especially important because `AcceptScheduleCommits` batches up to 50 intents, this overhead scales directly with the `CHUNK_SIZE` (currently set to `50` in `send_accept_tx()`) and `RESCHEDULE_CHUNK_SIZE` is set to `1000` !!

magicblock-committor-service/src/service.rs
```diff
@@ -179,29 +191,50 @@ where
...
+        loop {
+            // Read by chunks in order not to overload `IntentExecutionEngine`
+            let mut intent_bundles_chunk = outbox_bundles_reader
+                .read(RESCHEDULE_CHUNK_SIZE.get())
+                .await
+                .map_err(Into::into)?;
+            // Original blockhash is stale after restart; signal recovery so
+            // notify_commit_sent rebuilds the tx with a fresh ER blockhash.
+            intent_bundles_chunk.iter_mut().for_each(|b| {
+                b.inner.sent_transaction = Transaction::default()
```
snawaz, line 143: type-safety: I think we should change the type of `sent_transaction` to `               IntentSentTransaction`, so this recovery state is explicit instead of implicitly-encoded as an empty/default transaction:
```suggestion
                b.inner.sent_transaction = IntentSentTransaction::Recovered;
```

Or at least `Option<Transaction>` (less explicit, but still better than empty-transaction) if you do not want to move the definition of `IntentSentTransaction` to the crate where bundle is defined.

magicblock-committor-service/src/outbox/outbox_intent_bundles_reader.rs
snawaz, line 126: I agree with Rabbit.

The `unsafe` block isn't worth having just to avoid one memory allocation. We're not saving much, given that we're already paying for DB scanning, deserialization of all scanned items (including paying for their nested allocations), heap maintenance, sorting. 

So we should prefer the safe version here **unless** we have a benchmark showing this conversion matters.

```suggestion
  let mut items: Vec<OutboxIntentBundle> =
      heap.into_iter().map(|ordered| ordered.inner).collect();
```

snawaz: perf/simplify: When zooming out, I think the current bounded-heap approach has two downsides:

- The caller asks for `n` items in a loop (from `reschedule_intents()`), so we scan the DB, serialize all items, refill the buffer and discards the extra items (`> capacity`), and then in the next iteration, we rescan the previously discarded items, and so on... and then repeat this again and again.

- It's quite verbose. Ordering wrapper and unsafe conversion adds to that as well.

**Alternative (simpler and performant):** 

We should replace `refill()` with `populate()` that populates the buffer with all _recovered_ items:

```rust
fn populate(&mut self) {
   // making `buffer: Vec` avoids another conversion!
   self.buffer = outbox_iter.collect::<Vec<_>>();

   // NOTE: sort in "descending" order, not "ascending"
   // Also, rename buffer -> desc_sorted_items
   self.buffer.sort_unstable_by_key(|b|  Reverse(b.id));
}
```

**Rationale:** 

Practically, there will not be lots of "recovered/persisted" intents. If we assume the number is **2000 recovered intents** and each intent has **3 accounts** (on an average), then we need `~2 MB` to store everything in the `self.buffer`. Even if it is `~10 MB` (or even more), that seems acceptable for startup recovery, and we avoid repeated DB scans/deserialization.

We should also consider storing a smaller value in the DB, not the full intent with the commit accounts. (can be a follow-up PR).

magicblock-committor-service/src/outbox/outbox_intent_bundles_reader.rs
```diff
@@ -0,0 +1,327 @@
...
+        }
+
+        // Refill if buffer runs low
+        if self.buffer.len() < n {
+            self.refill()?;
+        }
+
+        // Take available amount
+        let take = n.min(self.buffer.len());
+        Ok(self.buffer.drain(..take).collect())
```
snawaz, line 171: perf: draining from the front is expensive, as it requires the remaining items to shift left. 

If we sort `self.buffer` in descending order, we can drain from the end, which is a lot cheaper.

magicblock-committor-service/src/outbox/outbox_client.rs
```diff
@@ -0,0 +1,254 @@
...
+            let chunk_size = CHUNK_SIZE.min(remaining.len());
+            let tx = InstructionUtils::accept_scheduled_commits(
+                self.latest_block_provider.blockhash(),
+                remaining[..chunk_size].iter().map(|i| i.id),
+            );
+            let backoff_config = ExponentialBackoff {
+                max_elapsed_time: Some(Duration::from_secs(25)),
+                max_interval: Duration::from_secs(5),
+                ..ExponentialBackoff::default()
+            };
```
snawaz, line 110: nit: `backoff_config` can be moved out of the loop and reuse it in the loop.

magicblock-committor-service/src/outbox/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+            included_pubkeys: intent.get_all_committed_pubkeys(),
+            intent_sent_transaction: if intent
+                .sent_transaction
+                .signatures
+                .is_empty()
+            {
+                IntentSentTransaction::Recovered
+            } else {
+                IntentSentTransaction::Known(intent.sent_transaction.clone())
+            },
```
snawaz, line 90: type-safety: Better make `sent_transaction: IntentSentTransaction` or `sent_transaction: Option<Transaction>`. 

(see the other comment on the related thing).

magicblock-committor-service/src/service.rs
```diff
@@ -183,54 +191,50 @@ where
...
+            });
+            if intent_bundles_chunk.is_empty() {
+                return Ok(());
+            }
```
snawaz: nit: early-return can be moved earlier.

magicblock-committor-service/src/intent_engine/db.rs
```diff
@@ -0,0 +1,141 @@
+use std::{cell::RefCell, collections::VecDeque, sync::Arc};
+
+/// DB for storing intents that overflow committor channel
+use magicblock_accounts_db::{traits::AccountsBank, AccountsDb};
+use magicblock_core::intent::outbox::outbox_intent_pda;
+use magicblock_metrics::metrics;
+use magicblock_program::outbox_intent_bundles::OutboxIntentBundle;
+use solana_account::ReadableAccount;
+
+pub trait BacklogDB: Send + 'static {
```
snawaz, line 13: nit/docs: Can we add a short (1-2 lines) doc comment explaining what this backlog is for?

something along this line (my understanding):

> When the in-memory execution channel is full, additional intents are queued in `BacklogDB` while preserving ordering.

If my understanding is correct, then I've a question for `DummyIntentBacklog`. Please see next.

magicblock-committor-service/src/intent_engine/db.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    /// Returns the oldest (first stored) intent bundle
+    fn pop_intent_bundle(&self) -> DBResult<Option<OutboxIntentBundle>>;
+    fn is_empty(&self) -> bool;
+}
+
+pub struct DummyIntentBacklog {
+    accounts_db: Arc<AccountsDb>,
+    queue: RefCell<VecDeque<u64>>,
+}
```
snawaz: rename: `DummyIntentBacklog` sounds test-only, but it is used in production (`magic_validator.rs` and `magic_sys_adator.rs`).

```suggestion
pub struct AccountsDbIntentBacklog {
    accounts_db: Arc<AccountsDb>,
    queue: RefCell<VecDeque<u64>>,
}
```

magicblock-committor-service/src/intent_engine.rs
```diff
@@ -0,0 +1,87 @@
...
+    ) -> Result<(), IntentScheduleError> {
+        self.intent_schedule_handle.schedule(intent_bundles)
+    }
+
+    /// Creates a subscription for results of BaseIntent execution
+    pub fn subscribe_for_results(
+        &self,
+    ) -> broadcast::Receiver<BroadcastedIntentExecutionResult> {
+        self.result_subscriber.subscribe()
+    }
```
snawaz, line 90: readability: 

Could we merge `IntentScheduleHandle` into `IntentEngineHandle`?

We have multiple small types with very similar responsibilities, and `IntentEngineHandle::schedule()` mostly just forwards the call. That adds cognitive load because the reader has to jump through another type/layer without that layer solving a distinct and additional problem.

In fact, `IntentEngineHandle` should eventually merge into `CommittorProcessor` too. Currently, it seems nobody wants to do the actual work and instead keeps forwarding the calls to the next noun, until we run out of nouns: 

```
processor -> engine -> scheduler
```

Merging should simplify the code, particularly because even after the merge, the number of **fields** and **functions** will stay about the same in `CommittorProcessor`; only those forwarding functions will just be a few lines bigger. **But it will be easier to read, as it does not require jumping through multiple layers!**

magicblock-committor-service/src/intent_executor/strategy_executor/utils.rs
```diff
@@ -0,0 +1,800 @@
...
+        // Not found on-chain: could be genuinely never sent, or simply not
+        // landed yet.
+        None | Some(None) => {
+            if blockhash_still_valid {
+                Ok(PendingSignatureStatus::Pending)
+            } else {
+                info!(pending_signature = ?pending.signature, "Pending transaction's blockhash expired without landing");
+                Ok(PendingSignatureStatus::Rejected)
+            }
+        }
```
snawaz: bug: it's very subtle though.

We cannot combine `None | Some(None)`. The code in this branch is **partially** meaningful for `Some(None)` only, which means:

- the tx has not reached the requested commitment (which is currently `Finalized`) level yet.
  - That means "found but not finalized yet" and so the brach handling is partially correct.
- OR it was not received by the node (still in _transit_).
  - **We do not seem to handle this case**. It depends how quickly we ask for the statuses. 🤔 
- OR we never sent it at all.

However, `None` means the RPC node is misbehaving/buggy and isn't doing what [the doc][1] says. We should log `error!` (or maybe `panic!`) and create an alert. If we want to return non-Err, it should be `Pending`, not `Rejected` (which might lead to double-execute, double-spending etc).

[1]: https://solana.com/docs/rpc/http/getsignaturestatuses

magicblock-committor-service/src/intent_executor/strategy_executor/utils.rs
```diff
@@ -0,0 +1,800 @@
...
+) -> IntentExecutorResult<bool> {
+    loop {
+        match check_pending_signature(client, pending).await? {
+            PendingSignatureStatus::Succeeded => return Ok(true),
+            PendingSignatureStatus::Rejected => return Ok(false),
+            PendingSignatureStatus::Pending => {
+                sleep(PENDING_SIGNATURE_POLL_INTERVAL).await;
+            }
+        }
+    }
```
snawaz, line 282: concern: it's better use some loop-size, else it could loop infinitely in some unforeseen scenario.

magicblock-committor-service/src/intent_executor/strategy_executor/utils.rs
```diff
@@ -0,0 +1,800 @@
...
+) -> IntentExecutorResult<PendingSignatureStatus> {
+    let blockhash_still_valid = client
+        .is_blockhash_valid(&pending.blockhash)
+        .await
+        .map_err(IntentExecutorError::GetPendingSignatureStatusError)?;
+
+    let statuses = client
+        .get_signature_statuses_with_history(
+            std::slice::from_ref(&pending.signature),
+            CommitmentConfig::finalized(),
```
snawaz, line 211: bug-risk: do we really need `CommitmentConfig::finalized()` here?

Passing `finalized()` forces `get_signature_statuses_with_history()` convert a real status at a lower commitment level into `None`, which is discussed in the other comment where we lose the ability to distinguish:

- tx not found at all. (this is real `None`).
- tx found, but not finalized yet. (it is converted into `None`).

Those are very different cases for the **exact-once** guarantee. I think we should fetch the status as-is and interpret the commitment level here, in this function, together with the blockhash validity!

---

Having said that, I realize `get_signature_statuses_with_history` is **incorrectly named** because it does more than what the name says. 

**A facade API should match the semantics of the underlying API**. If we want commitment filtering, that should either be explicit in the name... or done by the caller.

magicblock-committor-service/src/intent_executor/intent_execution_client.rs
```diff
@@ -125,6 +74,102 @@ impl IntentExecutionClient {
...
+            .map_err(MagicBlockRpcClientError::from)?;
+        Ok(response
+            .value
+            .into_iter()
+            .map(|status| {
+                status.and_then(|status| {
+                    status
+                        .satisfies_commitment(commitment_config)
+                        .then(|| status.err.map_or(Ok(()), Err))
+                })
```
snawaz, line 142: design: if we want commitment filtering, then it should either be explicit in the name, or choose an entirely different name for the function, because a function with name `get_signature_statuses_with_history` should match the corresponding Solana RPC method contract, else we will end up having a well-known name to mean two different things in two different places.

programs/magicblock/src/intent_bundles/outbox_intent_bundles.rs
```diff
@@ -0,0 +1,148 @@
...
+use solana_hash::Hash;
+use solana_signature::Signature;
+
+use crate::magic_scheduled_base_intent::ScheduledIntentBundle;
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct OutboxIntentBundle {
+    pub inner: ScheduledIntentBundle,
+    status: OutboxIntentBundleStatus,
+}
```
snawaz, line 21: We should also store `bump` to enable quick validation in the subsequent calls.

programs/magicblock/src/intent_bundles/outbox/process_set_intent_execution_stage.rs
```diff
@@ -0,0 +1,128 @@
...
+            "SetIntentExecutionStage ERR: validator authority {} not in signers",
+            validator_auth
+        );
+        return Err(InstructionError::MissingRequiredSignature);
+    }
+
+    // Validate pda we about to apply transition to
+    let provided_pda =
+        get_instruction_pubkey_with_idx(transaction_context, INTENT_PDA_IDX)?;
+    let expected_pda = outbox_intent_pda(intent_id);
```
snawaz: If we store bump, we could use that to quickly validate the PDA.

programs/magicblock/src/intent_bundles/outbox/process_close_outbox_intent.rs
```diff
@@ -0,0 +1,403 @@
...
+                AccountMeta::new(sponsor, true),
+                AccountMeta::new(pda, false),
+                AccountMeta::new(EPHEMERAL_VAULT_PUBKEY, false),
+            ],
+            data: MagicBlockInstruction::CloseEphemeralAccount
+                .try_to_vec()
+                .map_err(|_| InstructionError::InvalidInstructionData)?,
+        },
+        &[],
+    )
```
snawaz: We could avoid the CPI call?

Please see the other comment on using `CreateEphemeralAccount` in CPI.

snawaz (changes requested):
Thanks for the huge work.

Overall direction looks good to me but this PR mixes large refactors with new behavior (we should avoid doing that in the future), so the real correctness/readability diff is hard to review.

I tried to review it as thoroughly as possible within a reasonable amount of time, and posted comments ranging from simplifying code with fewer indirections/layers to correctness concerns.

programs/magicblock/src/intent_bundles/outbox/process_accept_scheduled_commits.rs
taco-paco: Unfortunately I don't think this would be possible with new engine. ER accounts can't be created in magic-program anymore. We HAVE to CPI into special builtin magic_root as only it has a permission to create such accounts.

magicblock-committor-service/src/outbox/outbox_intent_bundles_reader.rs
taco-paco: The number of loaded intents are unknown. We had a case when 200k were loaded on restart.
Additionally, size of accounts is also unknown. Having a system that blindly loads the whole thing is incorrect IMO. 

With regards to performance under CAPACITY, which as you said will be the most likely case, the strict algorithmic complexity is the same O(n*logn) same as provided snippet. But in an unlikely scenario we load an unbounded amount of intents which is imo incorrect. Tho current system can be improved of course.

> We should also consider storing a smaller value in the DB  not the full intent with the commit accounts. (can be a follow-up PR).

By DB you mean AccountsDB? If so, this is rather complex to achieve as of now. We need to commit state at the moment of execution. Fetching state by pubkey during commit after(exection) would yield a potentially newer state. I could see that being possible only via replay from snapshot to a particular commit slot, which feels rather expensive as well and not sure it will be better, tradeoff here again :)

What I think would be better - charge user extra for large Intent. As intents are supposed to be closed this would be our compensation for inconvinience. 

If you mean other DB.
One of the reasons for Outbox was moving away from DB. The reason is that supplied state from db has also to become part of the whole fraud prood story. Replicas have to have it as well and so on.

magicblock-committor-service/src/intent_executor/strategy_executor/utils.rs
taco-paco: Valid thanks, returning Pending would be correct in that case. Will address

programs/magicblock/src/intent_bundles/outbox/process_close_outbox_intent.rs
taco-paco: Yes, we do now

magicblock-committor-service/src/service.rs
taco-paco: This is a breaking change for MagicContext, additionally within tx flow having enum `IntentSentTransaction` doesn't make a lot of sense. It makes sense further in committor service where retry logic is introduced. magic-program shouldn't be aware of it via some types

magicblock-committor-service/src/outbox/outbox_client.rs
taco-paco: truly nit, not sure i see a need in this, also makes us call clone instead all the time on `backoff_config`

magicblock-committor-service/src/outbox/outbox_intent_bundles_reader.rs
taco-paco: This is ring buffer not a vec, draining a prefix (drain(..take)) on a ring buffer just advances the internal head index, it doesn't shift the remaining elements. Same O(1)-per-item cost  .

magicblock-magic-program-api/src/outbox.rs
taco-paco: Valid, fixed

magicblock-magic-program-api/src/outbox.rs
taco-paco: Forgot about those, fixed

magicblock-committor-service/src/service.rs
taco-paco: Fixed

magicblock-committor-service/src/intent_engine/db.rs
taco-paco: Done

magicblock-committor-service/src/outbox/mod.rs
taco-paco: Answered here https://github.com/magicblock-labs/magicblock-validator/pull/1368#discussion_r3892120934

magicblock-committor-service/src/intent_executor/strategy_executor/utils.rs
taco-paco: Addressed for a `None` scenario, it has a limit now. Considered Pending a valid response as if we don't trust it we also may as well not trust status rpc returned

magicblock-committor-service/src/intent_executor/strategy_executor/utils.rs
taco-paco: Addressed for a `None` scenario, it has a limit now. Considered Pending a valid response as if we don't trust it we also may as well not trust status rpc returned

magicblock-committor-service/src/intent_engine.rs
taco-paco: I'd suggest refactorings in the future PRs.

> Could we merge IntentScheduleHandle into IntentEngineHandle?
We could, guess engine constructor could return handle that serves as a `Sink` for intents.

CommittorProcessor is pretty useless right now, I only left it in case we would expose some API for users for fetching info about intents.

programs/magicblock/src/intent_bundles/outbox/process_set_intent_execution_stage.rs
taco-paco: Done

programs/magicblock/src/intent_bundles/outbox_intent_bundles.rs
taco-paco: Done

magicblock-committor-service/src/intent_engine/db.rs
taco-paco: Done

PR #1379, opened 2026-07-01
feat: add metrics to see how many alts are typically used by intents

taco-paco:
## Summary

Have a suspicion that intent may use a lot of ALTs, adding metric to track

## Breaking Changes
- [ ] None

## Summary by CodeRabbit

* **New Features**
  * Added a new metric that tracks how many address lookup table entries are used per intent transaction.
  * Intent transaction preparation now records this count automatically, improving observability and monitoring.


snawaz (approved):
Looks good to me!

GabrielePicco (approved):

PR #1357, opened 2026-06-22
feat: Schedule undelegation from observed DLP requests

snawaz:
request-undelegation flow: 
   
- subscription-path-flow:  `observes request PDAs` -> `ScheduleCommitAndUndelegate` -> `request-aware undelegation to close the request PDA, in the committor service`.
- ~schedule-commit-flow: now the regular path also sends the additional accounts to the finalize instruction, so that the DLP can detect "undelegation-request" and if so, finalize itself can undelegate the account~.
 
with retries and proper error loggings/handlings at each stage.

PR #1381, opened 2026-07-01
Implement rent-pending ATA materialization

### LINKED ISSUE #1575: Support rent-pending ATA materialization in the validator
## Goal

Allow an ephemeral rollup to create and fund a canonical rent-pending ATA before the corresponding base-layer eATA exists, then materialize it safely during commit.

## Acceptance criteria

- Create and recognize rent-pending ATAs for supported SPL Token and Token-2022 accounts, and reject invalid or empty creations.
- Record the required materialization metadata and charge the delegated payer during scheduling.
- Initialize and delegate the base-layer eATA before DLP commit work, with correct nonce, reimbursement, rollback, and validator-conflict handling.
- Include materialization work in pre-acceptance validation and cover the success and failure paths with tests.

Implementation: #1381

GabrielePicco:
## Summary
- add rent-pending ATA detection using the rent sysvar close-authority sentinel and validate newly created accounts finish execution with a positive token balance
- record materialization metadata during scheduling, charge the delegated payer, and prepend e-token initialize/delegate tasks before DLP commit work
- add coverage for scheduling metadata, payer charge, rollback, committor task ordering, and the delegated-elsewhere conflict path
- update the validator specification and crate notes for the new flow

Closes #1575

## Summary by CodeRabbit

* **New Features**
  * Added rent-pending canonical ATA/ephemeral ATA support via the **CreateRentPendingAta** instruction, including automatic scheduling metadata and eATA materialization during commit flows.
  * Enhanced commit/finalize execution with validator-aware rent-pending nonce/fee handling and rent-pending reimbursement logic.

* **Bug Fixes**
  * Rejects rent-pending ATA creations that end with zero token balance, with rollback enforced.
  * Adds stricter failure gating for existing eATA delegate/validator mismatches before DLP commit/undelegate steps.

* **Documentation**
  * Expanded protocol specs and guides covering rent-pending ATA rules, fees, compatibility, and failure semantics.

* **Tests**
  * Added coverage for rent-pending ATA creation rollback, fee/nonce behavior, and mismatch gating.


magicblock-magic-program-api/src/instruction.rs
```diff
@@ -184,12 +184,18 @@ pub enum MagicBlockInstruction {
...
+    /// # Account references
+    /// - **0.** `[SIGNER]` Payer/sponsor
+    /// - **1.** `[WRITE]`  Canonical ATA PDA
+    /// - **2.** `[]`       Mint
+    /// - **3.** `[]`       Token program
+    CreateRentPendingAta {
+        wallet_owner: Pubkey,
+        mint: Pubkey,
+        token_program: Pubkey,
+    },
```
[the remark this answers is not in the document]
GabrielePicco: This replaces the existing reserved `Unused` slot at discriminant 15; it does not add an ordinal before `CloneAccount`. `CloneAccount` remains 16 and all later discriminants are unchanged. The slot was explicitly reserved for repurposing, so no move is needed.

docs/mimd-rent-pending-ephemeral-ata.md
```diff
@@ -0,0 +1,486 @@
...
+
+```rust
+RENT_PENDING_ATA_CLOSE_AUTHORITY = solana_sysvar::rent::ID;
+```
+
+The rent sysvar is stable, easy to recognize, and not controlled by a normal
+keypair. No validator, Magic Program, or e-token instruction may expose a path
+that treats the rent sysvar as a signing authority for closing a local
+rent-pending ATA. The local account must clear or replace the close authority
+once it is no longer rent-pending.
```
snawaz: I don't get the rationale. Why not use a dedicated authority? That will be more recognizable.

docs/mimd-rent-pending-ephemeral-ata.md
```diff
@@ -0,0 +1,486 @@
...
+
+1. The local account pubkey is the canonical ATA for `(wallet_owner,
+   token_program, mint)`.
+2. The account owner is `TOKEN_PROGRAM_ID` or `TOKEN_2022_PROGRAM_ID`.
+3. Token account data parses successfully for that token program.
+4. Token account `owner` and `mint` match the canonical ATA derivation.
+5. Token account `close_authority` is
+   `Some(RENT_PENDING_ATA_CLOSE_AUTHORITY)`.
+6. Local account flags have `delegated == true`.
+7. Local account flags have `confined == false` and `undelegating == false`.
```
snawaz: I think it's a bad idea to overload the meaning of "delegated" when a corresponding account does not exist on the base. Can we do something else here?

docs/mimd-rent-pending-ephemeral-ata.md
```diff
@@ -0,0 +1,486 @@
...
+8. The account is not a native SOL / wrapped SOL token account.
+9. The base eATA is missing, not delegated, or already delegated to this
+   validator with compatible state. An eATA delegated to another validator is
+   not a valid rent-pending target.
```
snawaz: It seems to be a costly thing to do as it requires attempt-to-fetch-the-account.

magicblock-committor-service/src/intent_executor/task_info_fetcher.rs
```diff
@@ -53,13 +63,67 @@ pub trait TaskInfoFetcher: Send + Sync + 'static {
...
+    /// Fetches rent reimbursement addresses, using caller-provided defaults
+    /// when allowed metadata is missing on base.
+    async fn fetch_rent_reimbursements_with_missing_as(
+        &self,
+        pubkeys: &[Pubkey],
+        min_context_slot: u64,
+        missing_metadata_as: &HashMap<Pubkey, Pubkey>,
+    ) -> TaskInfoFetcherResult<HashMap<Pubkey, Pubkey>> {
+        let reimbursements = self
+            .fetch_rent_reimbursements(pubkeys, min_context_slot)
```
snawaz: Please note that `fetch_rent_reimbursements` is replaced by `fetch_delegation_metadata` in #1357. 

So when we rebase this branch, we'll have to replace the calls to `.fetch_rent_reimbursements()` with .`fetch_delegation_metadata()` with some minor adjustments.

magicblock-core/src/token_programs.rs
```diff
@@ -25,10 +27,52 @@ pub const ASSOCIATED_TOKEN_PROGRAM_ID: Pubkey =
...
+    pub ata_pubkey: Pubkey,
+    pub eata_pubkey: Pubkey,
+    pub token_program: Pubkey,
+    pub wallet_owner: Pubkey,
+    pub mint: Pubkey,
+    pub amount: u64,
+    pub token_account_data_len: u64,
+}
+
+struct ParsedTokenAccount {
```
snawaz: nit: I think the name could be `MiniTokenAccount` or `TokenAccountCommon` as it is supposed to represent the common details of the SPL token-accounts (if I correctly understood)?

magicblock-core/src/token_programs.rs
```diff
@@ -389,6 +433,135 @@ pub fn try_remap_ata_to_eata(
...
+        let account = StateWithExtensions::<Token2022Account>::unpack(data)
+            .ok()?
+            .base;
+        Some(ParsedTokenAccount {
+            mint: account.mint,
+            owner: account.owner,
+            amount: account.amount,
+            is_native: account.is_native,
+            close_authority: account.close_authority,
+        })
```
snawaz, line 555: nit: if you want, we can have one single branch (as first 165 bytes are same for both types):


```suggestion
    if is_supported_token_program(token_program) && data.len() >= SplAccount::LEN {
        let account = SplAccount::unpack(data[..SplAccount::LEN).ok()?;
        Some(ParsedTokenAccount {
            mint: account.mint,
            owner: account.owner,
            amount: account.amount,
            is_native: account.is_native,
            close_authority: account.close_authority,
        })
    } 
```

magicblock-magic-program-api/src/instruction.rs
```diff
@@ -184,12 +184,18 @@ pub enum MagicBlockInstruction {
...
-    /// Unsed instruction slot.
-    /// -- can be repurposed --
-    /// This variant was originally used for `ScheduleCommitFinalize`, but that
-    /// instruction was removed. It is intentionally left unused so the wire
-    /// discriminant can be repurposed in a future protocol update.
-    Unused,
+    /// Creates or verifies a local rent-pending projected ATA.
+    ///
+    /// # Account references
+    /// - **0.** `[SIGNER]` Payer/sponsor
```
snawaz: `[SIGNER]` should be `[SIGNER, WRITE]`.

magicblock-processor/src/executor/processing.rs
```diff
@@ -501,6 +540,20 @@ fn transaction_balances(
     )
 }
 
+fn created_rent_pending_ata_info(
```
snawaz: nit: `created_` in the beginning feels like a typo; renaming it to `create_` isn't correct either. 

Maybe rename it to `get_created_rent_pending_ata_info`? 

Or maybe `get_materialized_rent_pending_ata_info`? If so, rename the fields and related functions in the TLS file.

magicblock-core/src/token_programs.rs
```diff
@@ -389,6 +433,135 @@ pub fn try_remap_ata_to_eata(
...
+        || account.confined()
+        || account.undelegating()
+    {
+        return None;
+    }
+
+    let token_program = account.owner();
+    if !is_supported_token_program(token_program) {
+        return None;
+    }
```
snawaz: ✂️ because `token_program` is checked by the next function anyway (and it has to check that).

magicblock-core/src/token_programs.rs
```diff
@@ -389,6 +433,135 @@ pub fn try_remap_ata_to_eata(
...
+
+    [TOKEN_2022_PROGRAM_ID, TOKEN_PROGRAM_ID]
+        .into_iter()
+        .find_map(|token_program| {
+            try_get_rent_pending_ata_info_for_token_program(
+                pubkey,
+                account,
+                &token_program,
+            )
+        })
```
snawaz, line 478: Why not simply this, as it is done by `try_get_rent_pending_ata_info()`?:

```suggestion
   try_get_rent_pending_ata_info_for_token_program(
      pubkey,
      account,
      account.owner()
   )
```

magicblock-magic-program-api/src/instruction.rs
snawaz: We do not have to pass the same thing twice! 

`mint` and `token_program` are already passed as `accounts`. 

```suggestion
    CreateRentPendingAta {
        wallet_owner: Pubkey,
    },
```

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,687 @@
...
+    }
+    // Default owner is reserved as the MagicContext legacy-decode sentinel.
+    if wallet_owner == Pubkey::default() {
+        return Err(InstructionError::InvalidArgument);
+    }
+    if token_program != TOKEN_PROGRAM_ID
+        && token_program != TOKEN_2022_PROGRAM_ID
+    {
+        return Err(InstructionError::UnsupportedProgramId);
+    }
```
snawaz: nit: `is_supported_token_program()` can be used here.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,687 @@
...
+    let mint_pubkey =
+        *get_instruction_pubkey_with_idx(transaction_context, MINT_IDX)?;
+    let token_program_pubkey = *get_instruction_pubkey_with_idx(
+        transaction_context,
+        TOKEN_PROGRAM_IDX,
+    )?;
+
+    if mint_pubkey != mint || token_program_pubkey != token_program {
+        return Err(InstructionError::InvalidArgument);
+    }
```
snawaz: this check can be removed if we do not pass `mint` and `token_program` twice.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,687 @@
...
+        && account.owner() == &system_program::ID
+        && account.data().is_empty()
+}
+
+fn is_native_mint(mint: &Pubkey, token_program: &Pubkey) -> bool {
+    (*token_program == TOKEN_PROGRAM_ID
+        && *mint == spl_token::native_mint::id())
+        || (*token_program == TOKEN_2022_PROGRAM_ID
+            && *mint == spl_token_2022::native_mint::id())
+}
```
snawaz, line 155: nit: these are pretty useful and general; can be moved to utils or so.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,687 @@
...
+            &wallet_owner,
+            &mint,
+            &token_program,
+        ) || is_matching_existing_projected_ata(
+            &ata_pubkey,
+            &ata_shared,
+            &wallet_owner,
+            &mint,
+            &token_program,
+        ) {
```
snawaz, line 97: clarify: should it be `&&` instead of `||`?

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,687 @@
...
+}
+
+fn is_supported_rent_pending_account_extension(
+    extension: ExtensionType,
+) -> bool {
+    matches!(
+        extension,
+        ExtensionType::ImmutableOwner
+            | ExtensionType::NonTransferableAccount
+            | ExtensionType::PausableAccount
```
snawaz, line 216: curious: are we supporting only 3 extensions out of 29?

https://docs.rs/spl-token-2022/latest/spl_token_2022/extension/enum.ExtensionType.html

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,687 @@
...
+        mint,
+        &token_account_shape.required_extensions,
+        token_account_shape.initial_state,
+    )?;
+    acc.set_remote_slot(0);
+    acc.set_delegated(true);
+    acc.set_ephemeral(false);
+    acc.set_confined(false);
+    acc.set_undelegating(false);
+    ExecutionTlsStash::register_created_rent_pending_ata(ata_pubkey);
```
snawaz: this `register_created_rent_pending_ata` line should move after the next log below?

If in some rare scenario `ic_msg!()` fails, then Tls is already written and will not rollback probably? I also hope the current instruction is the _last_ instruction in the transaction, for the same reason.

magicblock-core/src/tls.rs
```diff
@@ -1,10 +1,13 @@
 use std::{cell::RefCell, collections::VecDeque};
 
 use magicblock_magic_program_api::args::TaskRequest;
+use solana_pubkey::Pubkey;
 
 #[derive(Default, Debug)]
 pub struct ExecutionTlsStash {
     tasks: VecDeque<TaskRequest>,
+    created_rent_pending_atas: VecDeque<Pubkey>,
+    recorded_rent_pending_ata_materializations: VecDeque<Pubkey>,
```
snawaz: The names can be improved (please see the other comment on `created_rent_pending_ata_info` for the context):

```suggestion
    materialized_rent_pending_atas: VecDeque<Pubkey>,
    scheduled_rent_pending_ata_materializations: VecDeque<Pubkey>,
```

magicblock-committor-service/src/intent_executor/task_info_fetcher.rs
```diff
@@ -53,13 +63,67 @@ pub trait TaskInfoFetcher: Send + Sync + 'static {
         min_context_slot: u64,
     ) -> TaskInfoFetcherResult<Vec<Pubkey>>;
 
+    /// Fetches rent reimbursement addresses, using caller-provided defaults
+    /// when allowed metadata is missing on base.
+    async fn fetch_rent_reimbursements_with_missing_as(
```
snawaz: naming: what does `_as` stands for (in the name)? I guess it is the actual English word `as`, not a short form of anything.

If so, I'd suggest these names:

```rust
// first choice
async fn fetch_rent_reimbursements_with_fallbacks(&self, 
  delegated_accounts: &[Pubkey],  // also renamed
  min_context_slot: u64,
  metadata_fallbacks: &HashMap<Pubkey, Pubkey> // matching name
) {
  ...
}

// second choice
async fn fetch_rent_reimbursements_with_defaults(&self,
  delegated_accounts: &[Pubkey], // also renamed
  min_context_slot: u64,
  metadata_defaults: &HashMap<Pubkey, Pubkey> // matching name
) {
  ...
}

```

magicblock-committor-service/src/tasks/mod.rs
```diff
@@ -45,6 +57,8 @@ pub enum TaskStrategy {
 
 #[derive(Clone, Debug)]
 pub enum BaseTaskImpl {
+    InitializeRentPendingAta(RentPendingAtaTask),
+    DelegateRentPendingAta(RentPendingAtaTask),
```
snawaz, line 51: code-organization: I think these (and other changes in this crate) do not belong to `magicblock-commitor-service` crate. It is already too convoluted, and till now all things are related to commits and other DLP instructions. 

We can have another crate for ESPL program and I believe it's going to be much simpler. What do you say? 🤔 

I'm not necessarily suggesting we do this in this PR itself. But I'd just like to know your thoughts on this direction. If the current approach also feels a bit odd to you, we can first make it work and then refactor it later.

magicblock-committor-service/src/tasks/mod.rs
```diff
@@ -141,6 +179,70 @@ impl LabelValue for BaseTaskImpl {
...
 }
 
+#[derive(Clone, Debug)]
+pub struct RentPendingAtaTask {
+    pub materialization: RentPendingAtaMaterialization,
+}
+
+impl RentPendingAtaTask {
+    const INITIALIZE_EPHEMERAL_ATA_IX: u8 = 0;
+    const DELEGATE_EPHEMERAL_ATA_IX: u8 = 4;
```
snawaz, line 185: question: can we include `e-token-api` as dep, so that we do not have to use these constants? we can also use the builders and many things then.

magicblock-committor-service/src/tasks/mod.rs
snawaz: As for these members, I think ideally we should have two different tasks for `InitializeRentPendingAta` and `DelegateRentPendingAta`. Currently, `RentPendingAtaTask` isn't a single task, it is a composite task.

magicblock-core/src/tls.rs
```diff
@@ -23,9 +26,41 @@ impl ExecutionTlsStash {
         EXECUTION_TLS_STASH.with_borrow_mut(|stash| stash.tasks.pop_front())
     }
 
+    pub fn register_created_rent_pending_ata(pubkey: Pubkey) {
+        EXECUTION_TLS_STASH.with_borrow_mut(|stash| {
+            stash.created_rent_pending_atas.push_back(pubkey)
+        });
+    }
+
+    pub fn next_created_rent_pending_ata() -> Option<Pubkey> {
```
snawaz: `pop_` has clear semantics, compared to `next_` which doesn't imply _remove_.

so it should be `pop_created_rent_pending_ata` or simply `remove_*`.

programs/magicblock/src/schedule_transactions/process_schedule_commit.rs
snawaz, line 1: concern: I'm not sure if we should use the same processor for both commits and ATA materializations.

snawaz (changes requested):
I've reviewed most parts thoroughly, though there are still a few areas I'd like to go through one more time before I'm fully confident...especially the scheduling part, as well as the current approach of executing the ESPL instructions in the committor service, and re-reading `docs/mimd-rent-pending-ephemeral-ata.md` (the phrase _ephemeral_ should be removed from the filename).

For now, I've posted quite a few comments. Hoping some of them can be addressed in the meantime (assuming they make sense 😄).

docs/mimd-rent-pending-ephemeral-ata.md
GabrielePicco: This is arbitrary, it's just an easy to recognize one, for which we can guarantee there is no associated private key

docs/mimd-rent-pending-ephemeral-ata.md
GabrielePicco: they would behave exactly as a `delegated` account from a security perspective, but not against clarifying the difference with an alias

docs/mimd-rent-pending-ephemeral-ata.md
GabrielePicco: yes

magicblock-core/src/token_programs.rs
GabrielePicco: Seems a stylistic preference, not against renaming to `TokenAccountCommon`

magicblock-committor-service/src/tasks/mod.rs
GabrielePicco: I agree for the refactoring, but for the task, is important to have it atomic so that it either fail or succeed in the same transaction

magicblock-processor/src/executor/processing.rs
GabrielePicco: Renamed the verifier and TLS flow to `newly_created_*`, `scheduled_*`, and `pop_*` in `738fa1dd4` so creation and base materialization remain distinct.

magicblock-core/src/token_programs.rs
GabrielePicco: After scheduling undelegation the account owner is intentionally replaced with the Delegation Program, so this path cannot use `account.owner()` to identify SPL vs Token-2022 and must try both layouts.

magicblock-magic-program-api/src/instruction.rs
GabrielePicco: Updated to `[SIGNER, WRITE]` in `738fa1dd4`.

magicblock-core/src/token_programs.rs
GabrielePicco: Removed the redundant pre-check in `738fa1dd4`; the parsing helper remains the single token-program gate.

magicblock-committor-service/src/intent_executor/task_info_fetcher.rs
GabrielePicco: Addressed by the merge in `a78db8b79`: the rent-pending path now uses delegation metadata with explicit missing-account fallbacks.

magicblock-core/src/token_programs.rs
GabrielePicco: Keeping the Token-2022 `StateWithExtensions` path so the full TLV layout is validated; parsing only the common 165-byte prefix would accept malformed extension data.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
GabrielePicco: `||` is intentional: idempotent creation accepts either an existing rent-pending marker or an already materialized projected ATA.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
GabrielePicco: This is not general 3-of-29 extension support: these are the required account-side marker extensions that can be initialized safely. Required stateful extensions are rejected until settlement preserves their state.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
GabrielePicco: These helpers encode rent-pending-specific extension acceptance and initialization policy and currently have no second caller, so they remain colocated with the processor.

magicblock-magic-program-api/src/instruction.rs
GabrielePicco: Removed `mint` and `token_program` from the payload in `738fa1dd4`; both are now derived from the instruction accounts.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
GabrielePicco: Exported and reused `is_supported_token_program()` in `738fa1dd4`.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
GabrielePicco: Removed along with the duplicated payload fields in `738fa1dd4`.

magicblock-core/src/tls.rs
GabrielePicco: Renamed the TLS fields to `newly_created_rent_pending_atas` and `scheduled_rent_pending_ata_materializations` in `738fa1dd4`.

magicblock-core/src/tls.rs
GabrielePicco: Renamed the destructive read to `pop_newly_created_rent_pending_ata()` in `738fa1dd4`.

magicblock-committor-service/src/tasks/mod.rs
GabrielePicco: The current API crate is unpublished `0.0.0`, adds Pinocchio/`wheels` dependencies, and does not expose complete Solana `Instruction` builders for these calls. Keeping the small builders local for now; a lightweight published client API would be the clean follow-up.

programs/magicblock/src/schedule_transactions/process_schedule_commit.rs
GabrielePicco: Keeping materialization in the commit scheduler because metadata, fees, and the commit intent must be recorded atomically. A separate processor would introduce a partial-scheduling boundary.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
GabrielePicco: `ic_msg!` cannot fail, TLS is cleared on every transaction outcome, and the verifier runs after the complete transaction, so the instruction does not need to be last.

magicblock-committor-service/src/intent_executor/task_info_fetcher.rs
GabrielePicco: Addressed by the merge in `a78db8b79`: the API is now `fetch_delegation_metadata_with_fallbacks`.

magicblock-committor-service/src/tasks/task_builder.rs
```diff
@@ -167,13 +215,18 @@ impl TasksBuilder for TaskBuilderImpl {
...
         } = Self::fetch_commit_stage_info(
             intent_bundle,
             task_info_fetcher,
             persister,
         )
         .await?;
 
+        tasks.extend(Self::rent_pending_materialization_tasks(
+            rent_pending_ata_materializations,
+        ));
```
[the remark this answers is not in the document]
GabrielePicco: Addressed in `c830b2d8e`. `IntentSizeValidator` now reuses the deduplicating materialization-task constructor from `TaskBuilderImpl`, so both initialize and delegate tasks are included in pre-acceptance commit-stage fit checks. Added direct task-list and fit-boundary regression coverage.

magicblock-committor-service/src/tasks/intent_size_validator.rs
```diff
@@ -54,11 +55,15 @@ impl IntentSizeValidator {
...
-    /// standalone/base actions, and the smallest-by-default
-    /// `Commit`/`CommitFinalize` task for each committed account.
+    /// standalone/base actions and rent-pending materializations, plus the
+    /// smallest-by-default `Commit`/`CommitFinalize` task for each committed
+    /// account.
     fn commit_tasks(intent: &MagicIntentBundle) -> Vec<BaseTaskImpl> {
         let mut tasks: Vec<BaseTaskImpl> =
             create_action_tasks(&intent.standalone_actions).collect();
+        tasks.extend(TaskBuilderImpl::rent_pending_materialization_tasks(
+            intent.rent_pending_ata_materializations.iter().cloned(),
```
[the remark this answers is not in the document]
GabrielePicco: There is a sufficient margin that covers for the idempotent creation

GabrielePicco:
Currently reworking the implementation to avoid custom commit/undelegation logic

PR #1387, opened 2026-07-02
fix: owner race on undelegation on clone failures

GabrielePicco:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved undelegation handling so the committed account snapshot now preserves the original owner during same-transaction undelegation flows.
  * Live accounts continue transitioning into the undelegating state without losing delegation status unexpectedly.

* **Tests**
  * Added coverage for scheduled undelegation behavior to verify the committed bundle and on-chain account state stay consistent after execution.


snawaz (approved):
LGTM!

PR #1388, opened 2026-07-02
Fix committor cleanup ordering

GabrielePicco:
## Summary
- run committor cleanup before unblocking conflicting intents so reused buffer PDAs cannot be closed by a previous detached cleanup
- narrow successful undelegation cache resets to the accounts actually undelegated while still resetting all committed accounts on execution errors
- add coverage for cleanup-before-next-conflicting-intent ordering

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache refresh behavior so only the affected undelegation-related entries are cleared during successful execution, while broader resets still occur on failures.
  * Increased the default cache capacity, allowing more recent entries to stay available before older ones are evicted.

### DIFF
--- a/magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
+++ b/magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
@@ -297,6 +297,14 @@ where
             warn!(error = ?err, "No result listeners");
         }
 
+        // Cleanup BEFORE unblocking conflicting intents: a detached cleanup
+        // can close buffers a next intent just re-initialized (same PDAs on
+        // nonce reuse). The result is already broadcast, so this only delays
+        // intents blocked on the same accounts.
+        if let Err(err) = executor.cleanup().await {
+            error!(error = ?err, "Failed to cleanup after intent");
+        }
+
         // Remove executed task from Scheduler to unblock other intents
         // SAFETY: Self::execute is called ONLY after IntentScheduler
         // successfully is able to schedule execution of some Intent
@@ -306,12 +314,6 @@ where
             .expect(POISONED_INNER_MSG)
             .complete(&intent)
             .expect("Valid completion of previously scheduled message");
-
-        tokio::spawn(async move {
-            if let Err(err) = executor.cleanup().await {
-                error!(error = ?err, "Failed to cleanup after intent");
-            }
-        });
 
         // Free worker
         drop(execution_permit);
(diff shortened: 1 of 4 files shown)


taco-paco (changes requested):

magicblock-committor-service/src/intent_execution_manager/intent_execution_engine.rs
```diff
@@ -297,6 +297,14 @@ where
             warn!(error = ?err, "No result listeners");
         }
 
+        // Cleanup BEFORE unblocking conflicting intents: a detached cleanup
+        // can close buffers a next intent just re-initialized (same PDAs on
+        // nonce reuse). The result is already broadcast, so this only delays
+        // intents blocked on the same accounts.
+        if let Err(err) = executor.cleanup().await {
```
taco-paco: There's no need to await here. There's no race-condition due to `close_buffers`.

`close_buffer=true` on success of intent - buffers can be cleaned up as intents after it shouldn't reuse those.

if it is false - we don't cleanup buffers. It doesn't leak anything as buffers has to be used as intent has to be fulfilled or buffers will be reinitialized by next one. Detach task is safe once we reset only undelegated pubkeys.

Additionally it is not desired to await here as it will slow down execution of qeueud Intents as they have to wait for cleanup. As it is safe to detach after the main issue addressed I don't see a reason to await here

See prior [PR](https://github.com/magicblock-labs/magicblock-validator/pull/1291) on the same matter if more context needed

magicblock-committor-service/src/intent_executor/mod.rs
```diff
@@ -544,19 +544,36 @@ where
...
+            .intent_bundle
+            .get_undelegate_intent_pubkeys()
+            .into_iter()
+            .chain(
+                base_intent
+                    .intent_bundle
+                    .get_commit_finalize_and_undelegate_intent_pubkeys(),
+            )
+            .flatten()
+            .collect();
```
taco-paco, line 558: This bit and addition below will be enough for a fix

GabrielePicco: Agreed, reverted — with the selective reset the successor reuses the incremented cached nonce, so the detached success-cleanup can't target a live buffer PDA, and the await would cost a base-layer round-trip per conflicting intent. 

One caveat for the record: close_buffers itself guards the failure path; success-path safety now rests on nonce-cache retention. The cache is a 1000-entry LRU and re-fetches use min_context_slot = remote_slot, which doesn't cover the previous finalize —  so an evicted entry (or a fast undelegate→re-delegate→commit) can still re-fetch stale and reproduce the collision. Rare enough to accept; if we ever want to close it fully, flooring min_context_slot at the previous intent's landed slot

taco-paco (approved):
LGTM! Thanks

magicblock-committor-service/src/intent_executor/mod.rs
taco-paco: With super huge load and cache eviction yeah there's a possibility that we can run into that case. But there must be an extreme load for that to happen and all of non overlapping intents.
With undelegation It could but I think it won't as remote-slot within account will definitely be different

PR #1396, opened 2026-07-07
Configure ledger RocksDB caching and compression

GabrielePicco:
## Summary
- add a configurable `ledger.block-cache-size` setting and pass it into ledger opening
- configure a shared RocksDB block cache with Bloom filters for ledger column families
- use LZ4 compression for fresh data and Zstd for bottommost ledger data


taco-paco (approved):
I don't mind but I'd imagine compaction speed would come in expense with this

PR #1397, opened 2026-07-07
fix: prevent transaction pipeline stall when gRPC streams wedge silently

GabrielePicco:
## Problem

gRPC laserstream connection can wedge half-open after an h2 protocol error: they stop delivering updates but never return an error. Recovery is error-driven, so nothing ever recover in this scenario. Is this happen, every `sendTransaction`/`simulateTransaction` hung indefinitely in `ensure_transaction_accounts` (blocks kept being produced with zero transactions, read RPCs stayed healthy) until a manual restart.

## Fix

- **Liveness watchdog** (`chain_laser_actor/actor.rs`): every account stream carries a slot-update filter, so a healthy connection delivers updates several times per second. If no stream update arrives for 30s while subscriptions exist, force the reconnect path. This covers connections that die without erroring.
- **Bounded stream writes** (`write_with_retry`): 10s timeout per write attempt, treated as a retryable error. A frozen sink can no longer wedge the single-threaded actor loop — this was the exact blocking point in the incident.
- **Bounded actor round-trips** (`chain_laser_client.rs`): all subscribe/unsubscribe/reconnect/shutdown calls and channel sends time out after 30s with a new `ChainPubsubActorTimeout` error instead of awaiting a wedged actor forever.
- **Fetch phase under deadline** (`fetch_cloner`): the batched wire fetch is now wrapped in `timeout_at` with the earliest claim deadline. Previously only the clone phase was deadline-bounded, so a hung fetch stranded all dedup waiters past their 60s budget.
- **Fail-fast at the RPC layer** (`aperture`): `ensure_transaction_accounts` is hard-bounded at 30s (new `EnsureAccountsTimeout` error), so even an unforeseen hang in the pipeline surfaces as transaction errors rather than silently hung requests.

### DIFF
--- a/magicblock-aperture/src/requests/http/mod.rs
+++ b/magicblock-aperture/src/requests/http/mod.rs
@@ -8,6 +8,7 @@ use hyper::{
     Request, Response,
 };
 use magicblock_accounts_db::traits::AccountsBank;
+use magicblock_chainlink::errors::ChainlinkError;
 use magicblock_core::{
     coordination_mode::CoordinationMode,
     link::transactions::{SanitizeableTransaction, WithEncoded},
@@ -285,7 +286,7 @@ impl HttpDispatcher {
         )
         .await
         .unwrap_or_else(|_elapsed| {
-            Err(magicblock_chainlink::errors::ChainlinkError::EnsureAccountsTimeout(
+            Err(ChainlinkError::EnsureAccountsTimeout(
                 ENSURE_ACCOUNTS_TIMEOUT.as_secs(),
             ))
         }) {
--- a/magicblock-aperture/src/requests/http/mod.rs
+++ b/magicblock-aperture/src/requests/http/mod.rs
@@ -1,5 +1,5 @@
 use core::str;
-use std::{mem::size_of, ops::Range};
+use std::{mem::size_of, ops::Range, time::Duration};
 
 use base64::{prelude::BASE64_STANDARD, Engine};
 use http_body_util::BodyExt;
@@ -269,16 +269,26 @@ impl HttpDispatcher {
         &self,
         transaction: &SanitizedTransaction,
     ) -> RpcResult<()> {
+        // Hard bound on account preparation: if the cloning pipeline is
+        // degraded the transaction must fail with an error instead of
+        // holding the request (and the client) hostage indefinitely.
+        const ENSURE_ACCOUNTS_TIMEOUT: Duration = Duration::from_secs(30);
+
         self.require_primary_rpc_method("transaction")?;
 
         let _timer = ENSURE_ACCOUNTS_TIME
             .with_label_values(&["transaction"])
             .start_timer();
-        match self
-            .chainlink
-            .ensure_transaction_accounts(transaction)
-            .await
-        {
+        match tokio::time::timeout(
+            ENSURE_ACCOUNTS_TIMEOUT,
+            self.chainlink.ensure_transaction_accounts(transaction),
+        )
+        .await
+        .unwrap_or_else(|_elapsed| {
+            Err(magicblock_chainlink::errors::ChainlinkError::EnsureAccountsTimeout(
+                ENSURE_ACCOUNTS_TIMEOUT.as_secs(),
+            ))
+        }) {
             Ok(res) if res.is_ok() => Ok(()),
             Ok(res) => {
                 debug!(%res, "Transaction account resolution encountered issues");
--- a/magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
+++ b/magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
@@ -247,6 +247,14 @@ impl<H: StreamHandle, S: StreamFactory<H>> ChainLaserActor<H, S> {
 
     #[instrument(skip(self), fields(client_id = %self.client_id))]
     pub async fn run(mut self) {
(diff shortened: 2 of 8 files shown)


magicblock-aperture/src/requests/http/mod.rs
```diff
@@ -269,16 +269,26 @@ impl HttpDispatcher {
...
-            .ensure_transaction_accounts(transaction)
-            .await
-        {
+        match tokio::time::timeout(
+            ENSURE_ACCOUNTS_TIMEOUT,
+            self.chainlink.ensure_transaction_accounts(transaction),
+        )
+        .await
+        .unwrap_or_else(|_elapsed| {
+            Err(magicblock_chainlink::errors::ChainlinkError::EnsureAccountsTimeout(
```
thlorenz: Would be nice to tell AI to import `magicblock_chainlink::errors::ChainlinkError` and then just use `ChainlinkError::EnsureAccountsTimeout` instead.

Maybe we need to add a code style doc to .agents. :)

magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
```diff
@@ -247,6 +247,14 @@ impl<H: StreamHandle, S: StreamFactory<H>> ChainLaserActor<H, S> {
 
     #[instrument(skip(self), fields(client_id = %self.client_id))]
     pub async fn run(mut self) {
+        // Every account stream carries a slot-update filter, so a healthy
+        // connection delivers updates several times per second. Prolonged
+        // silence while subscriptions exist means the connection died
+        // without erroring (e.g. an h2 half-open wedge) and the error-driven
+        // recovery will never fire - force a reconnect instead.
+        const STREAM_LIVENESS_TIMEOUT: Duration = Duration::from_secs(30);
```
thlorenz: We should receive an update every 400ms, shouldn't we make this timeout smaller then, i.e. 5 secs at the max?

magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
```diff
@@ -271,6 +286,7 @@ impl<H: StreamHandle, S: StreamFactory<H>> ChainLaserActor<H, S> {
                     }
                 },
                 update = self.stream_manager.next_update(), if self.stream_manager.has_any_subscriptions() => {
+                    last_stream_activity = tokio::time::Instant::now();
```
thlorenz: Maybe we could also import `tokio::time::Instant` and other `tokio::time` deps (above) to make the code read nicer.

magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
```diff
@@ -290,6 +306,28 @@ impl<H: StreamHandle, S: StreamFactory<H>> ChainLaserActor<H, S> {
                         }
                     }
                 },
+                _ = liveness_interval.tick() => {
+                    if !self.stream_manager.has_any_subscriptions() {
+                        last_stream_activity = tokio::time::Instant::now();
+                    } else if last_stream_activity.elapsed()
```
thlorenz, line 313: This liveness check also fires for program-only streams because `has_any_subscriptions()` includes the program stream, but `build_program_request` does not attach the slot-update filter that account streams use. 

The default startup program subscription can therefore sit idle with no updates and be forced through a reconnect every 30s (or if we lower this to every 5s) even though the connection is healthy. 

Could this either check only streams that actually include slot heartbeats, or add the same slot filter to program subscription requests?

thlorenz (changes requested):
I found one blocking issue in the new liveness watchdog: it treats program-only streams as requiring slot heartbeat updates, but program subscription requests do not include the slot filter.

Aside from that just minor/style issues.

magicblock-aperture/src/requests/http/mod.rs
GabrielePicco: Done in 526dba2c — ChainlinkError is imported and used directly.

magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
GabrielePicco: Done in 526dba2c — STREAM_LIVENESS_TIMEOUT is now 5s (check interval 2s, so detection within ~5-7s). Paired with the slot-heartbeat fix on program streams so the tighter window is safe for all stream types.

magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
GabrielePicco: Done in 526dba2c — interval, Instant, and MissedTickBehavior are imported from tokio::time.

magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
GabrielePicco: Good catch — done in 526dba2c. The slot-update filter is now attached to every stream via a shared slot_updates_filter() helper (build_program_request included), so program-only streams deliver the same ~400ms heartbeat and the liveness check is valid for them. Slot updates from program streams only bump chain_slot (process_subscription_update returns early), so no extra forwarding traffic.

thlorenz (approved):

PR #1401, opened 2026-07-07
fix: try to reconstruct the in-flight blockhash when in replica mode

bmuddha:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved replay and replication recovery after mid-slot restarts.
  * Transaction processing now resumes from the last persisted transaction, reducing duplicate or missed work.
  * Block verification is now rebuilt correctly when switching into replica mode at runtime.
  * Signature lookups now return transactions in the correct order for a slot and stay scoped to that slot.

### DIFF
--- a/magicblock-ledger/src/store/api.rs
+++ b/magicblock-ledger/src/store/api.rs
@@ -360,6 +360,48 @@ impl Ledger {
         }
     }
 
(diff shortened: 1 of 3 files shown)


GabrielePicco (approved):
LGTM!

PR #1408, opened 2026-07-10
fix: prevent intent tx signature aliasing on first commit after re-delegation

GabrielePicco:
## Problem

The committor builds intent transactions from fully deterministic inputs. After an undelegate → re-delegate cycle the delegation metadata nonce restarts at 0, so the next intent's first commit (nonce 1) can be **byte-identical** to a prior instance's landed commit:

## Fix

Intents whose commit uses nonce ≤ 1 are tagged with the existing per-intent uniqueness noop (spl-noop instruction carrying the intent id, already used for standalone-action intents), on **all** their stages — the two-stage finalize tx carries no nonce bytes, so it aliases across cycles exactly like the commit tx. Collisions are only possible at first-commit (the on-chain nonce never regresses otherwise), so other intents are left untouched. Retries of the same intent keep the same id, preserving intentional dedup.

## Summary by CodeRabbit

- **New Features**
  - Added per-intent uniqueness no-op markers to commit/finalize flows to improve action/message distinction.
  - Transaction sizing and lookup-table planning now reserve and simulate space for the marker instruction.
  - Added the required noop program to integration-test devnet configurations.
- **Bug Fixes**
  - Improved recovery when cached/stale commit IDs or transaction-size limits occur by preserving the correct uniqueness marker across retries.
- **Tests**
  - Expanded coverage for first-commit vs non-first commits, finalize-only, empty inputs, and verifying marker instruction presence/fit behavior.


magicblock-committor-service/src/tasks/task_strategist.rs
```diff
@@ -395,11 +399,14 @@ impl TaskStrategist {
...
         let calculate_tx_length = |tasks: &[BaseTaskImpl]| {
-            match TransactionUtils::assemble_tasks_tx(
+            // Reserve space for the constant-size uniqueness noop so fit
+            // decisions match the assembled transaction when the nonce is set.
+            match TransactionUtils::assemble_tasks_tx_with_standalone_action_nonce(
                 &Keypair::new(), // placeholder
                 tasks,
                 u64::default(), // placeholder
                 &[],
+                Some(0), // placeholder
```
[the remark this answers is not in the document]
GabrielePicco: Addressed in e6b67c63: the uniqueness nonce is now threaded into `TaskStrategist` (`build_execution_strategy`/`build_strategy`/both fit checks), so the noop is reserved only for strategies that will actually carry it, and the strategist sets it on the strategies it returns. Non-first commits are sized exactly as before; boundary tests restored to 22/23 plus a new test covering the with-nonce reservation.

magicblock-committor-service/src/tasks/task_strategist.rs
```diff
@@ -123,11 +123,14 @@ impl TaskStrategist {
     /// Builds execution strategy from [`BaseTask`]s
     /// 1. Optimizes tasks to fit in TX
     /// 2. Chooses the fastest execution mode for Tasks
+    /// `uniqueness_nonce` is rendered as a constant-size noop instruction on
+    /// every produced strategy, and is accounted for in all fit decisions.
```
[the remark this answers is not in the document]
GabrielePicco: Documented in 63c7992c: the guide's execute flow now includes the uniqueness-nonce step, a new "Per-intent uniqueness noop" section covers the aliasing rationale, scope (nonce <= 1 intents, all stages, standalone actions), fit accounting, and the base-layer noop-program requirement, and invariant 12 pins the behavior. The same commit moves the diff-args boundary tests to the new 636/637 limit (verified against a local devnet).

magicblock-committor-service/src/tasks/task_strategist.rs
```diff
@@ -279,8 +297,15 @@ impl TaskStrategist {
...
 
         // Create final tx
-        let instructions =
+        let mut instructions =
             TransactionUtils::tasks_instructions(&placeholder.pubkey(), tasks);
+        // Reserve space for the constant-size uniqueness noop so fit decisions
+        // match the assembled transaction.
+        if let Some(nonce) = uniqueness_nonce {
+            instructions.push(
+                TransactionUtils::standalone_action_noop_instruction(nonce),
```
taco-paco: Let's rename `standalone_action_noop_instruction` into `uniqueness_noop_instruction` as it is not only action_nonce now

magicblock-committor-service/src/intent_executor/utils.rs
```diff
@@ -177,7 +177,7 @@ pub(in crate::intent_executor) fn handle_cpi_limit_error(
     let commit_strategy = TransactionStrategy {
         optimized_tasks: commit_stage_tasks,
         lookup_tables_keys: commit_alt_pubkeys,
-        standalone_action_nonce: None,
+        standalone_action_nonce: strategy.standalone_action_nonce,
```
taco-paco: That is no longer only `standalone_action_nonce` but a broader `uniqueness_nonce`. Let's rename it accrodingly

magicblock-committor-service/src/tasks/task_strategist.rs
taco-paco:
```suggestion
    pub uniqueness_nonce: Option<u64>,
```

magicblock-committor-service/src/tasks/task_strategist.rs
```diff
@@ -279,8 +297,15 @@ impl TaskStrategist {
             TransactionUtils::dummy_lookup_table(&unique_involved_pubkeys);
 
         // Create final tx
-        let instructions =
+        let mut instructions =
             TransactionUtils::tasks_instructions(&placeholder.pubkey(), tasks);
+        // Reserve space for the constant-size uniqueness noop so fit decisions
+        // match the assembled transaction.
+        if let Some(nonce) = uniqueness_nonce {
+            instructions.push(
```
taco-paco: For correct `unique_involved_pubkeys` calculation `standalone_action_noop_instruction` also shall be supplied into `TransactionUtils::unique_involved_pubkeys(`.

Maybe `budget_instructions` could be renamed to something more general, like service_instructions or something better, where both budget and uniqueness ix would live

magicblock-committor-service/src/tasks/task_strategist.rs
taco-paco: Should include uniquness ix here

magicblock-committor-service/src/tasks/task_strategist.rs
```diff
@@ -279,8 +297,15 @@ impl TaskStrategist {
             TransactionUtils::dummy_lookup_table(&unique_involved_pubkeys);
 
         // Create final tx
-        let instructions =
+        let mut instructions =
             TransactionUtils::tasks_instructions(&placeholder.pubkey(), tasks);
+        // Reserve space for the constant-size uniqueness noop so fit decisions
```
taco-paco: The code here closely duplicates logic in `assemble_tasks_tx_with_standalone_action_nonce`. This already lead to us missing adding uniqueness ix here. 

Let's reuse `assemble_tasks_tx_with_standalone_action_nonce` to avoid future divergence.

The code could be something like that

```rust
  pub fn attempt_lookup_tables(
      tasks: &[BaseTaskImpl],
      standalone_action_nonce: Option<u64>,
  ) -> bool {
      let placeholder = Keypair::new();
      let lookup_tables = TransactionUtils::dummy_lookup_table(
          &Self::collect_lookup_table_keys(
              &placeholder.pubkey(),
              tasks,
              standalone_action_nonce,
          ),
      );
  
      TransactionUtils::assemble_tasks_tx_with_standalone_action_nonce(
          &placeholder,
          tasks,
          u64::default(),
          &lookup_tables,
          standalone_action_nonce,
      )
      .is_ok_and(|tx| {
          serialized_transaction_size(&tx) <= MAX_TRANSACTION_WIRE_SIZE
      })
  }
```

taco-paco (changes requested):
Looks good! Some renamings are due + some places have divergence in logic now which has to be corrected.

`attempt_lookup_tables(` could be simplified and made less error prone by reusing existing logic and avoiding duplication

magicblock-committor-service/src/tasks/task_strategist.rs
taco-paco: See proposed flow [here](https://github.com/magicblock-labs/magicblock-validator/pull/1408#discussion_r3556843101)

magicblock-committor-service/src/intent_executor/mod.rs
```diff
@@ -260,12 +261,16 @@ where
             (commit_tasks?, finalize_tasks?)
         };
 
+        let uniqueness_nonce = requires_uniqueness_nonce(&commit_tasks)
+            .then_some(intent_bundle.id);
```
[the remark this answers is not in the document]
GabrielePicco: Good catch — fixed in 85932316: `handle_commit_id_error` now re-tags the strategy with the intent id when the re-fetched commit id lands back on 1 (executors carry the intent id). Covered by a unit test (`test_commit_id_recovery_sets_uniqueness_nonce`).

magicblock-committor-service/src/tasks/task_strategist.rs
GabrielePicco: Renamed in 85932316 (`uniqueness_noop_instruction`, `UNIQUENESS_NOOP_PROGRAM_ID`, `assemble_tasks_tx_with_uniqueness_nonce`).

magicblock-committor-service/src/intent_executor/utils.rs
GabrielePicco: Renamed to `uniqueness_nonce` throughout in 85932316.

magicblock-committor-service/src/tasks/task_strategist.rs
GabrielePicco: Done in 85932316.

magicblock-committor-service/src/tasks/task_strategist.rs
GabrielePicco: Done in 85932316 — `attempt_lookup_tables` now assembles through `assemble_tasks_tx_with_uniqueness_nonce` with tables from `collect_lookup_table_keys`, exactly the inputs the real assembly uses, so the estimate can't diverge.

magicblock-committor-service/src/tasks/task_strategist.rs
GabrielePicco: Done in 85932316: `collect_lookup_table_keys` takes the nonce and folds the noop ix into `unique_involved_pubkeys`. Note it's size-neutral either way — the noop has no account metas and invoked program ids can't be moved into ALTs (`try_compile` keeps them static) — but it keeps the collection honest and both paths now share one assembly flow.

magicblock-committor-service/src/tasks/task_strategist.rs
GabrielePicco: Done in 85932316 (see thread above).

magicblock-committor-service/src/intent_executor/utils.rs
```diff
@@ -130,11 +131,19 @@ pub(in crate::intent_executor) async fn handle_commit_id_error<
         }
     }
 
+    // Re-fetched nonces land back on 1 after a re-delegation; such a retry is
+    // a first-commit transaction and must carry the uniqueness noop.
+    if strategy.uniqueness_nonce.is_none()
+        && super::requires_uniqueness_nonce(&strategy.optimized_tasks)
+    {
+        strategy.uniqueness_nonce = Some(intent_id);
```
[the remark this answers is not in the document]
GabrielePicco: Addressed in 4bb1f43d: when recovery re-tags the commit stage as a first commit, `patch_commit_strategy` now propagates the nonce to the already-built finalize strategy, so both stages carry the noop. Deliberately not refitting both stages: a boundary-sized stage that no longer fits with the noop fails loudly at the preparator's fail-early fit check instead of silently aliasing — the right failure mode for this rare path (stale cache + re-delegation + exact wire-size boundary) without re-running the strategist on strategies whose buffers/ALTs may already be prepared.

magicblock-committor-service/src/intent_executor/utils.rs
```diff
@@ -130,11 +131,19 @@ pub(in crate::intent_executor) async fn handle_commit_id_error<
         }
     }
 
+    // Re-fetched nonces land back on 1 after a re-delegation; such a retry is
+    // a first-commit transaction and must carry the uniqueness noop.
+    if strategy.uniqueness_nonce.is_none()
+        && super::requires_uniqueness_nonce(&strategy.optimized_tasks)
```
taco-paco, line 137: In case old ids were incorrect it can we very well be that old `uniqueness_nonce.is_none()` was also incorrect. 
We shall judge on the current state of nonces 
```suggestion
    // a first-commit transaction and must carry the uniqueness noop.
    if super::requires_uniqueness_nonce(&strategy.optimized_tasks)
```

magicblock-committor-service/src/intent_executor/two_stage_executor.rs
```diff
@@ -191,9 +194,17 @@ where
...
                     committed_pubkeys,
-                    &mut self.state.commit_strategy
+                    &mut self.state.commit_strategy,
+                    self.intent_id,
                 )
                     .await?;
+                // If recovery re-tagged the commit stage as a first commit,
+                // the finalize stage aliases the same way and must carry the
+                // uniqueness noop too.
+                if self.state.finalize_strategy.uniqueness_nonce.is_none() {
```
taco-paco, line 204: > // If recovery re-tagged the commit stage as a first commit,

Comment says this but we check a stale `uniqueness_nonce` from finalize strategy.

We wether:
1. Prior to patching remember old uniquness_nonce and then compare it with new one and only after update
2. Don't have if condition at all and just assign commit uniqueness_nonce to finalize one

I'm for 2

magicblock-committor-service/src/tasks/task_strategist.rs
taco-paco: Thanks looks much cleaner!

magicblock-committor-service/src/intent_executor/utils.rs
GabrielePicco: Fair point on judging from refetched state — but stepping back, this recovery path can't actually produce the alias: the patch loop calls `invalidate_cached_blockhash()` before every retry, so a rebuilt transaction always signs with a fresh blockhash, strictly newer than any landed transaction's. Byte-identity is only possible in the scoped case (first commit after re-delegation within the 5s blockhash-cache TTL), which the main fix covers. The re-tag here is already conservative defense-in-depth beyond the minimal fix — I'd keep it as-is to merge, and polish the recovery-path semantics (this condition + the finalize propagation guard) in a follow-up.

magicblock-committor-service/src/intent_executor/two_stage_executor.rs
GabrielePicco: Same scoping as the sibling thread: the recovery retry always re-signs with a fresh blockhash, so it can't alias a landed transaction — keeping this conservative guard as-is for the minimal fix and revisiting both recovery-path conditions in a follow-up.

magicblock-committor-service/src/intent_executor/utils.rs
taco-paco: We clean cached blockhashes anyway but still got into a situation where a flashtrade encountered aliasing. Cleaned cache on our side doesn't mean cleaned cache on RPC or that we can't get same blockhash and flashtrade proved it. This PR is solely to defeat aliasing and here we argue if it can happen or not :)

I agree that it is a defensive patch already but we should then make it correctly without relying on some external conditions. It is also very small change to make a follow up on it,

taco-paco (approved):

magicblock-committor-service/src/intent_executor/utils.rs
GabrielePicco: What change the transaction signature is the blockhash used to build/sign it, which is cached only on our side (for 5 seconds). Everything else is irrelevant, RPCs don't cache the blockchash

PR #1409, opened 2026-07-10
feat: add id for actions

taco-paco:
## Summary

Add a stable per-action id  so actions can be addressed independently of their position in the bundle — positional indexing.
This is needed for coming patching of intents as we have to refer to actions somehow

## Breaking Changes
- [ ] None
- [x] Yes — migration path described below

@lucacillario During deploy we need to make sure that MagicContext is empty

## Summary by CodeRabbit

* **Enhancements**
  * Scheduled actions now use stable identifiers, improving action retrieval and callback association.
  * Action construction and scheduling support consistent identifier assignment across bundles.

* **Bug Fixes**
  * Improved reliability when locating actions during scheduling and callback processing.

* **Tests**
  * Updated unit and integration test fixtures to include action identifiers.

### DIFF
--- a/programs/magicblock/src/magic_scheduled_base_intent.rs
+++ b/programs/magicblock/src/magic_scheduled_base_intent.rs
@@ -51,6 +51,10 @@ pub struct ConstructionContext<'a, 'ic, 'ix_data> {
     /// to the delegation program. Legacy `ScheduleBaseIntent` sets this to
     /// `false` for backward compatibility with old deployed contracts.
     pub secure: bool,
+    /// Next id to assign to a `BaseAction` being constructed. Incremented
+    /// as actions are built, so each action in a bundle gets a unique,
+    /// stable id in construction order.
+    next_action_id: u64,
 }
 
 impl<'a, 'ic, 'ix_data> ConstructionContext<'a, 'ic, 'ix_data> {
@@ -65,11 +69,18 @@ impl<'a, 'ic, 'ix_data> ConstructionContext<'a, 'ic, 'ix_data> {
             signers,
             invoke_context,
             secure,
+            next_action_id: 0,
         }
     }
 
     pub fn transaction_context(&self) -> &TransactionContext<'ix_data> {
         &*self.invoke_context.transaction_context
+    }
+
+    fn next_action_id(&mut self) -> u64 {
+        let id = self.next_action_id;
+        self.next_action_id += 1;
+        id
     }
 }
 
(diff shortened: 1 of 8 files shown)


programs/magicblock/src/magic_scheduled_base_intent.rs
```diff
@@ -246,10 +257,11 @@ impl From<MagicBaseIntent> for MagicIntentBundle {
 impl MagicIntentBundle {
     pub fn try_from_args(
         args: MagicIntentBundleArgs,
-        context: &ConstructionContext<'_, '_, '_>,
+        context: &mut ConstructionContext<'_, '_, '_>,
     ) -> Result<Self, InstructionError> {
         Self::validate(&args, context)?;
 
+        // NOTE: Order shall be identical to AddActionCallback's action ordering.
```
snawaz, line 264: concern: The **ordering requirement** makes this quite **error-prone**. If we're using IDs (which are supposed to replace indices), why should ordering matter at all?

Can we make this ordering-agnostic?

BTW, currently, are we constructing the same bundle more than once? If not, why is there such a "ordering" requirement to begin with?

snawaz (commented):
Posted one concern. Apart from that, everything looks straightforward and good.

programs/magicblock/src/magic_scheduled_base_intent.rs
```diff
@@ -65,12 +69,19 @@ impl<'a, 'ic, 'ix_data> ConstructionContext<'a, 'ic, 'ix_data> {
             signers,
             invoke_context,
             secure,
+            next_action_id: 0,
```
bmuddha, line 72: is this thing local to transaction, doesn't it need to be persisted or shared between transactions?

bmuddha (approved):
LGTM, but with a question to clarify

programs/magicblock/src/magic_scheduled_base_intent.rs
taco-paco: In context of validator this isn't a concern for us, but I left it for sdk/users. ID's are assigned in order specified here so our SDK needs to correctly refer to those indexes in order to append callbacks correctly or we would have to deserialize the whole intent which isn't desired.

programs/magicblock/src/magic_scheduled_base_intent.rs
taco-paco: This is local to Intent, within intent actions will get IDs. if action would aquire a pda it would be derived as "intent_id, action_id"

snawaz (approved):

PR #1411, opened 2026-07-10
fix: distinguish ER account presence from delegation status

snawaz:
This PR deprecates `accounts_delegated_on_base_and_er()` as it incorrectly treats **"missing-account-on-ER"** as **"not delegated"** and provides better API that separates `Missing` from `NotDelegated`. See `enum AccountStatusOnEr`.

## Summary by CodeRabbit

* **New Features**
  * Added detailed delegation status reporting for accounts, distinguishing missing, delegated, and non-delegated states.
  * Automatically recovers locally missing delegated accounts before processing undelegation requests.

* **Bug Fixes**
  * Improved undelegation readiness checks to prevent invalid requests from being processed.

* **Observability**
  * Added structured skip and drop reasons, including delegation state, receiver lag, subscriber availability, and relevant slot information.

PR #1402, opened 2026-07-07
refactor: move commit-and-undelegate scheduling service from magicblock-accounts to magicblock-services

snawaz:
## Breaking Changes
- [ ] None
- [ ] Yes — migration path described below

## Summary by CodeRabbit

* **New Features**
  * Added background handling for DLP undelegation requests and related scheduling during validator startup.
  * Updated validator flow to process scheduled intents directly, improving how commit and undelegation work is coordinated.

* **Bug Fixes**
  * Fixed ownership and routing of undelegation and commit processing so the correct service handles each step.
  * Removed outdated references to the legacy account-processing path across docs and configuration.

PR #1391, opened 2026-07-03
feat: Support validator-requested undelegation in DLP finalize paths

snawaz:
## Breaking Changes
- [ ] None
- [ ] Yes — migration path described below

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved undelegation handling so finalize flows no longer add a separate undelegation step when one isn’t needed.
  * Refined task scheduling to keep post-undelegation actions in the correct order.

* **Documentation**
  * Clarified finalize and undelegation behavior in the service and validator specs.
  * Updated guidance around auto-undelegation, request handling, and rent repayment.

* **Tests**
  * Expanded coverage for finalize and commit-finalize task behavior, including account expectations and optimized task counts.

PR #1378, opened 2026-07-01
feat: Impl UndelegationRequest backfill loop

snawaz:
- Added a backfill loop that periodically (every `5m` or so &mdash; configurable) fetches UndelegationRequest PDAs and processes them.

PR #1413, opened 2026-07-11
feat: auto-undelegate owner-requested accounts during commit

snawaz:
If the scheduled intent is only `Commit` / `CommitFinalize` (without planned _undelegate_) for an account whose metadata says `undelegation_requesor == OwnerProgram`, we commit/finalize and leave the account **delegated** even though the owner program is asking for undelegation!!

So this PR automatically appends `undelegate` tasks for **owner-program undelegation requests** in cases where undelegation is otherwise not planned.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved finalize-stage handling to automatically add the missing implicit undelegation for committed accounts with pending owner-program undelegation requests when not explicitly targeted.
  * Avoided generating undelegation tasks for accounts already covered by undelegation intent pubkeys.
  * Refined successful-execution cache resets so only accounts planned for undelegation are updated (failure behavior unchanged).

* **Tests**
  * Expanded coverage for implicit owner-program undelegation across standard and commit/finalize flows, including correct task fields and negative cases.

* **Documentation**
  * Updated finalize-stage DLP/undelegation guidance to reflect the new implicit undelegation behavior.


snawaz:
* **#1413** <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1413?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1413?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1411** <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1411?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1378** <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1378?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#1391](https://github.com/magicblock-labs/magicblock-validator/pull/1391) <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1391?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1402](https://github.com/magicblock-labs/magicblock-validator/pull/1402) <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1402?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1357** <a href="https://app.graphite.com/github/pr/magicblock-labs/magicblock-validator/1357?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

taco-paco (changes requested):
I don't think we can modify user created intents implicitly like that. If user scheduled some intent simultaneously we shall fulfill it first and then execute UndelegationRequest intent

snawaz:
> I don't think we can modify user created intents implicitly like that. If user scheduled some intent simultaneously we shall fulfill it first and then execute UndelegationRequest intent

Why not? Softwares evolve. We only need to document the behavior. 

I'm more interested in the technical reasons for why this shouldn't be implemented. Is there a technical constraint I'm missing? Otherwise, it feels like an artificial restriction.

taco-paco:
I don't see how such behaviour is provable on L1 within a fraud proof, In classic L2 any state that surfaces is a result of transaction and hence is provable wether by ZK or challanged and resolved. Changes in intent in the following way is a result not of an ER transaction but of an L1 transaction. I see it is as very chalenging proving L1 state within L1 itself. It is a same problem we have with readonly cloned accounts.
 
Now the flow that is standart in this case is:
1. User request on L1(Base) a transaction to be ran on L2(ER)
2. Within a certain timeframe this TX is included on L2, e.g execute
3. User request is closed on state commitment as there's a proof that transaction was included

In our context tx - undelegation request.

Additionally, all of the current flows - AML undelegations, cloning and patching of intents(upcoming) is done via transaction. This goes ot of the standart way as say chainlink could also modify account in place but we schedule a transaction instead

taco-paco:
This also introduces a race condition: 
Say for same account
1. User on ER scheduled Commit Intent 1
2. User on ER Scheduled Commit&Undelegate intent 2.
3. User on base requested Undelegation
4. Intent one gets picked up and Undelegation added to it
5. Intent 2 becomes impossible to execute

Say Intent 2 had a critical action that expected to execute final transfers on Base and it also already settled final balances on ER. This Intent would never be able to execute now. With undelegation request being executed in separate tx it is not an issue as TXs attempting to schedule Intent for that account would fail as account marked as undelegated

All in all I'm not in favor of this flow as it circles around proper rules of intent scheduling that happens during TX execution

One of other edge cases: Intent may become impossible to fit in txs. User schedules an intent and we check during execution if it is possible to fit in 2 txs. We say it is and intent gets scheduled. We add UndelegateTask here which makes intent to large even with the split on 2 txs.

GabrielePicco:
Closing as stale

PR #1415, opened 2026-07-13
feat: add ephemeral-system-program that handles ephemeral account lifecycle

taco-paco:
## Summary

extract ephemeral account creation logic into separate program . As for now both `magic-program` and `ephemeral-system-program` reuse the same code. `magic-program` can't CPI into `esp` as that would break existing integration because user would jave to supply `esp` AccountMeta in tx which they don't rn.

The purposes are: decoupling, avoiding circular cpis `magic-program(AcceptIntents)->outbox-program(create outbox)->magic-program(create ephemeral)`. With addition the flow will be `magic-program(AcceptIntents)->outbox-program(create outbox)->esp(create ephemeral)`

## Breaking Changes
- [x] None

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated ephemeral system program for creating, resizing, and closing ephemeral accounts.
  * Enabled ephemeral account operations through cross-program invocations.
  * Added support for interoperability between the existing MagicBlock program and the new system program.
  * Added validation to reject unsupported direct calls.

* **Bug Fixes**
  * Updated account protection and blacklist handling to include the ephemeral system program.


bmuddha (approved):
LGMT, a few notes:
1. I suggest we upgrade the ephemeral-sdk to use the new program, this way users will gradually move away from magic-program
2. We should consider keeping the codebase for all of the builtins in their own separate crates, instead of putting everything into magicprogram.

taco-paco:
> I suggest we upgrade the ephemeral-sdk to use the new program, this way users will gradually move away from magic-program

Will update it

> We should consider keeping the codebase for all of the builtins in their own separate crates, instead of putting everything into magicprogram.

I agree, I didn't do it as I considered it a broader refactoring that we can do.

PR #1418, opened 2026-07-13
fix: do not warn about the first false positive blockhash divergence

bmuddha:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved replica-mode block verification during scheduler startup.
  * The first replicated block is now exempt from divergence warnings while startup state is reconstructed.
  * Subsequent replicated blocks continue to be checked, with mismatches reported as errors.
  * The startup exemption is consumed once and is not restored by later mode changes.

* **Documentation**
  * Updated the scheduler security and performance specification to document the replica-mode verification behavior.


GabrielePicco (approved):
LGTM!

PR #1420, opened 2026-07-14
Fix buffer preparation signature aliasing

GabrielePicco:
## Summary

- append the existing intent uniqueness noop to buffer init, realloc, write, and cleanup transactions
- keep cleanup uniqueness strategy-scoped across commit-id recovery
- cover rapid commit-id reuse under a cached blockhash and lock maximum transaction sizes


taco-paco (approved):
Thanks! LGTM

PR #1425, opened 2026-07-15
fix: increase RPC fetch retries to survive RPC lag behind pubsub tip

GabrielePicco:
## Problem

When a subscription update arrives for an executable account (e.g. a program upgrade), the validator re-fetches the program + programdata with `min_context_slot` set to the update slot. The fetch retry budget (3 attempts, 400ms apart, ~0.85s total) may be shorter than the lag of the fetch RPC behind the pubsub/gRPC tip, so the fetch fails and the update is dropped permanently

## Fix

Increase `RPC_FETCH_MAX_RETRIES` from 3 to 15 (retry delay unchanged at 400ms), extending the min-context-slot chase window from ~0.85s to ~6s, which covers the observed RPC-behind-tip lag.


thlorenz (commented):
I will add a follow up to alert when even this timeout fails.

PR #1427, opened 2026-07-15
fix: hardenings against the blockhash divergence

bmuddha:
## Summary by CodeRabbit

- **Reliability**
  - Improved replica resynchronization when restarting and when entering replica mode.
  - Enhanced replay processing and slot transition handling for more consistent state updates.
- **Monitoring**
  - Added new metrics to track replica stream gaps, blockhash divergence, and initial blockhash mismatches.
- **Messaging**
  - Improved replica continuity validation with explicit stream-gap detection.
  - Upgraded broker publishing confirmations and duplicate-message reporting.
  - Replica resume now starts from a more accurate persisted position.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4044,6 +4044,7 @@ dependencies = [
  "magicblock-config",
  "magicblock-core",
  "magicblock-ledger",
+ "magicblock-metrics",
  "notify",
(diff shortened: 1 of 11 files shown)


lucacillario (approved):
LGTM

PR #1492, opened 2026-07-27
fix(committor): propagate tx assembly errors in prepare_for_strategy

### LINKED ISSUE #1018: Transaction preparation can panic in committor workers
## Problem

**Report**
What happened:
In `magicblock-committor-service/src/transaction_preparator/mod.rs` at line 106, there is a `.expect()` call that can cause a runtime panic in committor workers:
```rust
let message = TransactionUtils::assemble_tasks_tx(
    authority,
    &tx_strategy.optimized_tasks,
    self.compute_budget_config.compute_unit_price,
    &lookup_tables,
)
.expect("Possibility to assemble checked above")
.message;
```

Expected:
Error should be propagated via `?` through the existing `PreparatorResult` error path instead of panicking.

Repro:
The lookup tables passed to the second `assemble_tasks_tx` call may differ from the first check at line 86, making the assumption in `.expect()` not strictly guaranteed.

Logs/links:
https://github.com/magicblock-labs/magicblock-validator/blob/master/magicblock-committor-service/src/transaction_preparator/mod.rs#L106

I'd like to work on this fix if no one is assigned yet.

**Commit / version**
master

**Environment**
_No response_

## Reproduction

Not yet documented.

## Expected

The affected flow completes without the described failure.

## Context

Add the affected revision and path when known.

crazywriter1:
## Summary

`TransactionPreparatorImpl::prepare_for_strategy` used `.expect()` when assembling the final transaction message. A prior feasibility check runs against dummy lookup tables, so assembly can still fail with the real prepared data — causing the committor worker to panic.

This change replaces `.expect()` with `?` so assembly failures are propagated as errors instead of panicking.

Fixes #1018

## Test plan

- [x] `cargo check -p magicblock-committor-service`
- [x] `cargo test -p magicblock-committor-service`

## Summary by CodeRabbit

* **Bug Fixes**
  * Transaction preparation now reports message assembly errors gracefully instead of terminating unexpectedly.
  * Improved error propagation allows callers to handle failed transaction preparation safely.

### DIFF
--- a/magicblock-committor-service/src/transaction_preparator/mod.rs
+++ b/magicblock-committor-service/src/transaction_preparator/mod.rs
@@ -101,14 +101,15 @@ impl TransactionPreparator for TransactionPreparatorImpl {
             .await?;
         metrics::observe_committor_intent_alt_count(lookup_tables.len());
 
-        let message = TransactionUtils::assemble_tasks_tx_with_uniqueness_nonce(
-            authority,
-            &tx_strategy.optimized_tasks,
-            self.compute_budget_config.compute_unit_price,
-            &lookup_tables,
-            tx_strategy.uniqueness_nonce,
-        )?
-        .message;
+        let message =
+            TransactionUtils::assemble_tasks_tx_with_uniqueness_nonce(
+                authority,
+                &tx_strategy.optimized_tasks,
+                self.compute_budget_config.compute_unit_price,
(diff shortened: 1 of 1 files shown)


crazywriter1:
Friendly ping. I would appreciate it if you could check it when you have time. @snawaz @GabrielePicco

PR #1494, opened 2026-07-28
perf: arm secondary subscriptions on gRPC first

GabrielePicco:
## Summary

- start read-only fetch-owned secondary watches with `GrpcPreferred` coverage, with full fallback whenever connected gRPC coverage is incomplete
- preserve the strongest pending transport requirement across both fetch-dedup layers and serialize terminal classification, waiter notification, and any downgrade
- make SubMux attach/reconnect publication, same-key settlement, actor cleanup, and reconciliation cancellation-safe
- keep the secondary LRU independently bounded at 20,000 by default so it can be raised without proportional websocket churn while complete gRPC coverage is healthy
- document the transport, reconciliation, and publication invariants

## Validation

- `make fmt`
- `make lint`
- `cargo test -p magicblock-chainlink --all-targets --all-features` — 388 unit tests and all integration targets passed; one pre-existing same-slot redelegation test remains ignored
- `cargo nextest run --workspace --exclude magicblock-aml --exclude magicblock-aperture` — 870 passed, 2 skipped

The local sandbox rejects localhost socket binding for the excluded AML/Aperture cases; GitHub CI covers those suites.

## Summary by CodeRabbit

* **New Features**
  * Full-coverage transport escalation is now applied during deduplicated fetch+clone flows when required.
  * Added more explicit, generation-safe subscription publication and settlement behavior for both account and program subscriptions.
  * Split independent LRU capacity controls for primary vs secondary subscription caches.

* **Bug Fixes**
  * Prevented stale classification/coverage state from changing transport tiers.
  * Improved coordination and preservation of WebSocket vs gRPC watching, including safer reconnect and cleanup after cancellations.

* **Documentation**
  * Expanded guidance on transport coverage rules, eviction/reconciliation semantics, and key invariants.


magicblock-chainlink/src/remote_account_provider/mod.rs
```diff
@@ -595,183 +836,1204 @@ struct SubscriptionTierCtx<U: ChainPubsubClient> {
...
+            self.pubsub_client.finalize_subscription_removal(&pubkey);
+        }
+    }
+
+    async fn prefer_grpc_after_fetch_resolution(
+        &self,
+        pubkey: Pubkey,
+        _subscription_guard: SubscriptionKeyGuard,
+    ) {
+        let pending_requires_full = self
```
thlorenz, line 1373: This check can race with a new non-read pending fetch. `prefer_grpc_after_fetch_resolution` holds the per-key subscription guard, but `try_get_multi` inserts or escalates `fetching_accounts` before it takes that guard, so a transaction/internal requester can set `requires_full_coverage` after this one-time check while this function is awaiting `prefer_grpc_subscription`. 

The downgrade then commits `GrpcPreferred` even though a full-coverage fetch is pending, violating the invariant that any concurrent non-read requester gets full websocket plus gRPC coverage. 

Please either publish the new pending full requirement under the same per-key guard, or re-check `fetching_accounts` after the await and restore/publish `Full` before committing if a full requirement appeared.

thlorenz (approved):
Tons of changes + added complexity.
I reviewed as good as I could, but should we maybe first consider if we want to keep websockets at all before merging this?

All this complexity would fall away if we wouldn't.

GabrielePicco:
Closing as unnecessary

PR #1501, opened 2026-07-28
fix: retry failed intents

taco-paco:
## Summary

Reintroduces retrying failed intents on restart (previously reverted after a bug where a replayed intent could apply stale state if the account had been undelegated/redelegated since the intent was originally scheduled).

Before replaying a failed intent, we now verify:
- The account is still in the same delegation session it was in when the intent was built (hasn't been undelegated/redelegated since).
-  The commit hasn't already landed on chain.

Only intents that pass both checks are retried. Pending intents (still in-flight, most recent) are recovered as before, without these checks.

To support the session check, we now persist the base chain slot observed when an intent was scheduled - previously dropped on reload, defaulting to 0. Includes a DB migration for existing rows.

Note: intent execution durability is a question of a different PR and saga — tracked separately. Here we only try to replay intents that we can and that are safe to replay; everything else is covered by that larger effort.

With this in mind, the nonce comparison was relaxed so it covers both single-stage and two-stage execution. This could lead to an intent being executed twice, but that's fine since the resulting state is the same either way.

## Breaking Changes
- [ ] None

## Summary by CodeRabbit

* **New Features**
  * Improved intent recovery by loading both pending and failed/recoverable commit intents directly from persisted state.
  * Preserved per-account commit identifiers and remote slot data during recovery and rescheduling.
  * Strengthened rescheduling safeguards with delegation-session consistency checks and commit-nonce availability validation.

* **Bug Fixes**
  * Automatically migrates existing databases to store and backfill `remote_slot` for older commit-status rows.
  * Ensures recovered account and commit data round-trip `remote_slot` correctly to prevent scheduling from stale sessions or reused nonces.


magicblock-committor-service/src/persist/commit_persister.rs
```diff
@@ -164,6 +165,23 @@ impl IntentPersisterImpl {
...
+
+    /// Loads pending/failed bundles for restart recovery, each paired with
+    /// the `commit_id` (nonce) its committed accounts were persisted with.
+    /// See [`RecoveredIntent`] for why this isn't part of the
+    /// `IntentPersister::pending_intent_bundles` trait method.
+    pub fn recoverable_intents(
+        &self,
+        min_created_at: u64,
+    ) -> CommitPersistResult<Vec<RecoveredIntent>> {
+        let commits_db = self.commits_db.lock().expect(POISONED_MUTEX_MSG);
```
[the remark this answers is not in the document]
thlorenz: FIne on mutexes.

magicblock-committor-service/src/service.rs
```diff
@@ -371,55 +385,116 @@ where
...
+    /// at the same `remote_slot` recorded when the intent was scheduled.
+    async fn is_same_delegation_session(
+        &self,
+        pubkeys: &[Pubkey],
+        recovered: &RecoveredIntent,
+    ) -> ChainlinkResult<bool> {
+        let recovered_accounts = recovered.bundle.get_all_committed_accounts();
+        let current_accounts = self
+            .chainlink
+            .fetch_accounts(
```
thlorenz, line 466: This does not necessarily fetch current base-layer state. 
For a delegated, non-undelegating account already in the bank, chainlink treats the local copy as a hit and skips the remote fetch. 
Startup also preserves delegated bank accounts. 
After an offline undelegate/redelegate cycle, this can therefore compare the persisted slot with the same stale local slot and approve replay into a new delegation session. 
If the nonce reset, the old account state can then be committed into that new session. 
Please verify the session against current base-layer delegation metadata rather than the protected local account copy.

magicblock-committor-service/src/persist/db.rs
```diff
@@ -366,19 +374,40 @@ impl CommittsDb {
...
+    /// since real base chain slots are never 0.
+    fn migrate_add_remote_slot_column(&self) -> Result<()> {
+        let has_column = self
+            .conn
+            .prepare("SELECT 1 FROM pragma_table_info('commit_status') WHERE name = 'remote_slot'")?
+            .exists([])?;
+        if has_column {
+            return Ok(());
+        }
+        self.conn.execute_batch(
```
thlorenz, line 408: Every pre-upgrade pending row is backfilled with `remote_slot = 0`, but recovery later requires exact equality with the current account's normally nonzero slot. 
That silently drops all pending settlement work present during an upgrade, while the previous implementation would recover it after checking delegation. 
If this is not acceptable then provide a safe legacy reconciliation path, or record an explicit durable failure outcome instead of making these rows unrecoverable.

magicblock-committor-service/src/service.rs
```diff
@@ -371,55 +385,116 @@ where
...
+        &self,
+        pubkeys: &[Pubkey],
+        recovered: &RecoveredIntent,
+    ) -> TaskInfoFetcherResult<bool> {
+        let current_nonces = self
+            .processor
+            .fetch_current_commit_nonces(pubkeys, 0)
+            .await?;
+
+        let valid = recovered.commit_ids.iter().all(|(pubkey, commit_id)| {
```
thlorenz, line 513: A `FailedFinalize` row is written only after the commit stage returned a commit signature, so its persisted commit ID has normally already landed and equals the current on-chain nonce. 
Such a row can never pass this predicate, even though the new query explicitly loads `FailedFinalize` for retry. 
Retrying the full commit would also be wrong after that stage landed. 

We need to distinguish this state and resume only the remaining finalize work, or exclude it from the claimed recovery set with an explicit durable outcome.

magicblock-committor-service/src/service.rs
```diff
@@ -371,55 +385,116 @@ where
...
+    }
+
+    /// False if any committed account's persisted `commit_id` has already
+    /// been consumed on chain, or couldn't be verified.
+    async fn is_valid_nonce(
+        &self,
+        pubkeys: &[Pubkey],
+        recovered: &RecoveredIntent,
+    ) -> TaskInfoFetcherResult<bool> {
+        let current_nonces = self
```
thlorenz, line 508: Using `min_context_slot = 0` allows a lagging RPC response to pass this replay gate. 
The returned nonce is cached, so the later bundle refresh does not make task execution refetch it at the newer account slot. 
That can seed the retry with a stale nonce and produce another avoidable failure. Please fence this read at the verified base-layer slot and preserve that freshness through execution.

thlorenz (commented):
Aside from the remaining coderabbit issues I found a few issues as well.

The recovery path has two blocking safety and durability issues: it can approve stale local delegated state after an offline re-delegation, and the schema migration makes all legacy pending intents unrecoverable. 

Failed-finalize rows also cannot pass the new nonce gate, and nonce validation uses an unfenced RPC read that is then cached. These issues conflict with the base-layer synchronization, restart durability, and commit lifecycle invariants.

magicblock-committor-service/src/service.rs
taco-paco: > For a delegated, non-undelegating account already in the bank, chainlink treats the local copy as a hit and skips the remote fetch.

That what we want. If it is in the bank and **delegated** it is owned by ER and `remote_slot` allows us to check delegation "session". If `intent.remote_slot == current.remote_slot` it is same session. If not(higher likely) it means account was undelegated and then delegated.

> After an offline undelegate/redelegate cycle, this can therefore compare the persisted slot with the same stale local slot and approve replay into a new delegation session.

1. If account is delegated it's remote_slot is not stale. 
2. It can't approve replay into new session as we compare remote_slots and distinguish sessions using it

For `!delegated` or `undelegating`chainlink will fetch new data if as we want.

magicblock-committor-service/src/persist/db.rs
taco-paco: We can't replay intents we're not sure about. 
Atm we have: 
1. No failed intents on nodes
2. Failed intents can be resolved via script and then ignoreing here is safe

All new failed intent until `Outbox` is merged will  have correctly identified `remote_slot`

magicblock-committor-service/src/service.rs
taco-paco: Fixed

taco-paco:
@thlorenz 
Note: durability is not a purpose of this PR and is barely achievable by current Persister architecture. Durability is a separate ongoing saga that is implemented this [PR](https://github.com/magicblock-labs/magicblock-validator/pull/1368). Here we just try to retry intents that we 100% sure about. Durability is desirable but is out of scope of this PR

> it can approve stale local delegated state after an offline re-delegation

Delegated accounts can't be re-delegated offline as they have to be undelegated first. More detailed answer [here](https://github.com/magicblock-labs/magicblock-validator/pull/1501#discussion_r3675503296).

> and the schema migration makes all legacy pending intents unrecoverable.

Legacy intents will be addressed via script and post migration will be recovered via current flow.

>nonce validation uses an unfenced RPC read that is then cached

Fixed

> Failed-finalize rows also cannot pass the new nonce gate

Majority of intents are executed in 1 tx for which this isn't true. For example on devnet-asia: 143 single stage intents an 0 are executed in 2 txs. Tho I will think more on how that can be addressed

magicblock-committor-service/src/persist/db.rs
```diff
@@ -585,6 +615,35 @@ impl CommittsDb {
...
+        );
+        let stmt = &mut self.conn.prepare(&query)?;
+        let mut rows = stmt.query(params![
+            "FailedFinalize",
+            "FailedProcess",
+            u64_into_i64(min_created_at)
+        ])?;
+
+        extract_committor_rows(&mut rows)
+    }
```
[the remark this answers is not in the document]
taco-paco: Fine in test

magicblock-committor-service/src/service.rs
taco-paco: Note, intent execution durability is question of a different PR and saga [here](https://github.com/magicblock-labs/magicblock-validator/pull/1368). Here we try to replay intents that we can and that are safe to do. Everything else will be covered in big PR mentioned above.

With this said, relaxed nonce comparison so it would cover both SingleStage and TwoStage. That could lead to intent executed twice but it is fine as states are the same

magicblock-committor-service/src/persist/commit_persister.rs
```diff
@@ -164,6 +165,23 @@ impl IntentPersisterImpl {
...
+        let mut rows =
+            commits_db.get_pending_commit_statuses(min_created_at)?;
+        rows.extend(commits_db.get_failed_commit_statuses(min_created_at)?);
+        drop(commits_db);
+
+        Ok(pending_rows_to_recovered_intents(rows, min_created_at))
+    }
```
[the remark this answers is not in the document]
taco-paco: Fixed

magicblock-committor-service/src/service.rs
```diff
@@ -371,55 +385,116 @@ where
...
+            .processor
+            .fetch_current_commit_nonces(pubkeys, 0)
+            .await?;
+
+        let valid = recovered.commit_ids.iter().all(|(pubkey, commit_id)| {
+            current_nonces
+                .get(pubkey)
+                .is_some_and(|current_nonce| commit_id > current_nonce)
         });
```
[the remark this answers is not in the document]
taco-paco: This is the best effort safe retry, If it doesn;t fall into conditions it is more safe to abandon and manually resolve than commit an old state, Durable execution is question of different PR

GabrielePicco (approved):

PR #1551, opened 2026-08-13
ci: always prepare releases from master

GabrielePicco:
## Summary

The prepare-release workflow targeted `github.event.repository.default_branch` (currently `dev`) for the checkout, the release-PR base, and the dispatch gate. Releases should only and always be created from `master`, so this pins the workflow to `master`:

- The release branch is always cut from `master` (`ref: master` in the checkout).
- The release PR always targets `master` (`BASE_BRANCH: master`).
- The `github.ref_name == default_branch` gate is removed: dispatching from any branch (including the UI default `dev`) now works and still releases from `master`, instead of silently skipping.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -19,16 +19,16 @@ permissions:
 jobs:
   prepare-release:
     name: Prepare release PR
-    if: github.ref_name == github.event.repository.default_branch
     runs-on: blacksmith-32vcpu-ubuntu-2404
     env:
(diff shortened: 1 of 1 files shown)


bmuddha (approved):
LGTM, but may be targetting master for this PR is also a good idea, dev will inherit the change after merge

GabrielePicco:
> LGTM, but may be targetting master for this PR is also a good idea, dev will inherit the change after merge

This is actually OK as the actions use the default branch (now dev) by default

PR #1574, opened 2026-08-16
fix(chainlink): return error on lagging fetch response slot

crazywriter1:
## Summary
- After the `min_context_slot` retry loop in `RemoteAccountProvider::fetch`, a lagging response slot used `assert!`, which panics in release builds
- Sibling path `fetch_multi_rpc_only` already returns `Err` for the same condition
- Replace the assert with `notify_error` + `return` so waiters get `AccountResolutionsFailed` instead of crashing the process
- Resolves the open TODO: retry is already handled above; remaining lag is treated as an error

## Test plan
- [x] `cargo +nightly fmt --check -p magicblock-chainlink -- --config-path rustfmt-nightly.toml`
- [x] `cargo test -p magicblock-chainlink remote_account_provider`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fetch reliability when responses are older than the required context slot.
  * Waiting requests now receive a clear fetch error instead of triggering an application crash.

### DIFF
--- a/magicblock-chainlink/src/remote_account_provider/mod.rs
+++ b/magicblock-chainlink/src/remote_account_provider/mod.rs
@@ -4132,9 +4132,19 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
                 };
             };
 
-            // TODO: should we retry if not or respond with an error?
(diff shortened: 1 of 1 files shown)


crazywriter1:
@coderabbitai review

GabrielePicco (approved):

PR #1583, opened 2026-08-19
fix: backstop program upgrade detection with periodic programdata sweep

GabrielePicco:
## Problem

The persistent programdata watches from #1529 detect LoaderV3 upgrades only while their pubsub subscription is alive. A subscription can silently die upstream while local bookkeeping keeps it (ghost — e.g. lost across a provider reconnect), after which **no notification ever arrives again** and nothing detects or repairs the loss:

- `watch_programdata` returns `AlreadyInstalled` on a bare `programdata_index` hit, so re-clones (including read-forced ones) never re-subscribe.
- All upgrade checks were notification-driven: zero notifications ⇒ zero checks, forever.

## Fix

**Periodic transport-independent sweep** (`sweep_programdata_watches`, every 5 min): fetch a 12-byte programdata prefix (state tag + deploy slot) per watched program over RPC and route the program into the existing `handle_executable_sub_update` reload path when

```
deploy_slot > bank_program.remote_slot()
```

The bank copy was fetched at `remote_slot`, so it already reflects every deploy up to that slot — the criterion needs no stored baseline and cannot false-positive into spurious re-clones. Heals every ghost cause since it never touches pubsub. Bounded against unhealthy RPC endpoints by a rotating start cursor (no tail starvation) and a consecutive-failure breaker. Cost: one 12-byte `dataSlice` fetch per watched program per interval; detection latency for ghosted watches drops from ∞ to ≤ ~6 min.

**Watch-lifecycle race closed** (found during adversarial review): the notification-side bank-miss cleanup could pop the watch while a concurrent clone that had seen `AlreadyInstalled` was in flight, leaving the materialized program permanently unwatched. The cleanup now re-checks the bank under the index lock before popping, and the clone success path re-asserts the watch after materialization — every interleaving now ends with a materialized program holding a watch.

Known scope limit (documented): authority-only rotations keep the deploy slot and remain covered only by the live-subscription path; the sweep targets missed bytecode upgrades.

## Testing

- New test: upgrade with no notification delivered — the sweep alone must reload the program; sweeps are no-ops before the upgrade and after the reload; programdata is not cloned into the bank.
- Full `magicblock-chainlink` suite: 394 passed, including both #1529 regression tests. Clippy clean.
- Live-validated the diagnosis end-to-end on devnet-tee via probe-program upgrade experiments (fresh watches detect within 1 slot; deferred-verify burst path works).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection and handling of program upgrades, including upgrades that occur without notifications.
  * Preserved upgrade monitoring when program cloning is skipped or temporarily fails.
  * Cleaned up outdated monitoring state when programs are removed or evicted.
* **Reliability**
  * Added bounded, automatically maintained monitoring for program data.
  * Improved executable reload behavior after detected updates.
  * Expanded coverage for delegation, collision, concurrency, and upgrade scenarios.

### DIFF
--- a/magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
+++ b/magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
@@ -1164,8 +1164,9 @@ where
     }
 
     /// Holds a persistent programdata subscription for a LoaderV3 program so
-    /// upgrades stay observable; both pubkeys are preferred onto gRPC
-    /// transport. Installed before the clone so an upgrade notification
+    /// upgrades stay observable; it keeps full coverage on every transport
+    /// leg, while the program account is preferred onto gRPC.
+    /// Installed before the clone so an upgrade notification
     /// landing mid-clone already routes; held once per program (idempotent
     /// across reloads). Failing to hold the subscription is an error but
     /// non-fatal to the clone: the watch is retried on the next load.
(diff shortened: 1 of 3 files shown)


GabrielePicco:
Added a second commit: programdata watches now keep dual-transport coverage (WS + gRPC) instead of being narrowed to a single gRPC leg — `prefer_grpc_subscription` is kept only for the program account (pre-Agave-4.2 chains notify hot program accounts every slot, which would flood websocket legs; programdata only changes on deploys, so full coverage costs one sub slot per WS leg and ~zero traffic). With this, losing the fast notification path requires every transport to fail at once, and the sweep remains the transport-independent floor.

bmuddha (approved):

bzawisto (approved):

PR #1584, opened 2026-08-19
fix: detect silently dead gRPC streams via per-stream liveness

GabrielePicco:
Follow-up to #1583 — the root-cause fix for the transport layer that ghosted the programdata watch in the first place.

## Problem

The laser actor's liveness watchdog kept **one actor-wide** last-activity timestamp. Every stream carries a slot-update filter (an update ~every chain slot), so any live stream kept resetting the watchdog while a sibling stream sat dead — an h2 half-open wedge or a failed SDK-internal resume kills a single stream without erroring, and the actor deliberately leaves reconnectable statuses to the SDK. The dead stream's accounts silently lost every notification, and no repair path could help: resubscribe attempts dedupe against the client's own `is_subscribed()` bookkeeping, which still said subscribed. This is the mechanism behind the devnet-tee incident where a program missed every upgrade for three days (probe subscriptions worked instantly on the same node because they opened fresh streams).

## Fix

Track last activity **per stream** in the `StreamManager`:

- Recorded in `next_update` for **successful** items only — a stream stuck in an SDK retry loop emitting statuses must not keep resetting the watchdog.
- Seeded with a **startup grace** (30s) at stream insertion: SDK connection setup and upstream activation/backfill legitimately delay the first update past the steady-state 5s window, and a spurious trip would tear down healthy siblings.
- **Carried over on promotion** (CurrentNew → UnoptimizedOld relocates the same live stream), so a stream already going quiet cannot reset its age by moving.
- A handle whose stream is **missing from the map** ended silently — tokio's `StreamMap` reaps finished children during polling with no notification while siblings remain — and is reported as infinite silence (expected-topology check).

The actor's liveness tick reconnects when **any** stream is silent past the window. Recovery is unchanged and force-resubscribes by construction: `signal_connection_issue` clears the manager, so the SubMux reconnect restores every subscription from the authoritative tracker with no already-subscribed dedupe in the way. Detection of the incident class goes from *never* to ≤7s (5s window + 2s check interval).

## Testing

Five new StreamManager tests: silent sibling domination (the exact masking bug), silently-ended stream detection, startup grace, error items not counting as liveness, and the no-streams idle case. Full chainlink suite: 399 passed. The ended-stream test polls with no update queued so the `StreamMap`'s randomized poll order cannot make it flaky.

Stacked on #1583: that PR's sweep remains the transport-independent floor (≤5 min); this PR restores the fast path (~1 slot) by making single-stream death detectable and recoverable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection liveness monitoring by tracking activity independently for each stream.
  * Reconnects are now triggered when any active stream becomes silent beyond the five-second threshold.
  * Prevented idle actors without active streams from being incorrectly disconnected.
  * Added startup grace handling and ignored failed updates when determining stream activity.
  * Improved handling of closed, missing, and dropped streams.

### DIFF
--- a/magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
+++ b/magicblock-chainlink/src/remote_account_provider/chain_laser_actor/actor.rs
@@ -25,7 +25,7 @@ use solana_pubkey::Pubkey;
 use solana_sdk_ids::sysvar::clock;
 use tokio::{
     sync::{mpsc, oneshot},
-    time::{interval, Instant, MissedTickBehavior},
+    time::{interval, MissedTickBehavior},
 };
 use tonic::Code;
 use tracing::*;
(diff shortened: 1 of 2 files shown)


bmuddha (approved):
LGTM

bzawisto (approved):

Issue #878, opened 2026-01-22
Add realloc instructions for data growth over 10 KB


snawaz:
## Goal

Add realloc instructions for data growth over 10 KB.

## Why

The current behavior or architecture does not provide this outcome.

## Scope

**Request**
Solana does not allow account reallocation of more than ~10KB in a single instruction.

So if: 

```
(committed_account_datalen - base_account_datalen) > 10KB
```

then all commit instruction variants will fail due to the runtime realloc limit.

To handle this scenario, the committor-service should detect large data growth and issue one or more dedicated realloc instructions before sending the commit instruction. This keeps commit instructions within Solana’s realloc constraints and avoids runtime failures!

**It's a low-priority issue.**

PR #1588, opened 2026-08-21
hotfix: restrict large commits

taco-paco:
## What changed

Here we prohibit scheduling commit of account that is larger than 10_240 threshold. 
TODO: This restriction will be removed once #878 will be resolved.

We consider this to be non-breaking(likely) change for users as on mainnet currently there're 0 delegated accounts larger than 10_240 cap on mainnet.

Closes #ISSUE

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented scheduling or committing accounts that exceed the permitted data-growth limit.
  * Oversized accounts now fail with an invalid account data error instead of being processed.

* **Tests**
  * Added coverage confirming oversized accounts are rejected.
  * Added coverage confirming accounts exactly at the permitted limit succeed.

### DIFF
--- a/programs/magicblock/src/schedule_transactions/process_schedule_commit.rs
+++ b/programs/magicblock/src/schedule_transactions/process_schedule_commit.rs
@@ -6,6 +6,7 @@ use magicblock_core::intent::{
 };
 // no direct token remap helpers needed here; handled in CommittedAccount builder
 use solana_account::{ReadableAccount, WritableAccount};
+use solana_account_info::MAX_PERMITTED_DATA_INCREASE;
 use solana_instruction::error::InstructionError;
 use solana_log_collector::ic_msg;
 use solana_program_runtime::invoke_context::InvokeContext;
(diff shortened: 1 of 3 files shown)


programs/magicblock/src/schedule_transactions/process_schedule_commit.rs
```diff
@@ -169,6 +170,19 @@ pub(crate) fn process_schedule_commit(
...
+            > MAX_PERMITTED_DATA_INCREASE
+        {
+            ic_msg!(
+                invoke_context,
+                "ScheduleCommit ERR: account {} is too large to be committed",
+                acc_pubkey
+            );
+            return Err(InstructionError::InvalidAccountData);
+        }
+
```
snawaz, line 185: Do we need this restriction as it fails anyway (though at later stage)? Since the next set of PRs will actually fix it, we'll have to undo the changes in this PR.

taco-paco: Yes, if we know we fail, we shouldn't allow scheduling an commit at the first place. 
The contract with regards to Commit/Undelegate - if it got scheduled we are obliged to execute it, if we 
physically can't we shouldn;t make such promises, hence failing tx.

Later this shall be removed but only once this case is addressed on dlp level

snawaz (approved):
Looks good to me.

PR #1589, opened 2026-08-21
feat: Magic ATA support (v2)

### LINKED ISSUE #1575: Support rent-pending ATA materialization in the validator

GabrielePicco:
## What

Adds Magic ATA support to the validator: SPL associated token accounts can be created directly inside the ER for destinations whose base-layer eATA has not been materialized yet, marked with a `SysvarRent` close-authority sentinel. A Magic ATA exists and behaves as a normal token account inside the ER only. This replaces eager base-layer materialization for private transfer destinations and enables the encrypted-destination shuttle flow in the ephemeral SPL token program.

A funded Magic ATA takes precedence over a base-layer delegation of the same ATA. The two merge only once the Magic ATA is drained to zero through the shuttle flow, after which the eATA projection replaces it.

## Key changes

- **Magic program**: new `CreateMagicAta` and `CloseMagicAta` instructions (`process_create_magic_ata.rs`, `process_close_magic_ata.rs`). Create is idempotent and sponsor-payable; close is a no-op unless the ATA matches the Magic ATA marker for the signing owner and holds zero tokens, so it can be appended unconditionally to withdrawal flows.
- **Chainlink cloner**: `ata_projection.rs` recognizes Magic ATAs so cloning/projection does not clobber ER-created destination accounts, and logs when a funded Magic ATA blocks a base projection.
- **Processor**: executor support for the new instructions and scheduled-commit handling for Magic ATAs.
- **Core**: token-program helpers for building/inspecting Magic ATAs.

## Testing

- New integration suite `test-cloning/tests/11_magic_ata.rs` covering create/fund/close, idempotency, and the shuttle merge-to-encrypted-destination path end to end.
- Updated `ephemeral_token_program.so` fixture (built from the matching e-token branch).
- Existing `ephemeral_accounts` and schedule-commit tests extended.

## Summary by CodeRabbit

- **New Features**
  - Added support for creating Magic ATAs for SPL Token and Token-2022 assets.
  - Added instructions to create and close Magic ATAs.
  - Added support for closing drained Magic ATAs and withdrawing their assets.
  - Added handling for supported Token-2022 extensions and default account states.

- **Bug Fixes**
  - Drained Magic ATAs can now be replaced by projected accounts without blocking synchronization.
  - Projections preserve the validated ATA layout when available.
  - Funded or undelegating Magic ATAs remain protected from replacement.
  - Unfunded Magic ATA creation is rejected and rolled back.
  - Magic ATAs cannot be committed or undelegated through unsupported flows.


GabrielePicco:
Paired with [magicblock-labs/ephemeral-spl-token#127](https://github.com/magicblock-labs/ephemeral-spl-token/pull/127), which consumes `CreateRentPendingAta`/`CloseRentPendingAta` from this PR. The `ephemeral_token_program.so` fixture here is built from that branch, so the two should land together.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,660 @@
...
+    initialize_token_data(
+        acc.data_as_mut_slice(),
+        &token_program,
+        wallet_owner,
+        mint,
+        &token_account_shape.required_extensions,
+        token_account_shape.initial_state,
+    )?;
+    acc.set_remote_slot(0);
+    acc.set_delegated(true);
```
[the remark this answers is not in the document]
GabrielePicco: The delegated flag here is the ER's local-writability mechanism, not a claim of base-layer delegation, and the danger §1.7 guards against — committing unbacked state — is structurally prevented: `process_schedule_commit.rs` rejects any rent-pending ATA with `InvalidAccountData` before it can be committed or undelegated ("account {} is a rent-pending ATA and cannot be committed or undelegated; use the shuttle withdrawal flow"). So such an account can never commit to a base ATA or to a non-existent eATA.

The lifecycle is also validator-controlled rather than user-forged: only the Magic program creates these accounts, it stamps the rent-sysvar close-authority sentinel, and the transaction is rolled back unless the account ends funded. Funds leave through the shuttle withdrawal flow, which materializes the eATA at that point. Happy to add an explicit carve-out to §1.7 in a follow-up so the invariant text names this case.

magicblock-processor/src/executor/processing.rs
```diff
@@ -467,6 +468,35 @@ impl super::TransactionExecutor {
...
+        {
+            let info = txn
+                .accounts
+                .iter()
+                .find(|(key, _)| key == &pubkey)
+                .and_then(|(_, acc)| {
+                    try_get_rent_pending_ata_info(&pubkey, acc)
+                });
+            match info {
+                Some(info) if info.amount > 0 => {}
```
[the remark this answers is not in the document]
GabrielePicco: Worth stating the trade-off explicitly rather than closing it silently. These accounts are ER-local: they are never committed to base (`process_schedule_commit.rs` rejects them outright), so this is bounded to a single rollup's state, not persistent base-layer AccountsDb growth. Creation also is not free to trigger — each one needs a signed transaction with its fees plus a real token transfer, since the transaction is rolled back unless the ATA ends with a positive balance, and `CloseRentPendingAta` reclaims a drained account.

What remains true is that a mint owner can pay to create many cheap ER accounts. That is the same surface any ER write has, and an ER is single-tenant and permissioned by its operator, so rate limiting belongs at the RPC/ACL layer rather than by charging rent here, which would defeat the point of the feature (destinations that do not yet have materialized base state). Tracking as a known trade-off rather than redesigning in this PR.

test-integration/test-cloning/tests/11_rent_pending_ata.rs
```diff
@@ -0,0 +1,808 @@
...
+use spl_associated_token_account_interface::instruction::create_associated_token_account_idempotent;
+use spl_token::{instruction as spl_token_ix, state::Mint};
+use test_kit::init_logger;
+
+const SOURCE_EATA_BALANCE: u64 = 200;
+const RECEIVE_AMOUNT: u64 = 60;
+
+const MAGIC_PROGRAM_ID: Pubkey =
+    Pubkey::from_str_const("Magic11111111111111111111111111111111111111");
+const CLOSE_RENT_PENDING_ATA_DISCRIMINATOR: u32 = 26;
```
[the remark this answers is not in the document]
GabrielePicco: Fixed — the test now imports `ID as MAGIC_PROGRAM_ID` and `MagicBlockInstruction` from `magicblock-magic-program-api` (close data is built with `MagicBlockInstruction::CloseRentPendingAta.try_to_vec()`) and `RENT_PENDING_ATA_CLOSE_AUTHORITY` from `magicblock-core`, so a change to any of those breaks the build instead of silently exercising the wrong instruction.

test-integration/test-cloning/tests/11_rent_pending_ata.rs
```diff
@@ -0,0 +1,808 @@
...
+        .unwrap()
+        .get_account_with_commitment(account, CommitmentConfig::confirmed())
+        .unwrap()
+        .value
+        .is_some_and(|account| {
+            account.data.len() >= 165
+                && account.data[129..133] == 1u32.to_le_bytes()
+                && account.data[133..165] == RENT_SYSVAR_ID.to_bytes()
+        })
+}
```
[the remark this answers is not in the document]
GabrielePicco: Fixed in the same pass — the marker check now compares against `magicblock_core::token_programs::RENT_PENDING_ATA_CLOSE_AUTHORITY` instead of a locally re-declared rent sysvar id. The byte offsets stay inline deliberately: the test asserts the on-wire SPL layout a client would see, so it should fail if that layout ever shifts.

test-integration/test-cloning/tests/11_rent_pending_ata.rs
```diff
@@ -0,0 +1,808 @@
...
+        mint,
+    );
+    let mut tx = Transaction::new_with_payer(&[ix], Some(&ephem_payer.pubkey()));
+    assert!(
+        ctx.send_and_confirm_transaction_ephem(&mut tx, &[&ephem_payer])
+            .map(|(_, confirmed)| confirmed)
+            .unwrap_or(false)
+            == false,
+        "unfunded rent-pending ATA creation must fail"
+    );
```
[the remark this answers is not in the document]
GabrielePicco: Already fixed — this is now `assert!(!ctx.send_and_confirm_transaction_ephem(...).map(...).unwrap_or(false), ...)`. It was the `clippy::bool_comparison` failure in CI; lint is green.

test-integration/test-cloning/tests/11_rent_pending_ata.rs
```diff
@@ -0,0 +1,808 @@
...
+        Some(10_000_000),
+        Some(255),
+    ));
+    let mut setup_tx =
+        Transaction::new_with_payer(&setup_ixs, Some(&fee_payer.pubkey()));
+    let (_sig, confirmed) = ctx
+        .send_and_confirm_transaction_chain(&mut setup_tx, &[fee_payer])
+        .unwrap();
+    assert!(confirmed, "withdrawal setup transaction failed");
+}
```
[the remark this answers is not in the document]
GabrielePicco: Not changing this. Each test in the suite uses its own keypairs, mint and shuttle id, so there is no shared mutable state to serialize on, and the file follows the same pattern as the other test-cloning suites, which the runner already schedules per-suite. CI is green on the full integration job.

magicblock-processor/src/executor/processing.rs
```diff
@@ -467,6 +468,35 @@ impl super::TransactionExecutor {
...
         }
+
+        if executed.execution_details.status.is_err() {
+            return;
+        }
+
+        // Anti-spam invariant: a rent-pending ATA must end its creation
+        // transaction funded, otherwise the whole transaction is rolled back.
+        while let Some(pubkey) =
+            ExecutionTlsStash::pop_newly_created_rent_pending_ata()
```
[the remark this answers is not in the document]
GabrielePicco: Good catch — this was real and is now fixed. Post-delegation actions do execute with the validator as fee payer, so the privileged early return skipped the funding invariant, and a crafted action could have created an unfunded rent-pending ATA.

`verify_account_states` now scopes the privileged bypass to the confined-account check only; the rent-pending funding invariant runs for every transaction. The legitimate flows are unaffected because they fund in the same transaction (instruction 34 creates then transfers, instruction 35 is paired with an SPL transfer), and the cloner never registers these accounts — only the Magic program's `CreateRentPendingAta` does.

Added `test_create_rent_pending_ata_zero_balance_rolls_back_for_privileged_payer`, which marks the fee payer privileged and asserts the rollback; I confirmed it fails against the old code and passes against the fix.

magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
```diff
@@ -714,3 +714,104 @@ where
...
+/// freshly delegated eATA.
+fn base_ata_blocks_projection(
+    ata_pubkey: &Pubkey,
+    base_ata: &AccountSharedData,
+) -> bool {
+    let is_drained_rent_pending_ata =
+        try_get_rent_pending_ata_info(ata_pubkey, base_ata)
+            .is_some_and(|info| info.amount == 0);
+    base_ata.undelegating()
+        || (base_ata.delegated() && !is_drained_rent_pending_ata)
```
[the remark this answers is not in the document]
GabrielePicco: Confirmed and fixed — the trace was right. `base_ata_blocks_projection` deliberately lets a drained rent-pending ATA be replaced, but the account is still marked delegated until it is closed, so `local_delegated_clone_target_active` treated it as live delegated state and deduplicated the replacement projection away, leaving the stale zero balance.

Extracted the predicate as `is_drained_rent_pending_ata` in `ata_projection.rs` and used it in both places, so the dedup guard now exempts drained rent-pending targets while funded ones, undelegating ones and ordinary delegated ATAs keep blocking as before. Added `drained_rent_pending_ata_is_not_live_delegated_state` next to the existing guard tests; the chainlink suite is green.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -2028,10 +2028,18 @@ where
...
+    /// would hide later eATA deposits.
     fn local_delegated_clone_target_active(&self, pubkey: Pubkey) -> bool {
         self.accounts_bank
             .get_account(&pubkey)
-            .is_some_and(|account| account.delegated())
+            .is_some_and(|account| {
+                account.delegated()
+                    && !ata_projection::is_drained_rent_pending_ata(
+                        &pubkey, &account,
+                    )
```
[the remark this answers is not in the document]
GabrielePicco: Right again, and fixed — the dedup exemption alone was not enough. `validate_mutable` -> `validate_not_delegated` rejects any `delegated && !undelegating` destination, which a drained rent-pending ATA is, so the replacement clone failed with `AccountIsDelegated`.

`validate_mutable` now returns early for a drained rent-pending ATA, so a clone can replace the placeholder. The exemption is scoped to the clone paths only — `mutate_accounts` still goes through `validate_not_delegated` unchanged — and funded rent-pending ATAs still fall through and block, so a balance is never clobbered. `CloneAccount` remains validator-authority gated.

Added `test_clone_account_replaces_drained_rent_pending_ata`, which builds the marked zero-balance SPL layout and asserts the clone now succeeds; verified it fails with `AccountIsDelegated` against the code without this change. Full `magicblock-program` suite is green (105 tests), plus fmt and clippy.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,660 @@
...
+        && account.owner() == &system_program::ID
+        && account.data().is_empty()
+}
+
+fn is_native_mint(mint: &Pubkey, token_program: &Pubkey) -> bool {
+    (*token_program == TOKEN_PROGRAM_ID
+        && *mint == spl_token::native_mint::id())
+        || (*token_program == TOKEN_2022_PROGRAM_ID
+            && *mint == spl_token_2022::native_mint::id())
+}
```
snawaz, line 155: I think token-program comparison isn't required as the native_mint pubkeys are static values.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,660 @@
...
+            &wallet_owner,
+            &mint,
+            &token_program,
+        ) || is_matching_existing_projected_ata(
+            &ata_pubkey,
+            &ata_shared,
+            &wallet_owner,
+            &mint,
+            &token_program,
+        ) {
```
snawaz, line 97: I'm not sure if we should use `is_matching_existing_projected_ata` to detect idempotency. It seems to confuse the semantic of the instruction the PR is implementing.

magicblock-core/src/token_programs.rs
```diff
@@ -26,9 +27,34 @@ pub const ASSOCIATED_TOKEN_PROGRAM_ID: Pubkey =
 pub const EATA_PROGRAM_ID: Pubkey =
     pubkey!("SPLxh1LVZzEkX99H6rqYizhytLWPZVV296zyYDPagv2");
 
+// Marker written into a rent-pending ATA's close authority. It is a data
+// marker only, never a signer grant: the sysvar cannot sign, so these
+// accounts are closed exclusively via the Magic Program.
+pub const RENT_PENDING_ATA_CLOSE_AUTHORITY: Pubkey =
+    solana_program::sysvar::rent::ID;
+
```
snawaz, line 35: This looks hacky to me. If it can be anything, then we should use a dedicated ID.  It'll make things friendlier, including even the logs on the explorer as well as node logs.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,660 @@
...
+    mint_data: &[u8],
+    token_program: &Pubkey,
+) -> Result<TokenAccountShape, InstructionError> {
+    if *token_program == TOKEN_PROGRAM_ID {
+        SplMint::unpack(mint_data)
+            .map_err(|_| InstructionError::InvalidAccountData)?;
+        Ok(TokenAccountShape {
+            len: SplAccount::LEN,
+            required_extensions: Vec::new(),
+            initial_state: Token2022AccountState::Initialized,
```
snawaz, line 167: `initial_state: Token2022AccountState::Initialized` looks contradictory, even though it is fine. Please do not create the alias `Token2022AccountState` and use `AccountState` as it is.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
```diff
@@ -0,0 +1,660 @@
...
+    required_extensions: &[ExtensionType],
+    initial_state: Token2022AccountState,
+) -> Result<(), InstructionError> {
+    if *token_program == TOKEN_PROGRAM_ID {
+        let account = SplAccount {
+            mint,
+            owner: wallet_owner,
+            amount: 0,
+            delegate: COption::None,
+            state: SplAccountState::Initialized,
```
snawaz, line 234: No need to hardcode: `state: SplAccountState::Initialized`. The passed argument can be used here after we remove the alias `Token2022AccountState` and use `AccountState` directly (as noted in the previous comment).

programs/magicblock/src/clone_account/common.rs
```diff
@@ -118,6 +119,14 @@ pub fn validate_mutable(
...
         return Err(MagicBlockProgramError::AccountIsEphemeral.into());
     }
+    // A drained rent-pending ATA holds no value and no base delegation: a
+    // clone may replace it with the real eATA projection. Funded ones still
+    // block, so a balance is never clobbered.
+    if try_get_rent_pending_ata_info(pubkey, &account.to_account_shared_data())
+        .is_some_and(|info| info.amount == 0)
+    {
+        return Ok(());
+    }
```
snawaz, line 129: This function name is so misleading: what the name says and what the code does are almost opposite, e.g `ephemeral` accounts are supposed to be mutable, yet it is causing error _"cannot be mutated"_.

While the function of the form `validate_<object-or-entity>` are good, the function of the form `validate_<characteristic>` are often confusing because it does not say whether the characteristic must be _present_ (required) or _absent_ (not required).

magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
```diff
@@ -239,7 +239,7 @@ where
         }
     };
 
-    if base_ata.delegated() || base_ata.undelegating() {
+    if base_ata_blocks_projection(&ata_pubkey, &base_ata) {
```
snawaz, line 242: I wonder: why do we call `base_ata`, not simply `ata`? `base_` and then `delegated()` makes the code contradictory/confusing?

snawaz (changes requested):
Overall, looks good to me.

One question:

- If ER has a rent-pending ATA with non-zero balance and then the same ATA is created on the base chain and delegated to the ER, then what will happen? and how do we handle this case? **It seems the delegation itself will succeed but cloning in the ER will fail?** If so, then the users will be clueless as to why their code in the ER isn't working.

---

Apart from that, we must rename this feature to something more apt and less confusing:

Currently, _rent-pending_ is misleading because it describes the base-chain materialization concern, a concern we do not intend to support as a part of this functionality. Also, the name feels like.. the functionality is not fully functional and something yet to be done whereas the fact is that it is **fully functional** and a **complete** functionality in itself (and there is [nothing pending][1] which will ever be completed):

- **an ATA that exists and behaves normally inside the magic validator only.**

My recommendation:

- **Magic ATA** &mdash; Magic-local ATA is more precise but `local` probably is implied anyway if we cannot easily interpret _Magic_ in any other. So `Magic ATA` seems reasonable to me.
 

[1]: https://github.com/magicblock-labs/magicblock-validator/pull/1381

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
GabrielePicco: Agreed, the two native mint pubkeys are distinct static values, so the token-program comparison was redundant. Simplified to a two-way pubkey check in 92a5059bc5159a116747232305cbb83318bb1c18.

programs/magicblock/src/ephemeral_accounts/process_create_magic_ata.rs
```diff
@@ -0,0 +1,660 @@
...
+            &wallet_owner,
+            &mint,
+            &token_program,
+        ) || is_matching_existing_projected_ata(
+            &ata_pubkey,
+            &ata_shared,
+            &wallet_owner,
+            &mint,
+            &token_program,
+        ) {
```
GabrielePicco: Kept as is, with the reasoning: the e-token `EnsureRentPendingDestination` instruction CPIs this create unconditionally before every transfer, and the destination may already be a normal ATA projected from a delegated eATA. Treating that case as success is what keeps transfers to existing recipients working. The contract of the instruction is "ensure a usable ATA exists locally for (owner, mint)": it creates a Magic ATA when the address is empty, and is a no-op when a matching Magic ATA or a matching projected ATA is already there. Anything else at the address is rejected.

magicblock-core/src/token_programs.rs
GabrielePicco: Kept the rent sysvar for now. It is intentionally a pure data marker: it can never sign, so a Magic ATA is closeable only through the Magic Program. The constant is now `MAGIC_ATA_CLOSE_AUTHORITY` after the rename, which should make the intent clearer at the use sites. If explorer readability turns out to matter, switching to a dedicated Magic-derived key is a validator-only change since neither e-token nor the SDK inspect the sentinel.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
GabrielePicco: Done in 92a5059bc5159a116747232305cbb83318bb1c18: the alias is gone and `AccountState` is used directly.

programs/magicblock/src/ephemeral_accounts/process_create_rent_pending_ata.rs
GabrielePicco: Done in 92a5059bc5159a116747232305cbb83318bb1c18: the SPL branch now carries the passed `initial_state` over. The two token crates define separate `AccountState` enums with the same u8 layout, so it goes through the primitive.

programs/magicblock/src/clone_account/common.rs
GabrielePicco: Fair point. Renamed to `validate_clone_target` in 92a5059bc5159a116747232305cbb83318bb1c18: it names the entity and matches what it guards, a clone may overwrite the account only if it is neither ephemeral nor actively delegated.

magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
GabrielePicco: You are right, and this PR made the contradiction real since the account can now be a locally created Magic ATA with no base counterpart. Renamed to `ata` (`fetch_remote_ata`, `ata_blocks_projection`) in 92a5059bc5159a116747232305cbb83318bb1c18.

GabrielePicco:
Thanks for the review. Addressed in 92a5059bc5159a116747232305cbb83318bb1c18.

**Renamed to Magic ATA** across the validator: instructions are now `CreateMagicAta` / `CloseMagicAta`, with `MagicAtaInfo`, `MAGIC_ATA_CLOSE_AUTHORITY`, the processor files, the integration suite (`11_magic_ata.rs`), logs and comments following. Discriminators are unchanged so the e-token CPI stays wire-compatible. The e-token side is renamed in magicblock-labs/ephemeral-spl-token#127 (`EnsureMagicAtaDestination`, `MergeShuttleIntoMagicAta`, renumbered to 34/35/36 after the merge with main); the pinocchio SDK helpers keep their names for now.

**On the funded Magic ATA + base delegation question**: the delegation succeeds on base. In the ER, the eATA projection is skipped because the local funded ATA is delegated and not drained (`ata_blocks_projection`). Nothing fails for the user: transfers keep working in the ER against the Magic ATA balance, but the base deposit does not show up until the Magic ATA is drained. This is by design: a funded Magic ATA takes precedence over a base delegation of the same ATA, and the two merge only once the Magic ATA is drained to zero through the shuttle flow, after which the eATA projection replaces it. That invariant is now documented on `MagicAtaInfo` and `ata_blocks_projection`, and the skip site emits a `warn!` naming the ATA and eATA so it is visible in node logs. Deposits into an encrypted destination that already holds a Magic ATA should go through the shuttle merge path in e-token, which executes in the ER.

magicblock-chainlink/src/chainlink/fetch_cloner/mod.rs
```diff
@@ -2117,10 +2117,16 @@ where
...
+    /// A drained Magic ATA is delegated but replaceable: its eATA
+    /// projection must not be deduplicated away, or the stale zero balance
+    /// would hide later eATA deposits.
     fn local_delegated_clone_target_active(&self, pubkey: Pubkey) -> bool {
         self.accounts_bank
             .get_account(&pubkey)
-            .is_some_and(|account| account.delegated())
+            .is_some_and(|account| {
+                account.delegated()
+                    && !ata_projection::is_drained_magic_ata(&pubkey, &account)
```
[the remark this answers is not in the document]
GabrielePicco: Confirmed and fixed in 24f3c9cb3fe830c548749830bfd7070ff8a31a59. The ownership deduper in `clone_account_with_ownership` and its post-failure reconciliation still treated a drained Magic ATA as an authoritative local delegation, so the eATA projection was skipped after passing the entry guard. All three dedupers now go through one Magic ATA exemption (`local_delegation_is_authoritative`), with each site keeping its existing undelegating semantics. Covered by a chainlink unit test on the satisfies-check and by `test_magic_ata_drained_without_close_is_replaced_by_eata_projection` in the cloning suite, which drains a Magic ATA with a plain transfer, delegates an eATA for the same wallet on base, and asserts the ER shows the projected balance with the Magic ATA marker gone.

programs/magicblock/src/ephemeral_accounts/process_create_magic_ata.rs
```diff
@@ -0,0 +1,655 @@
...
+    initial_state: AccountState,
+) -> Result<(), InstructionError> {
+    if *token_program == TOKEN_PROGRAM_ID {
+        // Both token crates define their own AccountState enum with the same
+        // u8 layout, so the shared initial state is carried over by value.
+        let state = SplAccountState::try_from(initial_state as u8)
+            .map_err(|_| InstructionError::InvalidAccountData)?;
+        let account = SplAccount {
+            mint,
+            owner: wallet_owner,
```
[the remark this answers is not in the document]
GabrielePicco: Not addressing in this PR. Only the wallet owner can run `SetAuthority(AccountOwner)` on their own legacy Magic ATA, and doing so orphans their own account: `CloseMagicAta` no-ops and their own later eATA projection stays hidden until they restore the owner. No third party is affected and no funds are at risk; base assets remain recoverable through base-layer undelegation. It is also not specific to Magic ATAs: a legacy base ATA with a reassigned owner fails the same derivation check in the eATA remap. Guarding it would mean intercepting SPL Token instructions in the executor for a self-inflicted edge case, which is not worth it here. Token-2022 Magic ATAs already carry `ImmutableOwner`.

programs/magicblock/src/ephemeral_accounts/process_create_magic_ata.rs
```diff
@@ -0,0 +1,655 @@
...
+            &ata_pubkey,
+            &ata_shared,
+            &wallet_owner,
+            &mint,
+            &token_program,
+        ) {
+            return Ok(());
+        }
+        return Err(InstructionError::InvalidAccountData);
+    }
```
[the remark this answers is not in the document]
GabrielePicco: Not changing this. The precedence rule is deliberate: a funded Magic ATA wins until it is drained through the shuttle flow, because the alternative of folding both balances into one projection would commit a local-only balance back to an eATA that does not hold it on base, breaking the commit-target invariant. As a griefing vector it costs the attacker real tokens and ER fees on every round while the victim keeps the dust, can drain at any time, and can always recover base funds via base-layer undelegation. We document it as a known property rather than add reconciliation machinery for it.

magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
```diff
@@ -240,12 +240,21 @@ where
         }
     };
 
-    if base_ata.delegated() || base_ata.undelegating() {
+    if ata_blocks_projection(&ata_pubkey, &ata) {
```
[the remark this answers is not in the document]
GabrielePicco: Confirmed and fixed in b798c17001fb22d0a0390ab10c96ff61b83e9923. The lookup now treats a drained Magic ATA as a layout of last resort: it fetches the base ATA for the layout when one exists and falls back to the local Magic ATA only when the base ATA does not exist yet. Covered by `test_drained_magic_ata_projection_uses_base_ata_layout`, which projects over a drained Magic ATA from a frozen base ATA with rent and asserts the projection lands frozen with the base lamports (it fails on the previous lookup with Initialized vs Frozen), and by a rent assertion added to the end-to-end replacement test in the cloning suite.

magicblock-chainlink/src/chainlink/fetch_cloner/ata_projection.rs
```diff
@@ -213,39 +213,57 @@ where
...
-            fetch_remote_base_ata(
+            let remote_ata = fetch_remote_ata(
                 this,
                 &ata_pubkeys,
                 eata_account.remote_slot(),
                 companion_fetch_log_context,
             )
-            .await?
+            .await;
+            remote_ata.or(drained_magic_ata)?
```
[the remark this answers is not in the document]
GabrielePicco: Confirmed and fixed in b36e66092a6399df29e9adc84543d8fdfd808e6e. `fetch_remote_ata` now distinguishes found, confirmed absent (every candidate `NotFound`) and unavailable (fetch error or a non-fresh snapshot). Only a confirmed absence falls back to the local Magic ATA layout; on unavailable the drained account is retained and the projection is retried on the next update, as before. Covered by `test_drained_magic_ata_projection_falls_back_to_local_layout_when_base_ata_absent` alongside the base-layout test.

magicblock-core/src/token_programs.rs
snawaz: Still feels hacky. A unique ID can be easily created and maintained.. that's a better approach. It'll also look on the explorer.

snawaz (approved):
Looks good to me now. Thanks for making the last moment changes. 

Also, it's up to you to reconsider using a different ID for the CLOSE_AUTHORITY. I believe that's a cleaner approach.

GabrielePicco:
> CLOSE_AUTHORITY

Thanks. Keeping CLOSE_AUTHORITY for now as an artistic signature

PR #1630, opened 2026-09-01
feat(verifier): enable replication relays

### LINKED ISSUE #1617: Enable verifier replication relays
## Goal

Allow a verifier with the source authority key to relay its replicated Engine state to downstream verifiers.

## Why

`magicblock-verifier` currently starts only `ReplicationClient`. Engine supports shared-key followers acting as replication relays, but the verifier neither exposes downstream dispatcher configuration nor starts `ReplicationDispatcher`, so downstream verifiers cannot connect to it.

## Scope

- Add downstream bind-address and follower-allowlist settings to verifier replication configuration.
- Start `ReplicationDispatcher` for each verifier Engine lifecycle, including after snapshot-driven reopen.
- Coordinate dispatcher startup, failure, and shutdown through the Engine `ShutdownManager`.
- Preserve upstream client authentication and replication behavior.
- Preserve distinct-key verifiers as terminal leaves; dispatcher startup must remain disabled when the local signer differs from the represented source authority.
- Keep leader-only RPC, settlement, scheduling, Chainlink, and admin services disabled.

bmuddha:
## What changed

- add downstream bind-address and follower-allowlist settings to verifier replication configuration
- start a downstream replication dispatcher for each verifier Engine generation when followers are allowed
- coordinate replication startup, failure, restart, and shutdown through the Engine lifecycle manager

Closes #1617

## Impact

Existing verifier configurations remain valid through default downstream settings. Distinct-key verifiers remain terminal leaves because Engine rejects dispatcher startup before binding when the local signer does not represent the source authority. Leader-only services remain disabled.

## Reviewer notes

The relay authority boundary remains owned by `ReplicationDispatcher::spawn`. Snapshot-driven reopen reconstructs both upstream replication and the configured downstream dispatcher while the metrics endpoint retains its process lifetime.

### DIFF
--- a/bins/magicblock-verifier/src/main.rs
+++ b/bins/magicblock-verifier/src/main.rs
@@ -4,7 +4,7 @@ use anyhow::{Context, Result};
 use engine::Engine;
 use magicblock_config::{ConfigError, VerifierParams};
 use nucleus::shutdown::{Service, ShutdownManager, ShutdownReason};
-use replicator::ReplicationClient;
+use replicator::{ReplicationClient, ReplicationDispatcher};
 use tokio::sync::mpsc;
 use tracing::{error, info};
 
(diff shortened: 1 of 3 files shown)


bmuddha:
@coderabbit review

PR #1631, opened 2026-09-01
fix: mirror base chain rent parameters in the ephemeral validator

GabrielePicco:
## Problem

[SIMD-0194](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0194-deprecate-rent-exemption-threshold.md) and [SIMD-0437](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0437-incremental-rent-reduction.md) lower Solana's rent rate in five incremental steps: 6960 → 6333 → 5080 → 2575 → 1322 → 696 lamports per byte. The first step is already live on devnet (Agave v4.3: `lamports_per_byte_year: 6333`, `exemption_threshold: 1.0`).

The ephemeral validator used a hardcoded `Rent::default()` (6960 effective) for both the SVM rent-state checks and the rent sysvar programs read. Any divergence from the base chain's rate is broken, in both directions:

- **ER rate higher than base:** accounts funded at the base chain's (lower) rent-exempt minimum fall below the ER's threshold, so cloning them into the ER fails the rent-state check with `InsufficientFundsForRent`. This is happening on devnet today for any account created at the new minimum.
- **ER rate lower than base:** users can fund or drain accounts inside the ER to balances the base layer refuses on commit. Since commits are paid by the validator under the assumption rent was paid 1:1 in the ER, the validator either eats the difference or the commit fails and the intent is stuck — a cheap drain/griefing vector.

The only safe value is an exact match with the base chain, which changes over time as the reduction steps activate — so it cannot be hardcoded.

## Fix

- **`magic_validator.rs`** — at startup the primary fetches the base chain's rent sysvar (`sync_rent_from_base_chain`) and stores it as the ER's own rent sysvar, overwriting the value held by an existing ledger. Startup fails if the fetch fails. Replicas and disabled-chainlink modes skip the fetch.
- **`magicblock-processor`** — `build_svm_env` now derives `TransactionProcessingEnvironment::rent` from that local rent sysvar (`read_rent_sysvar`), falling back to `Rent::default()` only when the sysvar is absent (tests, chainlink-disabled ledgers). Since `prepare_sysvars` writes the sysvar from the environment rent, the rent-state check and the rate programs fund new accounts at are identical by construction.

Replicas inherit the sysvar via state replication and derive the same environment from it, so replay stays deterministic even if the base chain steps down between primary and replica startup.

## Known limitations

Rent is a startup-frozen input to execution, so a rent-reduction step activating on the base chain interacts with running validators in three bounded ways (all rare — five activations total — and all resolved by a restart):

- A step activating **while the validator runs** requires a restart to be picked up; until then, cloning of accounts created at the new lower minimum fails. A follow-up watchdog comparing the remote rent sysvar against the local one (alert or graceful restart on mismatch) would close this gap.
- The rent sync runs only **after ledger replay** and executors adopt it at the next block registration, so replay always executes under the persisted rate its transactions were recorded with. The residual window is a crash between the sync write and the next persisted state, where the unreplayed ledger tail replays under the new rate.
- During a **rolling restart across an activation**, a primary and replica can briefly hold different rates since each fetches at its own boot.

Offline mode deliberately keeps the last persisted rent so recovered ledgers replay under the rate they were produced with.

### DIFF
--- a/magicblock-api/src/magic_validator.rs
+++ b/magicblock-api/src/magic_validator.rs
@@ -365,18 +365,6 @@ impl MagicValidator {
         }
         let base_fee = config.validator.basefee;
 
-        // Rent must match the base layer exactly: commits there are paid by
-        // the validator under the assumption they were paid 1:1 in the ER.
-        // Replicas fetch as well, since the primary's direct AccountsDb
-        // write of the sysvar is not carried over by replication. Offline
-        // mode keeps the default rent and needs no remote.
-        if config.lifecycle.needs_remote_account_provider() {
-            let step_start = Instant::now();
-            Self::sync_rent_from_base_chain(config.rpc_url(), &accountsdb)
-                .await?;
-            log_timing("startup", "sync_rent_from_base_chain", step_start);
-        }
-
         let svm_env = build_svm_env(&accountsdb, latest_block.blockhash, 0);
         let feature_set = svm_env.feature_set.clone();
         let txn_scheduler_state = TransactionSchedulerState {
@@ -827,13 +815,15 @@ impl MagicValidator {
     /// while a lower rate admits account states the base layer refuses on
     /// commit — settlement the validator pays for assuming 1:1 rent parity.
     ///
-    /// NOTE: rent is a startup-frozen input to execution. A rate reduction
-    /// activating on the base chain while the validator runs requires a
-    /// restart to be picked up, a ledger spanning a rate change replays
-    /// entirely under the rate fetched at boot (snapshot on activation to
-    /// avoid replay divergence), and Offline mode deliberately keeps the
-    /// last persisted value so recovered ledgers replay under the rate
-    /// they were produced with.
+    /// Called only after ledger replay has completed: replay must run under
+    /// the persisted rent its transactions were recorded with, and the new
+    /// value takes effect at the next block boundary (see `refresh_rent` in
+    /// the transaction executor).
+    ///
+    /// NOTE: a rate reduction activating on the base chain while the
+    /// validator runs still requires a restart to be picked up. Offline mode
+    /// deliberately keeps the last persisted value so recovered ledgers
+    /// replay under the rate they were produced with.
     async fn sync_rent_from_base_chain(
         rpc_url: &str,
         accountsdb: &AccountsDb,
@@ -1143,6 +1133,24 @@ impl MagicValidator {
         self.maybe_process_ledger().await?;
 
         log_timing("startup", "maybe_process_ledger", step_start);
+
+        // Rent must match the base layer exactly: commits there are paid by
+        // the validator under the assumption they were paid 1:1 in the ER.
+        // Synced only after ledger replay, which must run under the persisted
+        // rent its transactions were recorded with; executors adopt the new
+        // value at the next block boundary (see `refresh_rent`). Replicas
+        // fetch as well, since the direct AccountsDb write of the sysvar is
+        // not carried over by replication. Offline mode keeps the persisted
+        // rent and needs no remote.
+        if self.config.lifecycle.needs_remote_account_provider() {
+            let step_start = Instant::now();
+            Self::sync_rent_from_base_chain(
+                self.config.rpc_url(),
+                &self.accountsdb,
+            )
+            .await?;
+            log_timing("startup", "sync_rent_from_base_chain", step_start);
+        }
 
         if self.config.accountsdb.defragment_on_startup {
             let step_start = Instant::now();
--- a/magicblock-api/src/magic_validator.rs
+++ b/magicblock-api/src/magic_validator.rs
@@ -367,10 +367,10 @@ impl MagicValidator {
 
         // Rent must match the base layer exactly: commits there are paid by
         // the validator under the assumption they were paid 1:1 in the ER.
-        // Replicas skip the fetch and inherit the sysvar via replication.
-        if !Self::replication_mode_uses_disabled_chainlink(
-            &config.validator.replication_mode,
-        ) {
+        // Replicas fetch as well, since the primary's direct AccountsDb
+        // write of the sysvar is not carried over by replication. Offline
+        // mode keeps the default rent and needs no remote.
+        if config.lifecycle.needs_remote_account_provider() {
             let step_start = Instant::now();
             Self::sync_rent_from_base_chain(config.rpc_url(), &accountsdb)
                 .await?;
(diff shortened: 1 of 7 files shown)


bmuddha (approved):
LGTM

bzawisto (approved):

magicblock-api/src/magic_validator.rs
```diff
@@ -362,6 +365,18 @@ impl MagicValidator {
         }
         let base_fee = config.validator.basefee;
 
+        // Rent must match the base layer exactly: commits there are paid by
+        // the validator under the assumption they were paid 1:1 in the ER.
+        // Replicas skip the fetch and inherit the sysvar via replication.
+        if !Self::replication_mode_uses_disabled_chainlink(
+            &config.validator.replication_mode,
+        ) {
```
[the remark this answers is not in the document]
GabrielePicco: Fixed in b85d152: the sync is now gated on `lifecycle.needs_remote_account_provider()` so replicas fetch from the base chain too, before `build_svm_env` freezes the rate.

magicblock-api/src/magic_validator.rs
```diff
@@ -362,6 +365,18 @@ impl MagicValidator {
...
+            let step_start = Instant::now();
+            Self::sync_rent_from_base_chain(config.rpc_url(), &accountsdb)
+                .await?;
```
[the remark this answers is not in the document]
GabrielePicco: Fixed in b85d152: the gate is now `lifecycle.needs_remote_account_provider()`, so Offline mode skips the fetch and keeps the local rent.

magicblock-api/src/magic_validator.rs
```diff
@@ -803,6 +818,40 @@ impl MagicValidator {
...
+        )
+        .get_account(&sysvar::rent::ID)
+        .await
+        .map_err(|e| err(&e))?;
+        let rent: Rent =
+            bincode::deserialize(&account.data).map_err(|e| err(&e))?;
+        info!(?rent, "Synced rent parameters from base chain");
+        let account = AccountSharedData::new_data(1, &rent, &sysvar::ID)
+            .map_err(|e| err(&e))?;
+        let _ = accountsdb.insert_account(&sysvar::rent::ID, &account);
```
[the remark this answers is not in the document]
GabrielePicco: Fixed in b85d152: the insertion error now propagates as `FailedToSyncBaseChainRent`.

magicblock-api/src/magic_validator.rs
```diff
@@ -362,6 +365,18 @@ impl MagicValidator {
...
 
+        // Rent must match the base layer exactly: commits there are paid by
+        // the validator under the assumption they were paid 1:1 in the ER.
+        // Replicas fetch as well, since the primary's direct AccountsDb
+        // write of the sysvar is not carried over by replication. Offline
+        // mode keeps the default rent and needs no remote.
+        if config.lifecycle.needs_remote_account_provider() {
+            let step_start = Instant::now();
+            Self::sync_rent_from_base_chain(config.rpc_url(), &accountsdb)
+                .await?;
```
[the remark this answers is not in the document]
GabrielePicco: Fixed in 6889e39: the sync now runs in `start()` after `maybe_process_ledger`, so replay executes under the persisted rent; executors adopt the new value at the next block registration (`refresh_rent`).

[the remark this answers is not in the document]
GabrielePicco: Accepted trade-off for now: account state changes (including clones) already reach replicas through the replicated transaction stream, and carrying the rent activation boundary itself through replication/snapshots is deliberately out of scope for this fix. Documented in the known limitations.

magicblock-processor/src/lib.rs
```diff
@@ -111,6 +119,19 @@ pub fn build_svm_env(
     }
 }
 
+/// Reads the rent parameters from the local rent sysvar account, falling
+/// back to the classic defaults when it's absent (fresh ledgers with
+/// chainlink disabled, tests).
+fn read_rent_sysvar(accountsdb: &AccountsDb) -> Rent {
+    let Some(account) = accountsdb.get_account(&sysvar::rent::ID) else {
+        return Rent::default();
```
snawaz, line 127: We should `error!` in the `else {}` block?

Or maybe the return should be `Option<Rent>` so that the caller will decide whether to fail or use the default. In any case, `error!` would be helpful.

magicblock-processor/src/lib.rs
```diff
@@ -111,6 +119,19 @@ pub fn build_svm_env(
...
+    };
+    bincode::deserialize(account.data()).unwrap_or_else(|err| {
+        error!(?err, "Failed to deserialize rent sysvar, using defaults");
+        Rent::default()
```
snawaz, line 131: It is a fatal-error, isn't it? Why would a rent-account fail to deserialize?

magicblock-processor/src/executor/mod.rs
```diff
@@ -236,6 +239,33 @@ impl TransactionExecutor {
...
+    /// captured at executor spawn can be stale; picking the sysvar up here
+    /// switches every executor within one block of the sync, keeping the
+    /// rent-state checks and the sysvar programs read consistent.
+    fn refresh_rent(&mut self) {
+        let Some(account) = self.accountsdb.get_account(&sysvar::rent::ID)
+        else {
+            return;
+        };
+        let Ok(rent) = bincode::deserialize::<Rent>(account.data()) else {
+            return;
```
snawaz, line 257: The `else{}` probably hides a serious problem?

magicblock-api/src/magic_validator.rs
```diff
@@ -1088,6 +1134,24 @@ impl MagicValidator {
...
+        // fetch as well, since the direct AccountsDb write of the sysvar is
+        // not carried over by replication. Offline mode keeps the persisted
+        // rent and needs no remote.
+        if self.config.lifecycle.needs_remote_account_provider() {
+            let step_start = Instant::now();
+            Self::sync_rent_from_base_chain(
+                self.config.rpc_url(),
+                &self.accountsdb,
+            )
+            .await?;
```
snawaz, line 1151: The error from `sync_rent_from_base_chain` should be carefully handled. Maybe it's not meaningful to continue in case of error.

snawaz (commented):
Posted couple of comments and questions.

magicblock-api/src/magic_validator.rs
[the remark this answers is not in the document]
GabrielePicco: tradeoff accepted, will follow up on this

magicblock-processor/src/lib.rs
GabrielePicco: Possible, decided to maintain it soft as of now, as it was previously harcoded

magicblock-processor/src/lib.rs
GabrielePicco: Same rationale as this: https://github.com/magicblock-labs/magicblock-validator/pull/1631#discussion_r3904631591, but agree, it shouldn't fail unless there are RPC issues

magicblock-api/src/magic_validator.rs
GabrielePicco: In general I agree, but used same rationale as commented here: https://github.com/magicblock-labs/magicblock-validator/pull/1631#discussion_r3904631591

PR #1641, opened 2026-09-03
fix(committor): return error when MagicContext is missing in accept loop

crazywriter1:
## Summary
- `accept_scheduled_intents` used `.expect()` when reading `MAGIC_CONTEXT_PUBKEY` from AccountsDb
- A missing MagicContext account would panic the committor accept task on every slot tick
- Return `InternalIntentClientError::MagicContextMissing` instead; the existing loop in `service.rs` already logs and continues on `Err`

Related to #1018 (same panic-in-worker family); does not fix the transaction_preparator site.

## Test plan
- [X] `cargo check -p magicblock-committor-service`

## Summary by CodeRabbit

- **Bug Fixes**
  - Scheduled intent acceptance now reports a clear error when the required MagicContext account is unavailable, instead of causing a panic.


bmuddha (approved):
LGTM

crazywriter1:
Thanks for the review sir.

PR #1643, opened 2026-09-03
fix(chainlink): return error on resolved-accounts length mismatch

crazywriter1:
## Summary
- After awaiting fetch waiters in `try_get_multi`, a length mismatch between `resolved_accounts` and `pubkeys` used `assert_eq!`, which panics in release builds
- Sibling path `fetch_multi_rpc_only` already returns `AccountResolutionsFailed` for cardinality mismatches
- Replace the assert with `Err(AccountResolutionsFailed)` so callers get a fetch error instead of a process crash
- Related to #1574 (same panic-to-error family in `remote_account_provider`); different site — lagging response slot vs resolved-accounts length

## Breaking Changes
- None

## Test plan
- [x] `cargo +nightly fmt --check -p magicblock-chainlink -- --config-path rustfmt-nightly.toml`


magicblock-chainlink/src/remote_account_provider/mod.rs
```diff
@@ -1782,11 +1782,21 @@ impl<T: ChainRpcClient, U: ChainPubsubClient> RemoteAccountProvider<T, U> {
...
+            // Prefer Err over assert_eq! so a mismatch fails the fetch like the
+            // cardinality check in `fetch_multi_rpc_only` instead of panicking.
+            if resolved_accounts.len() != pubkeys.len() {
+                return Err(
+                    RemoteAccountProviderError::AccountResolutionsFailed(
+                        format!(
+                            "Resolved {} accounts for {} requested accounts: {}",
+                            resolved_accounts.len(),
+                            pubkeys.len(),
+                            pubkeys_str(pubkeys)
```
bmuddha: No need to format the pubkeys here, returning counts is sufficient, this error should never happen, and it's mostly there to satisfy formal correctness verification.

bmuddha (changes requested):

magicblock-chainlink/src/remote_account_provider/mod.rs
crazywriter1: Thanks for the review. Addressed. error message now only includes the counts.

bmuddha (approved):

Issue #133, opened 2024-08-19
bug: program binary data re-cloning doesn't work properly


crypto-vincent:
## Summary

Originally reported by Zeebit. They are unable to automatically re-deploy programs inside the validator by deploying on devnet.

## Details

Checked that the cloning is indeed being triggered through the mutator, but for some reason the updated program data is not being used in transaction execution (probably some caching problem)

PR #1647, opened 2026-09-04
fix: derive the magic fee vault from the effective validator authority

GabrielePicco:
# fix: derive the magic fee vault from the effective validator authority

`magic_fee_vault_pubkey` derived the vault PDA from the raw local `validator_authority_id()`. Replicas run with an authority override, so a replica replaying the primary's transactions expected its *own* vault in the fee-vault slot and rejected the primary's, diverging on every delegated-payer commit. Deriving from `effective_validator_authority_id()` aligns the vault derivation with the authority handling everywhere else in the module (`validate_commit_schedule_permissions`, `validate_callback_accounts`).

(The override is a process-global `OnceLock`, so an in-process unit test would poison the rest of the suite — replica replay is covered by the replica integration environment.)

## History

Earlier revisions of this PR added explicit fee-vault instruction variants (`ScheduleCommitWithFeeVault`, tags 26/27). After review discussion the team concluded the vault slot is the CPI parent's responsibility — the magic program trusts its parent for the accounts it forwards, and callers with an optional vault slot validate it caller-side (ephemeral-spl-token#136 carries that guard). The variants were dropped in c3ca8a3; the instruction set is unchanged and only this bugfix remains.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fee-vault resolution now uses the effective validator authority, improving consistency when validator authority changes.
  - Explicit fee-vault declarations are now rejected for payers that do not require fees.
  - Fee-free transactions without an explicit fee-vault declaration continue to process normally.


magicblock-magic-program-api/src/instruction.rs
```diff
@@ -311,6 +311,24 @@ pub enum MagicBlockInstruction {
...
+    /// being inferred from the payer's delegation status. Its pubkey is
+    /// always validated, so it can never be reinterpreted as an account to
+    /// commit; the payer is charged only when delegated.
+    ///
+    /// # Account references
+    /// - **0.**   `[WRITE, SIGNER]` Payer requesting the commit to be scheduled
+    /// - **1.**   `[WRITE]`         Magic Context Account
+    /// - **2.**   `[WRITE]`         Magic fee vault of the executing validator
+    /// - **3..n** `[]`              Accounts to be committed
+    ScheduleCommitWithFeeVault,
```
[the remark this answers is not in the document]
GabrielePicco: The allowlist gates the opposite direction: `PRIVILEGED_MAGIC_DISCRIMINANTS` ([access_permissions.rs:13](https://github.com/magicblock-labs/magicblock-svm/blob/0395009/svm/src/access_permissions.rs#L13)) decides when a **privileged fee payer** (the validator authority) is granted `PrivilegedAccess::Full`, and it deliberately excludes every user-facing scheduling instruction — `ScheduleCommit` (1), `ScheduleCommitAndUndelegate` (2), `ScheduleBaseIntent` (5) and `ScheduleIntentBundle` (11) are all absent from it today. Tags 26/27 are scheduling instructions submitted by ordinary, non-privileged payers, so being outside the allowlist gives them exactly the classification the existing tags 1/2 have. The non-privileged path in `validate_accounts_access` is discriminant-agnostic (writable accounts must be delegated/ephemeral/undelegating — same accounts, same flags as tag 2), so nothing in the pinned runtime rejects or rolls back the new variants. Adding 26/27 to the allowlist would actually be wrong: it would let the validator authority bypass write checks for a user-schedulable instruction. End-to-end coverage of the new tags lands with the first caller (ephemeral-spl-token#136 emits tag 27 from its undelegate instruction; its suite asserts the emission and account order).

programs/magicblock/src/schedule_transactions/mod.rs
```diff
@@ -196,6 +205,12 @@ pub(crate) fn try_get_fee_vault<'a, 'ix_data>(
         return Err(InstructionError::MissingAccount);
     }
 
+    if !payer_requires_fee_vault {
+        // Explicitly provided vault with a payer that pays no fees:
+        // validated above, but nothing to charge.
+        return Ok(None);
+    }
```
snawaz: It makes the caller buggy because of the modified condition in the schedule-commit path. The account at `fee_vault_idx` is fee-vault (as ensured above) but the caller will assume there is no fee-vault and the account at `fee_vault_idx` will be treated as delegated account.

GabrielePicco: You're right that the return value was carrying two meanings there. Since our business logic only has two cases (non-delegated payer → nonce-limited commits, paid on undelegation; delegated payer → charged per commit, vault required and validated), the in-between state shouldn't exist at all: as of 5946288, an explicit fee vault with a non-fee-paying payer is rejected with InvalidInstructionData up front. `Ok(None)` now always means "this instruction carries no fee vault", and the committee offset derives from `is_some()` alone — the state you flagged is gone.

snawaz: The https://github.com/magicblock-labs/magicblock-validator/commit/5946288673dbdd86f08906621bff463ef9dfab43 looks good.

--

But when zooming out, the newly added instructions do not add anything new that the corresponding older instructions do not provide. When fee_vault is passed, then there is 100% overlap.

Instead of adding the new instructions, we could just reorder the checks as follows:

```rust
pub(crate) fn try_get_fee_vault<'a, 'ix_data>(
    transaction_context: &'a TransactionContext<'ix_data>,
    invoke_context: &InvokeContext,
    payer_idx: u16,
    fee_vault_idx: u16,
) -> Result<Option<InstructionAccount<'a, 'ix_data>>, InstructionError> {
    
    //
    // 1. first and foremost, check if the account at `fee_vault_idx` is really a fee_vault.
    // 
    
    let fee_vault =  {
        let vault_pubkey =
          get_instruction_pubkey_with_idx(transaction_context, fee_vault_idx)?;

        if vault_pubkey == &magic_fee_vault_pubkey() {
           let vault_account =
             get_instruction_account_with_idx(transaction_context, fee_vault_idx)?;
           let is_vault_writable =
             get_writable_with_idx(transaction_context, fee_vault_idx)?;
           if !vault_account.to_account_shared_data()?.delegated()
             || !is_vault_writable
           {
               ic_msg!(
                 invoke_context,
                 "ScheduleCommit ERR: magic fee vault must be writable and delegated"
               );
               return Err(InstructionError::IllegalOwner);
            }

            Some(vault_account)
        } else {
            None
        }
     };

    // 
    // 2. at this point, we could simply return as:
    //    return Ok(fee_vault);
    //
    // but probably that might break max-10-commits semantic (I'm not sure).
    // if so, we could do this:
    //
    
    // If we compute the next index in this function itself, maybe we 
    // could name it `next_account_idx` if `committees_start` is misleading (as 
    // the function is called from different places). I chose 
    // `committees_start` to communicate the idea.
    let committees_start = fee_vault_idx + fee_vault.is_some() as u16;
    
    let payer_requires_fee_vault = {
        let payer_account =
          get_instruction_account_with_idx(transaction_context, payer_idx)?;
   
        let payer = payer_account.to_account_shared_data()?;
        payer.delegated() && !payer.confined()
    };
    
    Ok(if payer_requires_fee_vault {
        if fee_vault.is_none() {
          return Err(..);
        }
        (fee_vault, committees_start)
    } else {
        // the caller will SKIP the fee_vault, even if 
        // fee_vault is present because committees_start 
        // will point at the next account.
        //
        // if we want, we may error for this case:
        //   if fee_vault.is_some() {
        //     return Err(..);
        //   }
        (None, committees_start)
    })
}
```

GabrielePicco: The "else" branch can't reject: with the old tags, a non-vault account at that index is every existing plain commit's first committee — same position, same bytes. Only the caller knows whether the slot was meant as a vault, and that intent is exactly one bit the implicit encoding has nowhere to carry; the new discriminant is that bit.

GabrielePicco: Also note that once the SDKs emit the explicit tags and callers migrate, the implicit vault inference can be deleted from tags 1/2 entirely: payer.delegated()-based slot interpretation goes away, tags 1/2 become pure "committees only", the explicit_fee_vault flag parameter disappears, and there's exactly one fee-vault code path. Opened 2 issue to track the refactoring: https://github.com/magicblock-labs/ephemeral-rollups-sdk/issues/287 and https://github.com/magicblock-labs/magicblock-validator/issues/1651

GabrielePicco: Done — you win this one on architecture: the magic program already trusts its CPI parent for every account it forwards (that's what the carve-out is), so guarding one caller's slot at the protocol layer was fighting the trust model instead of using it. c3ca8a3 drops the discriminators entirely; this PR is now only the replica fee-vault derivation fix, and the instruction set is untouched (no replay/rollout concerns). The #133 shape is closed caller-side instead: ephemeral-spl-token#136 now requires the forwarded sixth account to be delegation-program-owned (one line, no instruction data), which excludes both classes the committee permission would admit under e-token parenthood (token-program-owned via the carve-out, e-token-owned via the parent match), while everything DLP-owned self-rejects downstream. Simpler on both sides than where we started — thanks for pushing on it.

GabrielePicco: The core of the issue is that the calling smart contract always pass the `vault` and the user can decide to pass a not delegated payer which would make the `vault` part of the committed accounts.

PR #1649, opened 2026-09-05
fix(committor): map serialized_transaction_size failures to Result

crazywriter1:
## Summary
- `serialized_transaction_size` used double `.unwrap()` on `wincode::serialized_size` and `usize::try_from`, which can panic strategist / size-validator workers on serialize failure
- Return `Result<usize, SerializedTransactionSizeError>` instead
- Fit checks treat size errors as non-fitting (`false` / `usize::MAX`), matching existing “doesn’t fit” handling
- Related to #1018 (same panic-in-committor-worker family); different site — not covered by #1492

## Breaking Changes
- `serialized_transaction_size` now returns `Result<usize, SerializedTransactionSizeError>` instead of `usize` (no longer panics). No in-tree external callers; workspace crate, no major version bump.

## Test plan
- [x] `cargo check -p magicblock-committor-service --tests`
- [x] `cargo test -p magicblock-committor-service close_buffer_limit`

### DIFF
--- a/magicblock-committor-service/src/transactions.rs
+++ b/magicblock-committor-service/src/transactions.rs
@@ -1,13 +1,25 @@
 use solana_packet::PACKET_DATA_SIZE;
-use wincode::{SchemaWrite, config::DefaultConfig};
+use thiserror::Error;
+use wincode::{SchemaWrite, config::DefaultConfig, error::WriteError};
 
 /// Maximum serialized transaction size that can be sent over the wire.
 pub(crate) const MAX_TRANSACTION_WIRE_SIZE: usize = PACKET_DATA_SIZE;
 
-pub fn serialized_transaction_size<T>(transaction: &T) -> usize
+#[derive(Debug, Error)]
+pub enum SerializedTransactionSizeError {
+    #[error("Failed to compute serialized transaction size: {0}")]
+    Serialize(#[from] WriteError),
+    #[error("Serialized transaction size does not fit in usize")]
+    SizeOverflow,
+}
+
+pub fn serialized_transaction_size<T>(
+    transaction: &T,
+) -> Result<usize, SerializedTransactionSizeError>
 where
     T: SchemaWrite<DefaultConfig, Src = T> + ?Sized,
 {
-    // SAFETY: runs on transactions we already serialize before sending.
-    usize::try_from(wincode::serialized_size(transaction).unwrap()).unwrap()
+    let size = wincode::serialized_size(transaction)?;
+    usize::try_from(size)
+        .map_err(|_| SerializedTransactionSizeError::SizeOverflow)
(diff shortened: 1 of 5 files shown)


magicblock-committor-service/src/transactions.rs
```diff
@@ -1,13 +1,25 @@
...
+pub enum SerializedTransactionSizeError {
+    #[error("Failed to compute serialized transaction size: {0}")]
+    Serialize(#[from] WriteError),
+    #[error("Serialized transaction size does not fit in usize")]
+    SizeOverflow,
+}
+
+pub fn serialized_transaction_size<T>(
+    transaction: &T,
+) -> Result<usize, SerializedTransactionSizeError>
```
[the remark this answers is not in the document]
crazywriter1: Addressed in a79e7ff: documented the intentional `Result` signature change under `# Compatibility` on `serialized_transaction_size`. No in-tree external callers; workspace crate, no major version bump. PR Breaking Changes updated accordingly.

magicblock-committor-service/src/transactions.rs
```diff
@@ -1,13 +1,35 @@
...
+/// type is an intentional public API change; callers must handle the error.
+pub fn serialized_transaction_size<T>(
+    transaction: &T,
+) -> Result<usize, SerializedTransactionSizeError>
 where
     T: SchemaWrite<DefaultConfig, Src = T> + ?Sized,
 {
-    // SAFETY: runs on transactions we already serialize before sending.
-    usize::try_from(wincode::serialized_size(transaction).unwrap()).unwrap()
+    let size = wincode::serialized_size(transaction)?;
```
bmuddha: let's just `unwrap_or(usize::MAX) here, also we do not support 32 bit platforms (the code just won't compile) so casting `u64 as usize` is acceptable. This way all the clumsy error handling will not be necessary.

magicblock-committor-service/src/transactions.rs
```diff
@@ -1,13 +1,35 @@
...
+    SizeOverflow,
+}
+
+/// Returns the serialized wire size of `transaction`.
+///
+/// # Errors
+/// Returns [`SerializedTransactionSizeError`] if serialization fails or the size
+/// does not fit in `usize`.
+///
+/// # Compatibility
```
bmuddha: no need to AI narrate the code so verbosely.

bmuddha (changes requested):

crazywriter1:
Addressed `serialized_transaction_size` now returns `usize` via `unwrap_or(usize::MAX)` and `size as usize`. Removed the error type and verbose docs; callers are back to plain `usize` usage. @bmuddha

bmuddha (approved):

PR #1358, opened 2026-06-24
feat: adds actual parent_blockhash to ReplicaBlockInfoV4

0xzrf:
## Summary
Adds actual `parent_blockhash` in `ReplicaBlockInfoV4` in stead of a dummy value

## Changes
- Added a new `LatestBlockInner::new_with_parent()` that takes a parent_blockhash(to avoid breaking existing tests) and a new field `LatestBlockInner::parent_blockhash`
- Supply the `parent_blockhash` from `apply_replayed_block` and `prepare_block_as_primary` before it's updated to the latest blockhash
- uses the applied `parent_blockhash` inside `notify_block` instead of a dummy placeholder

## Breaking Changes
- [x] None

## Test Plan
Extended existing tests and added new tests to verify the changes. You can run the new tests using:
```bash
cargo test -p magicblock-ledger --test get_block test_write_block_stores_parent_blockhash
cargo test -p magicblock-processor --test replica_ordering test_replay_block
cargo test -p magicblock-processor --test scheduling test_primary_slot_transition_sets_parent_blockhash
```

## Release Notes

* **Bug Fixes**
  * Block metadata now correctly includes parent block information instead of placeholder values, ensuring proper block linkage and chain integrity.

* **Tests**
  * Added tests to verify parent block hash persistence and correctness during block transitions.

PR #1650, opened 2026-09-05
fix(aperture): wire geyser parent_blockhash from Block.parent

crazywriter1:
## Summary
- Geyser `ReplicaBlockInfoV4` used a hardcoded `parent_blockhash` placeholder (`11111…`)
- Wire it from engine `Block.parent`, which the sequencer already sets
- Intentionally scoped to `parent_blockhash` only — do not fill all three metadata fields in one PR
- `executed_transaction_count` / `entry_count` stay placeholders: engine `Block` does not expose them yet
- Related to #1358 (closed without merge; superseded by engine — parent is available now)

## Breaking Changes
- None

## Test plan
- [x] `cargo check -p magicblock-aperture --lib`

### DIFF
--- a/magicblock-aperture/src/geyser.rs
+++ b/magicblock-aperture/src/geyser.rs
@@ -210,6 +210,10 @@ impl GeyserPluginManager {
         }
     }
 
+    /// Delivers rooted slot status and block metadata for a sealed engine block.
+    ///
+    /// `parent_blockhash` is taken from [`Block::parent`]. Transaction and entry
+    /// counts remain placeholders until the engine retains those fields.
     fn notify_block(&self, block: Block) {
         let parent = block.slot.checked_sub(1);
         for plugin in &self.plugins {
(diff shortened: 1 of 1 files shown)


magicblock-aperture/src/geyser.rs
```diff
@@ -210,6 +210,10 @@ impl GeyserPluginManager {
         }
     }
 
+    /// Delivers rooted slot status and block metadata for a sealed engine block.
```
bmuddha: The comment is overly descriptive, annotating relevant placeholder inline would be more appropriate.

magicblock-aperture/src/geyser.rs
```diff
@@ -466,6 +473,35 @@ mod tests {
         }
     }
 
+    /// Ensures `notify_block` emits `Block.parent`, not the all-zero placeholder.
```
bmuddha: no need for test, since it's apparent from code that parent blockhash is alwasy set.

bmuddha (changes requested):

crazywriter1:
Thanks for the review. Addressed. Removed the rustdoc and the dedicated unit test; left a short inline note on the count placeholders. @bmuddha

bmuddha (approved):
LGTM

PR #1654, opened 2026-09-09
fix:  infer optional fee vault from payer and account layout

snawaz:
**No Breaking Changes**:

It only makes the intended design cleaner and less error prone.

- Magic fee vault is **optional** for `Schedule*` instructions unless the payer is delegated that requires the fee-vault path.
- Keep **AddActionCallback** fee-vault charging **mandatory**.
- The docs are updated to clearly explain the semantics.


GabrielePicco (approved):
LGTM!

